PR #31, opened 2015-08-12
If no cache, don't try to get a sequence number

gperciva:
Fix Solaris crash with --dry-run

Reported by:	Tim Bishop
Bug Bounty:	$200

### DIFF
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@ -423,8 +423,8 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	if ((d->dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
 		goto err4;
 
-	/* Get sequence number. */
-	if (multitape_sequence(cachedir, lastseq))
+	/* If this isn't a dry run, get the sequence number. */
+	if ((d->dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
 		goto err4;
 
 	/* Obtain write cookies from the storage and chunk layers. */
--- a/tar/storage/storage.h
+++ b/tar/storage/storage.h
@@ -78,7 +78,8 @@ void storage_read_free(STORAGE_R *);
  * Start a write transaction, presuming that ${lastseq} is the sequence
  * number of the last committed transaction, or zeroes if there is no
  * previous transaction; and store the sequence number of the new transaction
- * into ${seqnum}.  If ${dryrun} is nonzero, perform a dry run.
+ * into ${seqnum}.  If ${dryrun} is nonzero, perform a dry run
+ * (which ignores ${lastseq} and sets ${seqnum} to 32 0s).
  */
 STORAGE_W * storage_write_start(uint64_t, const uint8_t[32], uint8_t[32],
     int);
--- a/tar/storage/storage_write.c
+++ b/tar/storage/storage_write.c
@@ -121,7 +121,8 @@ raisesigs(struct storage_write_internal * S)
  * Start a write transaction, presuming that ${lastseq} is the sequence
  * number of the last committed transaction, or zeroes if there is no
  * previous transaction; and store the sequence number of the new transaction
- * into ${seqnum}.  If ${dryrun} is nonzero, perform a dry run.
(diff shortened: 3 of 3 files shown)


tar/multitape/multitape_write.c
```diff
@@ -423,8 +423,8 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	if ((d->dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
 		goto err4;
 
-	/* Get sequence number. */
-	if (multitape_sequence(cachedir, lastseq))
+	/* If we're using a cache, get the sequence number. */
+	if ((cachedir != NULL) && (multitape_sequence(cachedir, lastseq)))
```
cperciva: I'm wondering if this should be (cachedir != NULL) or if (d->dryrun == 0) would make more sense.  Also, are we going to end up using lastseq without it being initialized?

gperciva:
`lastseq` is only used in `storage_write_start`, which I changed to ignore `last_seq` if it's a dryrun.  Another options is, of course, to initialize `lastseq` to 0 (?) in `tar/multitape/multitape_write.c` and leave `tar/storage/storage_write.c` alone.  At this point, I don't think I'm in a good position to judge which is the better option.

Unless there's something more subtle then I'm aware of, I think we should be checking `cachedir` for `NULL`, since that's the actual thing that prompts the problem.  True, if they're not doing a dryrun and don't have a `cachedir`, then it should have bailed before we get to this point, but if some bits get flipped due to cosmic rays or whatever, checking `cachedir` seems like a slightly more "defensive programming" thing to do.

cperciva:
Please read: http://www.daemonology.net/blog/2015-08-13-null-is-a-bad-idea.html :-)

If you're going to check whether cachedir is NULL, it should be in the form of "if we're not doing a dry run, assert that cachedir is not NULL".  That might actually be a very good thing to do.  But you should never say "if this pointer is NULL, don't dereference it".

gperciva:
I'm not certain that I understand the relevance of the blog post.  With these fixes, I'm not printing "(null)" -- in fact, I'm specifically avoiding a function that would involve an sprintf NULL.

[argh, I can't make github do what I want.]

Looking above at lines 415 and 419:

    if ((cachedir != NULL) && (dirutil_needdir(cachedir)))
    ...
    if ((cachedir != NULL) && ((d->lockfd = multitape_lock(cachedir)) == -1))

should those also be checking `dryrun` instead of `cachedir`?  My reasoning was that it was more "obvious" to check cachedir before using cachedir.  If those lines were

    if ((!d->dryrun) && (dirutil_needdir(cachedir)))
    ...
    if ((!d->dryrun) && ((d->lockfd = multitape_lock(cachedir)) == -1))

then IMO they would be harder to understand -- the reader needs to know (or remember) that dryrun has no cache, while non-dryrun must have a cache.

From the sounds of things, you'd like to have an

    if (!dryrun)
        assert (cachedir != NULL);

at the top of this function, then check for `d->dryrun` in all those cases?

tar/multitape/multitape_write.c
```diff
@@ -423,8 +423,8 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	if ((d->dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
 		goto err4;
 
-	/* Get sequence number. */
-	if (multitape_sequence(cachedir, lastseq))
+	/* If this isn't a dry run, get the sequence number. */
+	if ((d->dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
```
gperciva: If we are doing a dryrun, this does leave `lastseq` unset and it is passed into `storage_write_start`.

However, as long as we are doing a dryrun, `storage_write_start` does not look at `lastseq`.  (code and comment changed below).

If you'd like me to add

    memset(lastseq, 0, 32);

here, I have no objection, but it's not required given the current state of the code.  (and hopefully nobody will be changing `storage_write_start` in the future in a manner which require `lastseq`, given the comment I added to its "contract".

gperciva:
For clarity: patch updated, and PR ready for review and potential merging.

PR #35, opened 2015-08-14
Solaris crash

gperciva:


libcperciva/util/sock_util.c
```diff
@@ -1,3 +1,5 @@
+#include <assert.h>
```
cperciva: Umm, why?

cperciva:
Can you give me a patch with just the bug fix and not the extra asserts?  I don't want new "features" right now.

And remind me to talk to you about assertion style after I return from BSDCam.  I know I said "feel free to sprinkle asserts" but this isn't exactly what I had in mind.

cperciva:
I've merged the non-assert parts of this in #39, and the PR name is confusing when what's left is just asserts... so I'm going to close this now and you can open a new PR for the asserts after I've had a chance to talk to you more about what I had in mind.

PR #37, opened 2015-08-15
Bash completion packages

gperciva:
First round of updates for bash-completion and Debian packages.

This does *not* enable `--with-bash-completion-dir` in the debian package.  I have not yet been able to test that successfully; I still have a lot to learn about debian packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,17 @@ but for experimental development, use:
     ./configure
     make
 
+
+Packaging notes
+---------------
+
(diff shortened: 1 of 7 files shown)


cperciva:
Does this need to be in 1.0.36?  It mostly looks innocuous -- only thing I'm not sure about is the default bash completion directory.

gperciva:
Updated with no default installation directory

gperciva:
Updated with only bash completion fixes

PR #39, opened 2015-08-17
Don't try to read an empty chunkification cache

gperciva:
This is the non-assert parts of https://github.com/Tarsnap/tarsnap/pull/35.

### DIFF
--- a/tar/write.c
+++ b/tar/write.c
@@ -294,8 +294,8 @@ tarsnap_mode_c(struct bsdtar *bsdtar)
 	    &bsdtar->cachedir_dev, &bsdtar->cachedir_ino))
 		bsdtar_errc(bsdtar, 1, 0, "%s", archive_error_string(a));
 
-	/* Read the chunkification cache. */
-	if (bsdtar->cachecrunch < 2) {
+	/* If the chunkification cache is enabled, read it. */
+	if ((bsdtar->cachecrunch < 2) && (bsdtar->cachedir != NULL)) {
 		bsdtar->chunk_cache = ccache_read(bsdtar->cachedir);
 		if (bsdtar->chunk_cache == NULL)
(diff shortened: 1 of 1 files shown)


tar/write.c
```diff
@@ -993,8 +993,9 @@ write_entry_backend(struct bsdtar *bsdtar, struct archive *a,
 	 * the chunkification cache to find the entry for the file (if one
 	 * already exists) and tell us if it can provide the entire file.
 	 */
-	if ((st != NULL) && S_ISREG(st->st_mode) && rpath != NULL &&
-	    archive_entry_size(entry) > 0 && bsdtar->cachecrunch < 2) {
+	if ((st != NULL) && S_ISREG(st->st_mode) && (rpath != NULL) &&
+	    (archive_entry_size(entry) > 0) && (bsdtar->cachecrunch < 2) &&
+	    (bsdtar->chunk_cache != NULL)) {
```
cperciva: Should this be (bsdtar->cachedir != NULL) ?  I don't think we're setting bsdtar->chunk_cache to NULL anywhere in the cache-not-being-used case.

gperciva:
`tar/bsdtar.c`, line 146:
	memset(bsdtar, 0, sizeof(*bsdtar));
?

cperciva:
Well... zeroing a structure isn't actually guaranteed to set pointers to NULL (weird portability issue) although people tend to assume that it does.  Good enough for now though.  After 1.0.36 we might want to clean up that code slightly by introducing "ccache_enabled" field which is set based on cachedir != NULL and cachecrunch < 2.

gperciva:
huh.  "Q: Seriously, have any actual machines really used nonzero null pointers, or different representations for pointers to different types?"
http://c-faq.com/null/machexamp.html

C portability is scary.

PR #43, opened 2015-08-25
Allocate bsdtar with NULL, and free() atexit()

gperciva:


tar/bsdtar.c
```diff
@@ -122,10 +125,65 @@ static void		 version(void);
...
+	bsdtar->gname_cache = NULL;
+	bsdtar->buff = NULL;
+	bsdtar->matching = NULL;
+	bsdtar->security = NULL;
+	bsdtar->uname_cache = NULL;
+	bsdtar->siginfo = NULL;
+	bsdtar->substitution = NULL;
+
+	if (atexit(bsdtar_atexit))
+		bsdtar_errc(bsdtar, 1, 0, "Cannot register atexit");
```
cperciva: Is it safe to call bsdtar_errc at this point?

tar/bsdtar.c
```diff
@@ -143,7 +201,7 @@ main(int argc, char **argv)
 	 * for ease of cleanup.
```
cperciva: This isn't stack-allocated any more.

tar/bsdtar.c
```diff
@@ -143,7 +201,7 @@ main(int argc, char **argv)
...
 	 */
 	bsdtar = &bsdtar_storage;
-	memset(bsdtar, 0, sizeof(*bsdtar));
+	bsdtar_init(bsdtar);
```
cperciva: Might be cleaner to make bsdtar_init return a pointer so that this becomes

    bsdtar = bsdtar_init();

tar/bsdtar.c
```diff
@@ -122,10 +125,65 @@ static void		 version(void);
...
+}
+
+static void
+bsdtar_atexit(void)
+{
+	struct bsdtar *bsdtar;
+
+	bsdtar = &bsdtar_storage;
+
+	/* These come from malloc(). */
```
cperciva: "Free arrays allocated by malloc."

tar/bsdtar.c
```diff
@@ -122,10 +125,65 @@ static void		 version(void);
...
+	free(bsdtar->tapenames);
+	free(bsdtar->configfiles);
+
+	/* These come from strdup(). */
```
cperciva: "Free strings allocated by strdup."

tar/bsdtar.c
gperciva: crap!  `bsdtar->progname`, right?  Should I initialize that to `""`, or just use an `fprintf() ; exit;` here?

gperciva: I'll go with my own exit code

gperciva:
PR updated.  I wrote my own "exit" code if there was a problem with `atexit` after initializing bsdtar.  Instead of that, I could set `bsdtar->progname` to something and call `bsdtar_errc()`.

tar/bsdtar.c
```diff
@@ -122,10 +125,72 @@ static void		 version(void);
...
+	bsdtar->siginfo = NULL;
+	bsdtar->substitution = NULL;
+
+	/* We don't have bsdtar->progname yet, so we can't use bsdtar_errc. */
+	if (atexit(bsdtar_atexit)) {
+		fprintf(stderr, "tarsnap: Could not register atexit.\n");
+		exit(1);
+	}
+
+	return bsdtar;
```
cperciva: FreeBSD style(9):
     Values in return statements should be enclosed in parentheses.

tar/bsdtar.c
```diff
@@ -122,10 +125,72 @@ static void		 version(void);
...
+	bsdtar->buff = NULL;
+	bsdtar->matching = NULL;
+	bsdtar->security = NULL;
+	bsdtar->uname_cache = NULL;
+	bsdtar->siginfo = NULL;
+	bsdtar->substitution = NULL;
+
+	/* We don't have bsdtar->progname yet, so we can't use bsdtar_errc. */
+	if (atexit(bsdtar_atexit)) {
+		fprintf(stderr, "tarsnap: Could not register atexit.\n");
```
cperciva: This works.  Another option would have been to set bsdtar->progname to "tarsnap" here, and then simplify lines 217-219 below.  I'm not sure which I consider cleaner, so let's stick with the version you have here.

gperciva: FWIW, I think it's cleaner to initialize `bsdtar->progname` inside `bsdtar_init` -- IMO, the returned value from an "init" function should be ready to use.  But I agree with sticking with this version for now, since I'm sick of seeing lots of junk in valgrind when I test other things.  :)

gperciva:
updated.  (I'm not certain if you get auto-notified when I push to a PR)

tar/bsdtar.c
```diff
@@ -122,10 +125,75 @@ static void		 version(void);
...
+	/* Free arrays allocated by malloc. */
+	free(bsdtar->tapenames);
+	free(bsdtar->configfiles);
+
+	/* Free strings allocated by strdup. */
+	free(bsdtar->cachedir);
+	free(bsdtar->homedir);
+}
+
+
```
cperciva: One empty line between functions, not two.

gperciva:
updated

PR #44, opened 2015-08-25
Avoid reading uninitialized memory during -x

gperciva:


cperciva:
Correct fix, incorrect commit message.  The problem isn't that a->path_safe is being accessed; it's that a->path_safe was never initialized.  So the fix is "initialize no-nasty-symlinks safe path".  (Remember, valgrind is a tool but not a very smart one.  It correctly identified that there was a bug, but you have to understand the code to figure out what the bug is.)

This would be a security problem except that I'm pretty sure this is allocated too soon for anyone to get anything nasty into a newly allocated buffer.  But please check if libarchive still has this bug, since this is the sort of thing which absolutely needs to be upstreamed.

gperciva:
Patch updated.  I'll check upstream later today.

gperciva:
Wait a bit; I'm not certain this is the right spot to put that fix.  Checking how things are done upstream.

gperciva:
I found a different fix upstream, and figured it would be better to use theirs.
https://github.com/Tarsnap/tarsnap/pull/53

PR #46, opened 2015-08-25
Memory fixes

gperciva:
For tarsnap_mode_c(): We need two independent clean-up paths in the code.  Other than moving code into the

    if (failure) {
        ...
    }

parts directly (which might not be terrible for the ccache_write() failure), I couldn't see a cleaner way of doing this.  However, I realize that two distinct clean-up paths is not usual for tarsnap, and the error labels might need renaming.  I didn't find any inspiration from looking at other `goto` labels in tarsnap.

### DIFF
--- a/tar/write.c
+++ b/tar/write.c
@@ -284,22 +284,28 @@ tarsnap_mode_c(struct bsdtar *bsdtar)
 	    bsdtar->argc_orig, bsdtar->argv_orig,
 	    bsdtar->option_print_stats, bsdtar->option_dryrun,
 	    bsdtar->creationtime);
-	if (bsdtar->write_cookie == NULL)
-		bsdtar_errc(bsdtar, 1, 0, "%s", archive_error_string(a));
+	if (bsdtar->write_cookie == NULL) {
+		bsdtar_warnc(bsdtar, 0, "%s", archive_error_string(a));
+		goto err_pre_write_archive;
+	}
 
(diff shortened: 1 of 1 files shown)


lib/netproto/netproto_connect.c
```diff
@@ -105,11 +116,16 @@ callback_cancel(void * cookie)
 static struct sock_addr **
 getserveraddr(void)
 {
-	static struct sock_addr ** srv_addr = NULL;
 	static time_t srv_time = (time_t)(-1);
 	struct sock_addr ** tmp_addr;
 	time_t tmp_time;
 
+	/* This should only happen once. */
```
cperciva: To make the code simpler, how about moving this...

cperciva:
Can you split the write.c bit off into a separate PR?  It's less obvious and I want to review it more later.

lib/netproto/netproto_connect.c
gperciva: I wasn't certain if there was any multi-threading issues here.  I mean, how long can `sock_resolve` take?  with the callbacks, I wasn't certain if you might call `getserveraddr` multiple times (before the first had finished).

However, I see that you don't use `pthread.h`, so I'll go ahead and make this change.  :)

gperciva:
Most of these commits were already accepted; I'm working on a better version of the final one.

PR #47, opened 2015-08-26
FIXME Free more multitape memory

gperciva:
(this PR needs attention from @cperciva, but it likely needs at least one more draft before being suitable for pushing)

This allows
    tarsnap -c --dry-run ~/whatever/
without any valgrind-reported memory leaks.
(this leak also appears with "tarsnap -c -f name ~/whatever")

In particular, there are 3 leaks, all in
tar/multitape/multitape_write.c

- 2 occur on line 203:
      index_new = realloc(S->index, indexalloc_new);
  (called via different callbacks)
- 1 is on line 559:
      hbuf_new = realloc(d->hbuf, hbufalloc_new);

FIXME: this patch does a few more "free()" than strictly necessary
to resolve this leak.  I figured that there was no harm in
free(NULL), but I could create a "minimal" patch instead.

### DIFF
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@ -898,6 +898,10 @@ writetape_close(TAPE_W * d)
 	chunkify_free(d->t.c);
 	chunkify_free(d->c.c);
 	chunkify_free(d->h.c);
+	free(d->t.index);
+	free(d->c.index);
+	free(d->h.index);
+	free(d->hbuf);
 	free(d->cachedir);
 	free(d->tapename);
 	free(d);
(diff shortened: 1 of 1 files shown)


cperciva:
Is this coming back with a new patch later?  I really did plan on looking at it...

gperciva:
The patch exists on a new branch:
https://github.com/Tarsnap/tarsnap/tree/free-multitape-memory
(although the only change is the commit message)

I was going to ask you tonight if you wanted me to rewrite multitape_open() to
use the bsdtar-init-free approach.  Namely, allocate it and write NULLs to
pointers, then have a writetape_free() that called free() on all pointers.
(ok, we didn't do the second part with bsdtar).  That would make
writetape_open() much easier to read: instead of 7 distinct error labels,
there would only be 2 or 3.

Also, given that not all platforms use 0s for NULL, that patch is slightly
dangerous in that does free(d->t.index) regardless of whether that was
allocated or not.

- Graham

PR #50, opened 2015-08-29
free(NULL) no-op

gperciva:


libcperciva/crypto/crypto_aes.c
```diff
@@ -151,6 +151,10 @@ void
 crypto_aes_key_free(struct crypto_aes_key * key)
 {
 
+	/* Behave consistently with free(NULL). */
```
cperciva: This should be after the `#ifdef CPUSUPPORT_X86_AESNI` block.

tar/ccache/ccache_read.c
```diff
@@ -162,6 +162,10 @@ callback_free(void * cookie, uint8_t * s, size_t slen, void * rec)
 {
 	struct ccache_record * ccr = rec;
 
+	/* Behave consistently with free(NULL). */
```
cperciva: Not here; this isn't a nameofdatastructure_free function.

tar/chunks/chunks_directory.c
```diff
@@ -82,6 +82,10 @@ callback_free(void * rec, void * cookie)
 {
 	struct chunkdata * ch = rec;
 
+	/* Behave consistently with free(NULL). */
```
cperciva: Not a _free function.

tar/storage/storage_read.c
```diff
@@ -613,6 +613,10 @@ callback_cache_free(void * record, void * cookie)
 {
 	struct read_file_cached * CF = record;
 
+	/* Behave consistently with free(NULL). */
```
cperciva: Not a _free function.

gperciva:
Updated

PR #53, opened 2015-08-29
Do not refer an uninitialized or empty symlink name.

gperciva:
Imported from libarchive v3.0.2.
https://github.com/libarchive/libarchive/commit/8e0dd6534ecc99ae4c53b58361e0591c64848d87
libarchive-SVN-Revision: 3909

gperciva: I renamed the file from

    libarchive/archive_write_disk_posix.c

to

    libarchive/archive_write_disk.c

but the patch is otherwise identical.

### DIFF
--- a/libarchive/archive_write_disk.c
+++ b/libarchive/archive_write_disk.c
@@ -1402,7 +1402,7 @@ current_fixup(struct archive_write_disk *a, const char *pathname)
 static int
 check_symlinks(struct archive_write_disk *a)
 {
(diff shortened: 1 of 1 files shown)


gperciva:
wait, I lost the date on this comit

gperciva:
ok, good to go.  I retained as much info from the original commit as possible -- author, date, grammatical mistake in the subject link, etc.  I figured that might make it easier when we update to a newer libarchive.

PR #55, opened 2015-08-30
Free DNS cache sock_addr with atexit()

gperciva:
Split off from https://github.com/Tarsnap/tarsnap/pull/52

### DIFF
--- a/lib/netproto/netproto_connect.c
+++ b/lib/netproto/netproto_connect.c
@@ -24,6 +24,16 @@ struct netproto_connect_cookie {
 	NETPROTO_CONNECTION * NC;
 };
 
+static struct sock_addr ** srv_addr = NULL;
+static void netproto_connect_atexit(void);
+
+static void
+netproto_connect_atexit(void)
+{
+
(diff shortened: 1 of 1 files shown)


lib/netproto/netproto_connect.c
```diff
@@ -129,6 +138,14 @@ getserveraddr(void)
 
 	/* If we have a new lookup, update the cache. */
 	if (tmp_addr != NULL) {
+
```
cperciva: No blank line here.  Yes, I know I said blank lines before comments.  But it's actually blank lines separating code from comments, and this is the start of a block so there's no preceding code at this indent level...

lib/netproto/netproto_connect.c
```diff
@@ -129,6 +138,14 @@ getserveraddr(void)
...
+		/* First time we run getserveraddr, register atexit. */
```
cperciva: This isn't necessarily the first time we run getserveraddr.  It's the first time we have an address to cache though.

PR #58, opened 2015-09-02
Tarsnap_mode_c() error path

gperciva:


tar/write.c
```diff
@@ -284,38 +285,63 @@ tarsnap_mode_c(struct bsdtar *bsdtar)
...
 	/* Free the chunkification cache. */
 	if (bsdtar->cachecrunch < 2)
 		ccache_free(bsdtar->chunk_cache);
+
+	/* Success */
+	return;
+
+err1:
+	ccache_free(bsdtar->chunk_cache);
+err0:
```
cperciva: You should have err2 and err1 here.  I reserve err0 for "no cleanup needed" error path gotos.

gperciva: done.  I could have added an err0 for the initial malloc erroc, but since you didn't mention that, I left it as `bsdtar_errc`.  I'll obviously change that too if you'd like.

in favor of changing: that seems to match other tarsnap style better (looking at random networking and crypto files)

against changing: we've got the `bsdtar` pointer, so we might as well call `bsdtar_errc` directly.

gperciva: no wait, I just looked at the actual diff of the patch.  (I'd forgotten what was there originally)

I'm changing the malloc failure to `goto err0`.

PR #59, opened 2015-09-04
avoid some memory operations with NULL

gperciva:
These arose from the clang static analyzer.

### DIFF
--- a/lib/network/tsnetwork_writeq.c
+++ b/lib/network/tsnetwork_writeq.c
@@ -2,6 +2,7 @@
 
 #include <sys/time.h>
 
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


lib/network/tsnetwork_writeq.c
```diff
@@ -35,6 +36,11 @@ dowrite(struct network_writeq_internal * Q)
 	struct network_writeq_buf * QB = Q->head;
 	struct timeval timeo;
 
+	/* Sanity checks to avoid memcpy(dest, NULL, num) */
```
cperciva: How do we (as developers) know that these assertions should pass?  "Because the code would crash otherwise" isn't a reason; it's just wishful thinking.

As it happens, two of these assertions can be optimized away by the compiler -- we've already dereferenced Q, and &foo is never equal to NULL.

But asserting that we can only get here if the queue is non-empty... that makes sense.

gperciva:
updated

PR #60, opened 2015-09-05
Add assert() before asprintf() involving "%s"

gperciva:
This does not add assert() for asprintf() calls which are protected with an
    if (var != NULL)
check.

### DIFF
--- a/lib/keyfile/keyfile.c
+++ b/lib/keyfile/keyfile.c
@@ -3,6 +3,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
+#include <assert.h>
 #include <fcntl.h>
 #include <stdint.h>
 #include <stdlib.h>
@@ -121,6 +122,9 @@ read_encrypted(const uint8_t * keybuf, size_t keylen, uint64_t * machinenum,
 	uint8_t * deckeybuf;
 	size_t deckeylen;
 	int rc;
+
+	/* The caller must pass a file name to be read. */
+	assert(filename != NULL);
 
 	/* Sanity-check size. */
 	if (keylen == 0) {
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@ -1,5 +1,6 @@
 #include "bsdtar_platform.h"
 
+#include <assert.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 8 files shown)


cperciva:
I think the asserts are all good, but the comments are wrong.  You need to explain why the assertions should hold, not state what would go wrong if they don't.

gperciva:
Hmm.  Do you still want assertions here, or would an if... goto err0; be better?

For example, suppose that
```
static int
read_encrypted(const uint8_t * keybuf, size_t keylen, uint64_t * machinenum,
    const char * filename, int keys)
```
is called with `filename == NULL`.  Would you like to generate an assert failure, or return a `-1` ?

(at first glance, 90% of these are filenames, 8% are suffices, and one is `d->tapename`.)

cperciva:
If a value should be non-NULL, assert that it's non-NULL.  Programmer errors get asserts; returning -1 is for if something went wrong in the execution, not for when there's something wrong with the code.

gperciva:
updated

lib/keyfile/keyfile.c
```diff
@@ -122,6 +123,9 @@ read_encrypted(const uint8_t * keybuf, size_t keylen, uint64_t * machinenum,
 	size_t deckeylen;
 	int rc;
 
+	/* It doesn't make sense to read an empty filename. */
+	assert(filename != NULL);
```
cperciva: NULL != "empty".  I'd go with something along the lines of "The caller must pass a file name to be read."

tar/multitape/multitape_write.c
```diff
@@ -715,6 +716,9 @@ flushtape(TAPE_W * d, int isapart, int extrastats)
 	struct tapemetaindex tmi;
 	char * tapename;
 
+	/* We need a tapename. */
```
cperciva: I'd add "(anonymous dry runs are assigned a fake name before this point)" to avoid confusion about that.

gperciva:
updated

tar/multitape/multitape_write.c
```diff
@@ -715,6 +716,12 @@ flushtape(TAPE_W * d, int isapart, int extrastats)
 	struct tapemetaindex tmi;
 	char * tapename;
 
+	/*
+	 * We need a tapename.  (anonymous dry runs are assigned a fake name
```
cperciva: Err, capitalization?  Punctuation?

gperciva:
updated

Issue #34, opened 2015-08-14
storage_write_start does dry runs a bit weirdly


cperciva:
We should probably change the `storage_write_start` API from "If ${dryrun} is nonzero, perform a dry run (which ignores ${lastseq} and sets ${seqnum} to 32 0s)." to "If ${dryrun} is nonzero, ignore ${lastseq} and ${seqnum}".  Or possibly push dryrun resolution higher in the stack even -- don't create a storage cookie at all, but instead just have the multitape and chunks layers be aware of dryruns and not issue calls to the storage layer?

But storage_write_start is doing a silly amount of pointless initialization given that a dry-run storage cookie should never do anything.  There's got to be a better way to organize this code.

gperciva:
Addressed in #61.

PR #61, opened 2015-09-05
Dryrun no storage cookie

gperciva:
This addresses #34.

### DIFF
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@ -56,10 +56,9 @@ struct multitape_write_internal {
 	char ** argv;		/* Command-line arguments. */
 	int stats_enabled;	/* Stats printed on close. */
 	int eof;		/* Tape is truncated at current position. */
-	int dryrun;		/* A dry run is being performed. */
 
 	/* Lower level cookies. */
-	STORAGE_W * S;		/* Storage layer write cookie. */
+	STORAGE_W * S;		/* Storage layer write cookie; NULL=dryrun. */
 	CHUNKS_W * C;		/* Chunk layer write cookie. */
 	int lockfd;		/* Lock on cache directory. */
 	uint8_t seqnum[32];	/* Transaction sequence number. */
@@ -440,9 +439,6 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	/* Record whether we should print archive statistics on close. */
 	d->stats_enabled = printstats;
 
-	/* Record whether this is a dry run. */
-	d->dryrun = dryrun;
-
 	/* If we're using a cache, make sure ${cachedir} exists. */
 	if ((cachedir != NULL) && (dirutil_needdir(cachedir)))
 		goto err3;
@@ -452,11 +448,11 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 		goto err3;
 
 	/* If this isn't a dry run, finish any pending commit. */
-	if ((d->dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
+	if ((dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
 		goto err4;
 
 	/* If this isn't a dry run, get the sequence number. */
-	if ((d->dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
+	if ((dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
 		goto err4;
 
 	/*
@@ -464,7 +460,7 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	 * layer.  If it is a dry run, set the storage cookie to NULL to
 	 * denote this fact.
 	 */
-	if (d->dryrun == 0) {
+	if (dryrun == 0) {
 		if ((d->S = storage_write_start(machinenum, lastseq,
 		    d->seqnum)) == NULL)
 			goto err4;
@@ -480,9 +476,9 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	 * the specified name or that plus ".part" (in case the user decides
 	 * to truncate the archive).
 	 */
-	if ((d->dryrun == 0) && tapepresent(d->S, "%s", tapename))
+	if (tapepresent(d->S, "%s", tapename))
 		goto err6;
-	if ((d->dryrun == 0) && tapepresent(d->S, "%s.part", tapename))
+	if (tapepresent(d->S, "%s.part", tapename))
 		goto err6;
 
 	/* Initialize streams. */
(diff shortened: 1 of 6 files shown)


cperciva:
Are you sure we're never accessing ->S when it is uninitialized now?

Also, I'm wondering if we want to have a "dryrun" parameter at all or if it can be mapped to "S == NULL".  Do you have any sense if this would make the code more or less clean?

gperciva:
First question: sure about never accessing `->S` ?  no.

Hmm, ```tar/multitape/multitape_write.c``` contains places that call `storage_write_free()`, but those would die on the non-0 `NULL` platforms.  Should I add `d->S = NULL` to the initializer?  if so, should I only do it for `d->S` or should I properly initialize the other pointers as well?

I'm confident (which is less than "sure" or "certain") that it's safe in ```tar/chunks/chunks_write.c``` because the only use is now:
```
/* If not a dry run, ask the storage layer to write the file for us. */
if ((C->dryrun == 0) && (storage_write_file(C->S, C->zbuf, zlen, 'c', hash))) {
```

Other than the "non-NULL platforms" issue, I'm also confident about
```
tar/multitape/multitape_write.c
```
because the only uses are:

* Instances where it's protected with `d->dryrun == 0`, such as:
  ```
  if ((d->dryrun == 0) && ((d->S = storage_write_start(machinenum,
  ```

* initializer of chunks_write:
  ```
  if ((d->C = chunks_write_start(cachedir, d->S, MAXCHUNK, d->dryrun)) == NULL)
  ```
  which might be dangerous, but this just records `C->S = S` inside the `chunk_write` code, and I'm confident about that.

* places that call `storage_write_free()`, which I mentioned at the top.

There are some uses of `->S` inside `chunks_delete.c` and `chunks_read.c`, but those appear to be STORAGE_D and STORAGE_R respectively.

```
tar/multitape/multitape_delete.c
```
uses `->S` for a STORAGE_R cookie (?)... and inside `deletetape()`, we have `d->S` being STORAGE_R and `S` being STORAGE_D.

```
tar/multitape/multitape_read.c
```
uses `->S` for a STORAGE_R cookie, so that's fine.

```
tar/multitape/multitape_stats.c
```
contains the string `->S` but blessedly it's `->SR` so I can easily see that they're all STORAGE_R.  Praise to His Noodly Appendages!

[would you consider a patch that replaced `->S` with the appropriate `->SR` / `->SW` / `->SD`?]

```
tar/storage/
```
naturally contains `->S`, but those are only called by `multitape_write.c` when `d->dryrun == 0`.

There's a bunch of `gds->Seconds` in `tar/getdate.c`, but of course that has nothing to do with this.

If I've missed anything, I'm very curious about what it is!


Second question: I think it's easier to reason about the code when there's a `if ((d->dryrun == 0) ...` rather than just using `(d->S == NULL)`.  In terms of readability, there's a few places that I think could be put inside a big `if (d->dryrun ==0)`, such as:

```
if ((d->dryrun == 0) && multitape_cleanstate(cachedir, machinenum, 0))
    ...
if ((d->dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
    ...
if ((d->dryrun == 0) && ((d->S = storage_write_start(machinenum,
    ...
```
but that wasn't quite what you were asking.

tar/multitape/multitape_write.c
```diff
@@ -747,22 +752,26 @@ flushtape(TAPE_W * d, int isapart, int extrastats)
...
-	 * Store archive metaindex.  Note that this must be done before the
-	 * archive metadata is stored, since it fills in fields in the archive
-	 * metadata concerning the index length and hash.
+	 * If not a dry run, store archive metaindex.  Note that this must be
+	 * done before the archive metadata is stored, since it fills in
+	 * fields in the archive metadata concerning the index length and
+	 * hash.
 	 */
-	if (multitape_metaindex_put(d->S, d->C, &tmi, &tmd, extrastats))
+	if ((d->dryrun == 0) && (multitape_metaindex_put(d->S, d->C, &tmi, &tmd, extrastats)))
```
cperciva: Doesn't this make the statistics from `tarsnap -c --dry-run --print-stats` wrong?  We need to create the metaindex and metadata in order to account for them in the statistics, even if we're not going to upload them.

gperciva:
(sorry, let me try that again in a more email-friendly way)

I added an `int dryrun` parameter to `multitape_metadata_put()` and `multitape_metaindex_put()`, but I'm uncertain about that API change.

Argument for adding `int dryrun`: it's easier for the programmer to see (and reason about) what's happening.
Argument for simply relying on `S==NULL` to indicate a dry run: it saves memory? it's slightly "neater" in terms of avoiding unnecessary variables?

gperciva:
(still a draft, it's not worthy of serious examination.  A bit more cleanup in `multitape_write.c` to come)

Yes, it looks much cleaner now relying on `S == NULL`.

tar/multitape/multitape_write.c
```diff
@@ -461,10 +461,18 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 	if ((d->dryrun == 0) && (multitape_sequence(cachedir, lastseq)))
 		goto err4;
 
-	/* Obtain write cookies from the storage and chunk layers. */
-	if ((d->S = storage_write_start(machinenum, lastseq,
-	    d->seqnum, d->dryrun)) == NULL)
-		goto err4;
+	/*
+	 * If this isn't a dry run, obtain a write cookie from the storage
+	 * layer.
```
cperciva: + "If it is a dry run, set the storage cookie to NULL to denote this fact."

tar/storage/storage_write.c
```diff
@@ -157,15 +150,11 @@ storage_write_start(uint64_t machinenum, const uint8_t lastseq[32],
 	}
 
 	/*
-	 * If this isn't a dry run, start a write transaction.
-	 * Otherwise, initialize the nonce to 0.
+	 * Start a write transaction.
```
cperciva: /* One-line comments like this. */

tar/storage/storage_write.c
```diff
@@ -306,8 +296,8 @@ storage_write_file(STORAGE_W * S, uint8_t * buf, size_t len,
 {
 	struct write_file_internal  * C;
 
-	/* If this is a dry run, return without doing anything. */
-	if (S->dryrun)
+	if (S == NULL)
+		/* Success for a dry run. */
```
cperciva: If you put a comment here, this becomes a multi-line if and you need braces.  But I would put the comment before the if.

Also, all we know here is that S is NULL.  It might be a dry run, or it might be for some completely other purpose.

gperciva:
What do you have in mind for long comments inside a struct?  This patches does

```
-	STORAGE_W * S;		/* Storage layer write cookie. */
+	STORAGE_W * S;		/* Storage layer write cookie; NULL if this is a dry run. */
```

but that exceeds 80 chars.  (even making the latter part "NULL for a dry run." still makes the line 85 chars)

gperciva:
slightly refactored to improve atomic-ness of commits (and commit messages)

gperciva:
updated

PR #63, opened 2015-09-07
Backport libarchive strdup allocation failure check

gperciva:
This is extracted from a much larger commit:
https://github.com/libarchive/libarchive/commit/40722a255fe8f25a6a8dcd9d66fda49478756b17#diff-0cef2313059c42005d9a2fcaf32023ddR231

(with a few changes for it to make sense with the Tarsnap libarchive code)

### DIFF
--- a/libarchive/archive_read_support_compression_program.c
+++ b/libarchive/archive_read_support_compression_program.c
@@ -154,8 +154,10 @@ archive_read_support_compression_program_signature(struct archive *_a,
 	 */
 	state = (struct program_bidder *)calloc(sizeof (*state), 1);
 	if (state == NULL)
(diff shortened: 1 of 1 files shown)


libarchive/archive_read_support_compression_program.c
```diff
@@ -156,6 +156,8 @@ archive_read_support_compression_program_signature(struct archive *_a,
 	if (state == NULL)
 		return (ARCHIVE_FATAL);
```
cperciva: Can we make this 'goto memerr' too?  I think that more closely matches the upstream anyway?

gperciva:
updated

PR #64, opened 2015-09-08
Keygen refactor network

gperciva:
This is the first half of the refactoring, broken up into minimal steps for hopefully-easier reviewing (and checking in the git history if necessary).


cperciva:
Don't use separate `keygen_data.h` and `keygen_network.h` files.  Just make it one `keygen.h` file.  The only point of using lots of small headers is to avoid link pollution (if you include a header file which declares a function, you theoretically need to link to the C file which implements it, even if you never actually use that function) but that's not an issue here.

gperciva:
updated

gperciva:
updated; I removed the unnecessary `keygen_network.h`, which was added in commit 2 and removed in commit 4.

gperciva:
Crap, I forgot to re-open this before pushing again.  :(

Continued in #70.

PR #67, opened 2015-09-18
Warn about creating an archive with 0 bytes

gperciva:
The "0 bytes" is not fooled by deduplication; creating two archives with
exactly the same files (so the second contains 0 new data) will not trigger
this warning.  This is because archive_position_compressed(a) returns
a->raw_position, which is advanced by archive_write_skip(a, s).

### DIFF
--- a/tar/write.c
+++ b/tar/write.c
@@ -450,6 +450,16 @@ write_archive(struct archive *a, struct bsdtar *bsdtar)
 		archive_entry_linkify(bsdtar->resolver, &entry, &sparse_entry);
 	}
 
+	/*
+	 * Check whether the archive is empty; this is unaffected by
+	 * deduplication.  Must be done before the archive is closed since
+	 * even empty archives end up with nonzero size due to tar EOF blocks.
+	 */
+	if ((!bsdtar->option_quiet) &&
+	    (archive_position_compressed(a) == 0)) {
+		bsdtar_warnc(bsdtar, 0, "Warning: Archive contains no files");
+	}
+
 	if (archive_write_close(a)) {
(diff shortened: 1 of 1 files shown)


gperciva:
libarchive 2.8.0 and 3.0.0 added `archive_file_count(a)`, which make this code slightly easier to read, but unfortunately we're not using those versions.  The patch itself looks fairly easy, though, so I could try backporting it: https://github.com/libarchive/libarchive/commit/ab649de78644b55fcb2e1fd5b056635c895dc4b0
That said, there was a bugfix here: https://github.com/libarchive/libarchive/commit/6f334d42006e25ad1a17805c8d527e90c2074a25 and possibly others.

tar/write.c
```diff
@@ -450,6 +450,15 @@ write_archive(struct archive *a, struct bsdtar *bsdtar)
 		archive_entry_linkify(bsdtar->resolver, &entry, &sparse_entry);
 	}
 
+	/*
+	 * Check whether the archive contains 0 bytes; this is unaffected by
```
cperciva: s/contains 0 bytes/is empty/

tar/write.c
```diff
@@ -450,6 +450,15 @@ write_archive(struct archive *a, struct bsdtar *bsdtar)
...
+	 * deduplication.  Must be done before the archive is closed.
```
cperciva: s/.$/ since even empty archives end up with nonzero size due to tar EOF blocks./

gperciva:
updated

gperciva:
updated with your second comment (about EOF).

Issue #62, opened 2015-09-06
check for keys doesn't account for @@archive


cperciva:
When we're checking to make sure we have all the keys we need, we're assuming that archive creation only requires write keys.  But if there's an @@archive directive, we will need read keys as well, which means we could end up at "Programmer error: key not available".

I'm not sure what the best solution is for this; scanning the command line for @@archive directives?  We need to be careful to not trigger on "-C @@foo" in case someone has a peculiarly named directory though.  We could postpone the check until we actually hit the @@archive directive, but it would be much better if we could inform the user that they're using the wrong key file promptly...

gperciva:
Note to self: assuming you have the file `ok.txt` in archive `a0`, reproduce with:
```
./tarsnap -c --dry-run --include "ok.txt" @@a0
```

gperciva:
Correction for testing:
```
mkdir @@blah
touch @@blah/foo.txt
./tarsnap -c --dry-run @@a0 -C @@blah foo.txt
```

gperciva:
Should be addressed in #68.

PR #68, opened 2015-09-22
Warn permissions

gperciva:
-c normally only requires write keys, but for `@@archive`, it requires read as
well.  This scans the modified argv (after processing normal options) for any
"@@" (in positions [0] and [1]), but it doesn't get fooled by "-C @@*".

Addresses #62.

This was tested with:
```
./tarsnap -c --dry-run @@a0
./tarsnap -c --dry-run @@a0 -C @@blah foo.txt
./tarsnap -c --dry-run @@a0 -C@@blah foo.txt
./tarsnap -c --dry-run -C @@blah foo.txt @@a0
./tarsnap -c --dry-run -C@@blah foo.txt @@a0
./tarsnap -c --dry-run -C @@blah foo.txt
./tarsnap -c --dry-run -C@@blah foo.txt
```

I highly suspect that you'd rather have this function in a separate file, or just generally to make it more "C-like" instead of "python-like".  But I figured that an initial draft was better than nothing.

Also, this produces the message:
```
$ ./tarsnap -c --dry-run @@a0
tarsnap: The read authorization key is required for -c but is not available
```

which might be slightly un-intuitive.  Or maybe it's ok.  Problem is, I don't want to display a warning whenever -c involves `@@a0` since there's not going to be any problem as long as they have read access as well.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -121,6 +121,7 @@ static void		 optq_push(struct bsdtar *, const char *,
 static void		 optq_pop(struct bsdtar *);
 static void		 set_mode(struct bsdtar *, int opt, const char *optstr);
 static void		 version(void);
+static int		 argv_has_archive_directive(struct bsdtar *bsdtar);
 
 /* A basic set of security flags to request from libarchive. */
 #define	SECURITY					\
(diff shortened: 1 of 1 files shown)


gperciva:
updated

gperciva:
right, we don't need to check for `"-C@@"` since the `@@` isn't in positions `[0] [1]`.  PR updated.

PR #70, opened 2015-09-24
Keygen refactor network

gperciva:
Crap, I forgot to re-open the old PR before pushing again.  :(

Previous discussion https://github.com/Tarsnap/tarsnap/pull/64

I've checked that every commit compiles (including running `autoreconf` and `configure` at each step)


gperciva:
updated according to latest comments.

PR #71, opened 2015-09-25
Compiler warnings

gperciva:
This backports 3 commits from libarchive, then enables -Wall -Wextra -Werror.

I know you don't like playing silly games with `(void)unused_var;` just to satisfy silly warnings about ignoring the returned value of `write()`, but we could have avoided https://github.com/Tarsnap/tarsnap/pull/69 if we had `-Werror`.  Also, I would feel better if I knew that my patches didn't cause any warnings, but when we have known warnings in the build, we tend to ignore anything that flashes on the screen (especially since they end up buried by other compile messages).

Any chance of this getting in?  Note that I have not directly modified any tarsnap `.c` code in order to achieve `-Werror` success; the only changes came from upstream libarchive.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,6 +2,8 @@
 
 AUTOMAKE_OPTIONS= foreign subdir-objects
 
+CFLAGS += -Wall -Wextra -Wno-unused-function
+
 #
 # What to build and install
 #
--- a/libarchive/archive_check_magic.c
+++ b/libarchive/archive_check_magic.c
@@ -69,8 +69,7 @@ diediedie(void)
 	/* Cause a breakpoint exception  */
 	DebugBreak();
 #endif
-	*(volatile char *)0 = 1;	/* Deliberately segfault and force a coredump. */
-	_exit(1);	/* If that didn't work, just exit with an error. */
+	abort();        /* Terminate the program abnormally. */
 }
 
 static const char *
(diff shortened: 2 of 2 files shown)


cperciva:
-Werror should absolutely not be in any shipping code.  Ever.  If you do that, perfectly good code suddenly breaks when gcc adds a new warning message.

Enabling more warnings is good though.  In the pre-git days I had lots of warnings turned on... they came automatically with the FreeBSD makefiles I was using.

BTW I often run `make all clean >/dev/null` in order to see warnings but not the build commands.  `make -s all clean` should also work.

cperciva:
Two of the three merges are good.  The one which adds bogus variables just to silence an idiotic compiler is not good.  If people are running an idiotic compiler they should get annoyed by warning messages.

gperciva:
ok, I'll get used to `make -s`.  Apparently somebody else warned that gcc will steal your cat if you enable -Werror in shipping code, so I'll go along with avoiding that.

gperciva:
(also, PR updated)

gperciva:
well, c26323e depends on 6096d32.  How much do you want to keep with libarchive upstream commits?  I could merge c26323e with the relevant part of 6096d32.

cperciva:
Yes, the final result there is correct.  It just took a weird detour.

I'm not too worried about having commit-for-commit alignment between the tarsnap tree and the libarchive tree.  When we move to a newer libarchive, I'm pretty sure it will be "take our changes and apply them to a newer libarchive", not "merge libarchive updates into tarsnap".

gperciva:
ok, PR adjusted.  There's over a screen of junk with `gcc` even with `make -s`, but I'll take another look at that (or else switch to always typing in `../configure CC=clang`).  At any rate, it's a step forward.

PR #80, opened 2015-10-05
Style scripts

gperciva:
These automate parts of checking for STYLE compliance.

- Checking that `#include` is alphabetical: works well (pending clarification about a possible misunderstanding, see https://github.com/Tarsnap/tarsnap/commit/9119bdda23dd1377c748697b312efe0919a02c44#commitcomment-13598335)
- Checking for newlines in function definitions: creates some false positives and possibly false negatives.  The script prints a warning at the end that it isn't perfect.  I still found it a useful tool.


cperciva:
Please put these into a new top-level /tools/ instead.  I want to keep "stuff used during development" separate from "code which gets installed".

gperciva:
Updated; do you want to have a `tools/release/` and `tools/devel/`, or just `release-tools/` and `tools/`?

cperciva:
Hmm, good question.  Let's leave release-tools where it is for now. I have a feeling we may end up having /tools/release/ and /tools/mdoc2man/ and /tools/style/, but we can always move things around later.

PR #81, opened 2015-10-10
Style: add space in "switch ("

gperciva:
switch is a keyword.  It needs a space after it.

Untouched:
  - libarchive/*
  - tar/util.c
  - tar/write.c

### DIFF
--- a/tar/multitape/multitape_read.c
+++ b/tar/multitape/multitape_read.c
@@ -83,7 +83,7 @@ stream_get_chunkheader(struct stream * S, CHUNKS_R * C)
 		readlen = stream_read(S->istr, (uint8_t *)&S->ch,
 		    sizeof(struct chunkheader), C);
 
-		switch(readlen) {
+		switch (readlen) {
 		case -1:
 			/* Error in stream_read. */
 			goto err0;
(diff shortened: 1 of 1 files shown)


tar/bsdtar.c
```diff
@@ -963,7 +963,7 @@ main(int argc, char **argv)
 	network_bwlimit(bsdtar->bwlimit_rate_down, bsdtar->bwlimit_rate_up);
 
 	/* Perform the requested operation. */
-	switch(bsdtar->mode) {
+	switch (bsdtar->mode) {
```
cperciva: Technically this line is libarchive code.  Can you give me a PR for just the tarsnap code, and perhaps send the libarchive style fix upstream?

gperciva: The libarchive committer guide specifically states: "Do not make unnecessary
changes to existing whitespace, capitalization, or spelling."

Do you think this counts as "necessary" in freebsd-land?  I wouldn't have
thought that it qualified.

cperciva:
In that case don't bother with the libarchive bit.  Last time I was contributing significant code to libarchive it didn't have a contributor's guide. ;-)

PR #84, opened 2015-10-13
Keygen refactor actual

gperciva:
I tried to make this easier to check the individual commits by breaking up the changes, but I may have gone overboard on this.

General plan:
1. refactor keygen.c into keygen.c + keygen_actual.c: keygen.c handles `argv` and calling usage(), then passes data to keygen_actual.c to do everything else.
2. refactor keyregen.c with the same style as keygen.c.
3. add a memory-leak fix that I noticed while testing the above.


gperciva:
updated to `return (-1)` from `keygen_actual()`.

gperciva:
Updated with an adjusted `int keymask`.

gperciva:
(re-upload to trigger travis CI)

PR #85, opened 2015-10-14
Update dates in man pages

gperciva:


gperciva:
this will be replaced by something like https://github.com/Tarsnap/scrypt/pull/17 if that method is good.

Issue #89, opened 2015-11-03
Keyless dry-run finds default config


cperciva:
Not the intended behaviour:
```
$ tarsnap --dry-run --print-stats --humanize-numbers -c /
tarsnap: stat(/root/tarsnap.key): No such file or directory
tarsnap: Cannot read key file: /root/tarsnap.key
```

I think in the command line processing we should record whether a key file or a cache directory were specified *on the command line*; if not (e.g., we have a keyfile set, but it wasn't done on the command line) then we need to add fallback mechanisms for dry runs.

I can't remember the details of cache directories enough to say exactly what needs to be done there without looking at the code, but I think there may also need to be some special "ignore the value which was set in a configuration file" handling there too.

gperciva:
I can't reproduce in an obvious way.  I mean, if I run

    $ tarsnap --dry-run --print-stats --humanize-numbers --no-default-config -c ~/backup

then it's fine.  If I run

    $ tarsnap --dry-run --print-stats --humanize-numbers -c ~/backup

without a `~/.tarsnaprc` then it's fine.  If my `~/.tarsnaprc` is empty (or commented out), then it's also fine.

If `~/.tarsnaprc` points to a non-existent `keyfile` then it complains, but surely that's a PEBCAK ?

Isn't the real problem that somebody ended up with `.../etc/tarsnap.conf` instead of `.../starsnap.conf.sample` ?  and that it wasn't obvious they could add `--no-default-config`?

gperciva:
Thinking "out loud", we don't want to automatically add `--no-default-config` if somebody specifies `--dry-run`, because their config file could usefully specify `--humanize-numbers` and their cache directory.  And we definitely want to use the cache directory if it exists, otherwise the dry run will report that all data would be uploaded (which doesn't change anything in the "pre-signup case", but is definitely not what we want for "post-signup" accounts).

My current thought is that if they specify `--dry-run` and tarsnap encounters an error, it could prompt them to use `--no-default-config` if they haven't signed up.  But the message should be clear that if they *have* signed up, and *have* generated a key, they shouldn't ignore the error.

Also, we should make sure that the tarsnap installation process doesn't end up with an unusable system.  `make install` installs `tarsnap.conf.sample` as such, so maybe we should send a polite request to whichever packager did this?  I've seen at least one few post-install script helpfully rename `tarsnap.conf.sample` to `tarsnap.conf` (to be fair, using tarsnap at all without a keyfile is a new feature).

[...]

PR #91, opened 2015-11-09
Dryrun keyless ignore config

gperciva:
Addresses #89 

I went back and forth half a dozen times between handling `fromconffile` in `load_keys()` vs. `dooption()`.  In the end, I went with `load_keys()` since `dooption()` is already pretty long, and each option isn't very long.  But I wimped out on doing *everything* in `load_keys()`; I left `bsdtar->have_keys = 1` in `dooption()` since that's where it was originally.  I think it might be better to move that into `load_keys()`; if we did that, `load_keys()` could go back to being `void` instead of the `int` that I changed it to.

Basically: adding the functionality was trivial.  Trying to make it match the desired style was not.  :|

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -112,7 +112,7 @@ static int		 configfile_helper(struct bsdtar *bsdtar,
 			     const char *line);
 static void		 dooption(struct bsdtar *, const char *,
 			     const char *, int);
-static void		 load_keys(struct bsdtar *, const char *path);
+static int		 load_keys(struct bsdtar *, const char *path);
 static void		 long_help(struct bsdtar *);
 static void		 only_mode(struct bsdtar *, const char *opt,
 			     const char *valid);
@@ -867,6 +867,19 @@ main(int argc, char **argv)
 		only_mode(bsdtar, "--strip-components", "xt");
 
 	/*
+	 * If the keyfile in the config file is invalid but we're doing a
+	 * dryrun, continue anyway (and don't use a cachedir).
+	 */
+	if (bsdtar->config_file_keyfile_failed && bsdtar->option_dryrun &&
+	    bsdtar->cachedir != NULL) {
+		bsdtar_warnc(bsdtar, 0,
+		    "Ignoring cachedir due to missing or invalid "
+		    "keyfile in config file.");
+		free(bsdtar->cachedir);
+		bsdtar->cachedir = NULL;
+	}
+
+	/*
 	 * Canonicalize the path to the cache directories.  This is
 	 * necessary since the tar code can change directories.
 	 */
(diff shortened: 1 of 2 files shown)


lib/keyfile/keyfile.c
```diff
@@ -374,6 +376,9 @@ keyfile_read(const char * filename, uint64_t * machinenum, int keys)
 		goto err0;
```
cperciva: This seems like it's begging for a `if (errno == ENOENT) goto notfound;` ?

tar/bsdtar.c
```diff
@@ -1687,16 +1702,33 @@ dooption(struct bsdtar *bsdtar, const char * conf_opt,
...
+	int rc = 1;
 
 	/* Load the key file. */
-	if (keyfile_read(path, &machinenum, ~0))
+	switch (keyfile_read(path, &machinenum, ~0)) {
+	case 0:
+		break;
+	case 1: /* file not found */
+		if (fromconffile) {
+			bsdtar_warnc(bsdtar, errno,
```
cperciva: Won't this result in "blahblahblah continuing anyway" immediately followed by a complaint about missing keys if we try to perform a non-dryrun when our config file points at a nonexistent key file?

lib/keyfile/keyfile.c
gperciva: what about
```
[EACCES]
Search permission is denied for a component of the path prefix.
```

I mean, should that also count as file-not-found?

gperciva:
Updated with cleaned-up code.  Behaviour is now:

```
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
$ ./tarsnap -c -f a0 --dry-run ~/backup
tarsnap: stat(/home/td/new-machine.keys-bad): No such file or directory
tarsnap: Cannot read key file: /home/td/new-machine.keys-bad
         but continuing anyway with dry-run.
tarsnap: Ignoring cachedir due to missing or invalid keyfile in config file.
tarsnap: Performing dry-run archival without keys
         (sizes may be slightly inaccurate)
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
$ ./tarsnap -c -f a0 ~/backup
tarsnap: stat(/home/td/new-machine.keys-bad): No such file or directory
tarsnap: Cannot read key file: /home/td/new-machine.keys-bad
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
```

gperciva:
Updated to continue a dryrun despite any keyfile-reading error.

```
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
$ ./tarsnap -c -f a1 ~/backup
tarsnap: fopen(/home/td/keys-no-filesystem-read.txt): Permission denied
tarsnap: Cannot read key file: /home/td/keys-no-filesystem-read.txt
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
$ ./tarsnap -c -f a1 --dry-run ~/backup
tarsnap: fopen(/home/td/keys-no-filesystem-read.txt): Permission denied
tarsnap: Ignoring cachedir due to missing or invalid keyfile in config file.
tarsnap: Performing dry-run archival without keys
         (sizes may be slightly inaccurate)
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
$ ./tarsnap -c -f a1 --dry-run --keyfile ~/keys-no-filesystem-read.txt ~/backup
tarsnap: fopen(/home/td/keys-no-filesystem-read.txt): Permission denied
tarsnap: Cannot read key file: /home/td/keys-no-filesystem-read.txt
td@gin: ~/src/tarsnap/build (dryrun-keyless-ignore-config)
```

Issue #87, opened 2015-10-31
tarsnap --print-key-id


cperciva:
Read a key file and print the machine ID #.  Will be undocumented; this is a private interface for the GUI to use.

gperciva:
Might I register a slight protest against "will be undocumented"?  Is there any foreseeable harm in documenting it with something like "This is unlikely to be useful when tarsnap is being invoked directly from the command line." (as we did for `--creationtime`) ?

gperciva:
Another idea: in addition to making it public, how about adding a separate `--print-key-permissions` (or `authorization` which would print whether that key had READ, WRITE, DELETE, NUKE permissions.  This arose on the mailing list in 2012:  http://mail.tarsnap.com/tarsnap-users/msg00542.html

gperciva:
This was added (with documentation) in #87.

PR #92, opened 2015-11-09
Add --print-key-id

gperciva:
This is not documented; it is a private interface for the GUI to use.

Addresses #87.


gperciva:
Addition to my questioning of the "will be undocumented" part (in #87): I assume that if it's undocumented, it shouldn't be added to bash completion.  If this is *really* what we want, then I should add a comment(s) to the patch so that anybody reading the code won't think that they've found a bug bounty in terms of this missing documentation and bash completion.

gperciva:
This will be wrapped into printing additional info about keyfiles in tarsnap-keymgmt instead.

PR #94, opened 2015-11-10
Chunkification for disk errors

gperciva:
There's a few things to note here, so I'll add comments.


gperciva:
Reminder: the first big question is whether I can use
```
srand(time(NULL));
...
if (rand() < bsdtar->probability_check_file * RAND_MAX)
```
to check if any particular file should be checked (rehashed).  I suspect that you're going to say "no, even though this use of randomness doesn't matter at all, I cannot allow any `rand()` to appear in the Tarsnap code base.  All my crypto friends would mock me mercilessly."

If so, then should I use `entropy_read()` with a 32-byte buffer?  or `crypto_entropy_read()`?  or...?

gperciva:
Let's revisit this later.

PR #97, opened 2015-11-12
Bsdtar better cleanup

gperciva:
Since we've got the `atexit` function now, we might as well use it properly.  This clears up some memory leaks when calling `bsdtar_errc` within parts of the code.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -196,6 +196,9 @@ bsdtar_atexit(void)
 #if HAVE_REGEX_H
 	cleanup_substitution(bsdtar);
 #endif
+
+	/* Clean up network layer. */
+	network_fini();
 }
 
 int
@@ -1023,9 +1026,6 @@ main(int argc, char **argv)
 	    "va = %12g ms^2  max = %12g ms\n",
 	    N, mu * 1000, va * 1000000, max * 1000);
 #endif
-
-	/* Clean up network layer. */
-	network_fini();
 
 #ifdef PROFILE
 	/*
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -190,6 +190,12 @@ bsdtar_atexit(void)
 	/* Free strings allocated by strdup. */
 	free(bsdtar->cachedir);
 	free(bsdtar->homedir);
+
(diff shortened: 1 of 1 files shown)


tar/bsdtar.c
```diff
@@ -1008,11 +1017,6 @@ main(int argc, char **argv)
 		break;
 	}
 
-	cleanup_exclusions(bsdtar);
```
cperciva: These are unchanged from libarchive, aren't they?  In general I think "keep code consistent with libarchive" trumps "reorganize code to make it cleaner".

gperciva:
Well, it's a trade-off of cleaning up memory vs. consistency with libarchive.  libarchive is peppered with `exit(1)` instead of returning an error message to the calling function (although some functions *do* return an error message).  As a result, any cleanup that's done at the end of `bsdtar.c` `main()` may or may not be run, depending on the exact error.

I was thinking that we could use `bsdtar_atexit()` to clean up bits and pieces which were abandoned if tarsnap encountered an `exit(1)`, instead of assuming that the execution path would reaturn to `main()`.  We probably won't be able to clean everything for all error cases, but at the very least we could deal with any `bsdtar->foo` allocations, network code, etc.

For example, if tarsnap fails at
```
        if (sigquit_init())
                exit(1);
```
in `tar/write.c`, we currently dump almost 1.5 Mb of un-de-allocated memory, all of which is still reachable (we still have pointers to it).

(in an ideal world, I would spend a few days error-code-returning libarchive, those patches would be accepted or discussed within days, there would be a libarchive release a week later, and then tarsnap would update to using that fixed version of libarchive.  But we don't live in that world, so I think that doing more cleanup in `bsdtar_atexit` is an appropriate compromise)

tar/bsdtar.c
gperciva: oh, in case there's any doubt: yes, moving that `cleanup_exclusions(bsdtar);` would be a change from libarchive code.  (see my earlier comment for why I still think it's appropriate in this case)

cperciva:
Fair enough.  Maybe worth asking Tim if he's interested in having bsdtar follow tarsnap's example here, though?

PR #98, opened 2015-11-13
tarsnap.conf.sample: add humanize-numbers (disabled)

gperciva:
This doesn't change the default behaviour, but I think that adding this as a
"hint" is useful.

### DIFF
--- a/tar/tarsnap.conf.sample
+++ b/tar/tarsnap.conf.sample
@@ -15,6 +15,11 @@ print-stats
 # Create a checkpoint once per GB of uploaded data.
 checkpoint-bytes 1G
 
+### Commonly useful options
+
+# Use SI prefixes to make numbers printed by --print-stats more readable
+#humanize-numbers
+
 ### Other options, not applicable to most systems
 
 # Aggressive network behaviour: Use multiple TCP connections when


cperciva:
The placement amidst a set of options which are turned on is potentially confusing.  Maybe add a new `### Commonly useful options` section?

gperciva:
updated

tar/tarsnap.conf.sample
```diff
@@ -15,6 +15,11 @@ print-stats
 # Create a checkpoint once per GB of uploaded data.
 checkpoint-bytes 1G
 
+### Commonly useful options
+
+## Use SI prefixes to make numbers printed by --print-stats more readable
```
cperciva: Why `##` ?

gperciva:
in python code, when I comment out some stuff I'm working on, I also add another `#` character in front of any comments related to the commented-out code.  (mainly because python doesn't have `/* ... */` and it's simpler to let my editor add another `#` character to every line)

I've updated the PR without that extra `#`.

cperciva:
I wondered if that was it.  I wouldn't mind (it makes perfect sense to add an extra level of commented-out-ness) except that I didn't do it that way in the rest of the file.

PR #102, opened 2015-12-01
keymgmt: print key ID and permissions

gperciva:


gperciva:
updated, along with bash completion and man page.

Not 100% certain about the desired man syntax with respect to {}, and casual googling is failing me.  This is what is does after this PR:
```
SYNOPSIS
     tarsnap-keymgmt --outkeyfile new-key-file [-r] [-w] [-d] [--nuke]
                     [--passphrased] [--passphrase-mem maxmem]
                     [--passphrase-time maxtime] key-file ...
     tarsnap-keymgmt {--print-key-id key-file}
     tarsnap-keymgmt {--print-key-permissions key-file}
```

cperciva:
From mdoc(7):

`If there are alternative values for a mandatory parameter, braces are used (together with `|') to enclose the value set.`

Take out the braces and you're fine.  Each separate line of the synopsis is an alternative way of running the command.

keymgmt/keymgmt.c
```diff
@@ -23,11 +23,74 @@ usage(void)
...
+	if (has_read && has_write && has_delete)
+		fprintf(stdout, "reading, writing, and deleting.\n");
+	if (has_read && has_write && !has_delete)
+		fprintf(stdout, "reading and writing.\n");
+	if (has_read && !has_write && has_delete)
+		fprintf(stdout, "reading and deleting.\n");
+	if (has_read && !has_write && !has_delete)
+		fprintf(stdout, "reading.\n");
+	if (!has_read && has_write && has_delete)
+		fprintf(stdout, "writing and deleting.\n");
```
cperciva: Make this "writing and nuking", since you can't delete archives without being able to read them.

keymgmt/keymgmt.c
```diff
@@ -23,11 +23,74 @@ usage(void)
...
+	if (!has_read && has_write && !has_delete)
+		fprintf(stdout, "writing.\n");
+	if (!has_read && !has_write && has_delete)
+		fprintf(stdout, "deleting.\n");
```
cperciva: nuking

keymgmt/keymgmt.c
```diff
@@ -23,11 +23,74 @@ usage(void)
...
+	uint64_t machinenum = (uint64_t)(-1);
+
+	/* Read keyfile and machine name. */
+	if (keyfile_read(keyfilename, &machinenum, ~0)) {
+		warnp("Cannot read key file: %s", keyfilename);
+		exit(1);
+	}
+
+	/* Print key ID. */
+	fprintf(stdout, "%li\n", machinenum);
```
cperciva: What's `%li`?

gperciva:
Updated.  `%li` is something that works on my machine with my keys, but obviously the terrifying-looking `"%" PRIu64 "\n"` is the correct solution.

cperciva:
For future reference, PRIu64 is a macro which is `llu` on some systems (where `long long` is 64-bits), `lu` on others (where `long` is 64-bits) and `u` on yet others (where `int` is 64-bits).  Since we know the value we're printing is 64 bits but we don't know what type that maps to, we need the macro.

PR #103, opened 2015-12-06
clean up memory in error paths for tarsnap-keygen

gperciva:


keygen/keygen_network.c
```diff
@@ -77,6 +77,9 @@ keygen_network_register(struct register_internal * C)
 	/* Success! */
 	return (0);
 
+err2:
+	netpacket_close(NPC);
+	network_fini();
```
cperciva: netpacket_close definitely belongs here, since NPC was allocated here; but network_fini should be called from the same function regardless of whether we're succeeding or failing.

That said, maybe the network_fini from keygen_actual.c belongs here?

gperciva: Style question here: if `network_fini()` is going to be called in the error path inside `keygen_network_register()` [which makes sense], then what happens to the error labels?

Currently, it seems that `err0` is only used for `return (-1);` and possibly `bsdtar->return_value = 1`.  There's nothing in `STYLE` that enforces this, but I vaguely recall you telling me that.

Should I use
```
err1:
        network_fini();
err0:
        return (-1);
```

even if nothing ever calls `err0` ?  Or am I mistaken in `err0` only being for a `return` statement, and it's ok to call `network_fini()` inside `err0:` ?

cperciva:
Correct, after `err0` we should only have "return a failure status", however that's implemented.  If we don't need an `err0` jump target then simply don't have one; leave a blank line, though.  See kivaloo/kvlds/btree_cleaning.c for example.

gperciva:
updated without a useless `err0:`

PR #105, opened 2015-12-13
Added an explanation of using the '-' character with the '-T' option …

Vintila:
Just a small update to the docs.


cperciva:
Thank you!  I adjusted this slightly but the content has been added.

Vintila:
You're welcome. I'll use this opportunity to say I think you have a great product and you run your business in a very fair manner.

PR #107, opened 2016-01-03
Replace malloc(x*y) with imalloc(x,y) which gives us a free integer o…

jmutkawoa:
…verflow check for little performance cost.


cperciva:
The `imalloc` function shouldn't be used directly; it exists only to be used via the `IMALLOC` macro.  So for example

    if ((*blist = malloc(*blistlen * sizeof(struct block))) == NULL) {

should turn into

    if (IMALLOC(*blist, *blistlen, struct block)) {

PR #108, opened 2016-01-04
dry-run - manpage hint

dirkx:
In a total greenfield situation - a valid tarsnap.conf may contain key and cachedir entries that stop the --dry-run flag from working.

### DIFF
--- a/tar/tarsnap.1-man.in
+++ b/tar/tarsnap.1-man.in
@@ -274,7 +274,13 @@ within a few kB or a fraction of a percent) to if
 \fB\%tarsnap\fP
 is run without the
 \fB\--dry-run\fP
-option.
+option.  Requires unsetting
+the 
+.Nm cachedir 
+and 
+.Nm keyfile
+entries in the configuration file if tarsnap-keygen was
+never run (i.e. there is no valid tarsnap key).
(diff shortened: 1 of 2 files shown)


gperciva:
Thanks for the commit!  I think that #91 addresses this by making the binary more forgiving (it will continue despite a non-readable keyfile).  A few weeks ago, we updated the FAQ to suggest

    tarsnap --dry-run --no-default-config --print-stats --humanize-numbers -c ~/mydir

to avoid this problem for people running 1.0.36.

dirkx:
> On 06 Jan 2016, at 21:59, Graham Percival <notifications@github.com> wrote:
> 
> Thanks for the commit! I think that #91 addresses this by making the binary more forgiving (it will continue despite a non-readable keyfile). A few weeks ago, we updated the FAQ to suggest
> 
> tarsnap --dry-run --no-default-config --print-stats --humanize-numbers -c ~/mydir
> 
> to avoid this problem for people running 1.0.36.

Lovely - Makes perfect sense.

Dw.

cperciva:
I've merged #91 so I'm going to close this now since it doesn't seem necessary any more.  Thanks anyway!

PR #109, opened 2016-01-04
Replace malloc(x*y) with IMALLOC() which gives us a free integer over…

jmutkawoa:
…flow check for little performance cost and also checks for malloc(0

### DIFF
--- a/recrypt/recrypt.c
+++ b/recrypt/recrypt.c
@@ -15,6 +15,7 @@
 #include "tarsnap_opt.h"
 #include "tsnetwork.h"
 #include "warnp.h"
+#include "imalloc.h"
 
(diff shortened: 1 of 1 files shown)


recrypt/recrypt.c
```diff
@@ -15,6 +15,7 @@
 #include "tarsnap_opt.h"
 #include "tsnetwork.h"
 #include "warnp.h"
+#include "imalloc.h"
```
cperciva: Includes should be sorted in alphabetical order within each block -- this should be between `#include "crypto.h"` and `#include "keyfile.h"`.

jmutkawoa:
I corrected this mistake via the ticket #118

PR #111, opened 2016-01-07
Debian packages

gperciva:
First step towards binary packages.  This makes it easier for people (including us) to generate their own `deb` files.

In offline communication, Colin mentioned adding `mkdebsource.sh` to a `make publish` target, but I don't see an existing target -- maybe he was thinking of scrypt / kivaloo / etc?  I could add a `publish:` target if desired.


gperciva:
A few more things to work on; will reopen later.

gperciva:
Due to complexities of signing this on non-deb systems, we're not adding a `make publish` target.

Instructions here:
https://github.com/Tarsnap/tarsnap/tree/debian-packages/pkg
https://github.com/Tarsnap/tarsnap/blob/debian-packages/pkg/DEBIAN.md

URL is not active yet.

gperciva:
need to play games in order to sort `1.0.37a` before `1.0.37` so that `dpkg` doesn't warn about "downgrading" an alpha release to a full release.

gperciva:
This now creates version numbers so that Debian sorts them correctly.

PR #115, opened 2016-01-08
man: emphasize that tarsnap-keyregen is required

gperciva:
This follows up from a mailing list question:
http://mail.tarsnap.com/tarsnap-users/msg00234.html

### DIFF
--- a/keyregen/tarsnap-keyregen.1-man
+++ b/keyregen/tarsnap-keyregen.1-man
@@ -26,7 +26,10 @@ and
 \fBtarsnap\fP(1).
 The term "compatible"
 here means that it is possible to re-encrypt archives stored with
-the first set of keys to be stored with the second set of keys.
+the first set of keys to be stored with the second set of keys.  This is
+required because Tarsnap has some keys which need to stay the same when
+re-encrypting data; otherwise, existing archives will become unreadable and
+cannot be used for deduplication.
 .PP
 The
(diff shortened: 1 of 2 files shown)


keyregen/tarsnap-keyregen.1-man
```diff
@@ -26,7 +26,10 @@ and
 \fBtarsnap\fP(1).
 The term "compatible"
 here means that it is possible to re-encrypt archives stored with
-the first set of keys to be stored with the second set of keys.
+the first set of keys to be stored with the second set of keys.  This is
+required because Tarsnap has some keys which need to stay the same when
+re-encrypting data; otherwise, existing archives will become unreadable and
```
cperciva: Hmm, I don't think re-encrypting to completely different keys would make archives *unreadable*.  It would just screw up the deduplication.

gperciva:
Quote from Colin Percival, Tue, 08 Feb 2011 07:39:24 -0800,
http://mail.tarsnap.com/tarsnap-users/msg00234.html

```
for example, there is a key used for mapping archive names to the 256-bit
names which identify metadata blocks.  If this key is changed, Tarsnap won't
be able to read archives since it won't be able to find the right metadata
blocks.
```

cperciva:
Oh right, that's a different key but one which is also preserved by recrypt.

PR #118, opened 2016-01-13
Replace malloc(x*y) with IMALLOC() which gives us a free integer over…

jmutkawoa:
…flow check for little performace cost and also checks for malloc(0)


cperciva:
Thanks!

PR #119, opened 2016-01-18
Only print "inaccurate sizes" message if --print-stats

gperciva:


tar/bsdtar.c
```diff
@@ -941,9 +941,11 @@ main(int argc, char **argv)
 			if (crypto_keys_generate(CRYPTO_KEYMASK_USER))
 				bsdtar_errc(bsdtar, 1, 0,
 				    "Error generating keys");
-			bsdtar_warnc(bsdtar, 0,
-			    "Performing dry-run archival without keys\n"
-			    "         (sizes may be slightly inaccurate)");
+			if (bsdtar->option_print_stats_set)
```
cperciva: I'm pretty sure this should be `option_print_stats`?  Because `--no-print-stats` will still turn on `option_print_stats_set`.

gperciva:
updated

PR #121, opened 2016-01-19
Magic getopt

gperciva:


cperciva:
When importing code from libcperciva into a project, I prefer to just have a single "Add foo from libcperciva" commit rather than cherry-picking the entire commit history for those files.  It's not too bad here but in some cases it would be horrible.

PR #126, opened 2016-01-21
Fix header alphabetical

gperciva:
This is a bit nitpicky, but I figured I might as well do it.


gperciva:
updated PR to remove the python file changes.

PR #129, opened 2016-01-21
Use magic getopt in tarsnap-keymgmt

gperciva:
Not as confident about this one.

### DIFF
--- a/keymgmt/keymgmt.c
+++ b/keymgmt/keymgmt.c
@@ -8,6 +8,7 @@
 #include <string.h>
 
 #include "crypto.h"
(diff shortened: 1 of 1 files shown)


cperciva:
This looks good.  The `char *` vs `const char *` bit in `strtok_r` is arguably a bug in the old code -- modifying strings in `argv` is a nasty thing to do, even if it is usually safe.

As for turning --print-key-id and --print-key-permissions into options which take arguments: Yes please.  You can even change `int print_key_id` to `char * print_key_id_file` to keep the code simple.

PR #133, opened 2016-02-03
Free upon errors in config file(s)

gperciva:
Makes valgrind happy if there's any problems in parsing the config file.  I'm not 100% happy with this PR since it's a bit awkward due to interfacing with libarchive code (longer comments in https://github.com/Tarsnap/tarsnap/commit/a52cd2e1220406c01a621d1116c5f3515880bda7), but it's worth bringing to your attention.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -168,6 +168,7 @@ bsdtar_init(void)
 	bsdtar->siginfo = NULL;
 	bsdtar->substitution = NULL;
 	bsdtar->keyfile = NULL;
+	bsdtar->conffile = NULL;
 
 	/* We don't have bsdtar->progname yet, so we can't use bsdtar_errc. */
 	if (atexit(bsdtar_atexit)) {
@@ -194,6 +195,7 @@ bsdtar_atexit(void)
 	free(bsdtar->homedir);
 	free(bsdtar->option_csv_filename);
 	free(bsdtar->keyfile);
+	free(bsdtar->conffile);
 
 	/* Free matching and (if applicable) substitution patterns. */
 	cleanup_exclusions(bsdtar);
@@ -214,7 +216,6 @@ main(int argc, char **argv)
 	char			 buff[16];
 	char			 cachedir[PATH_MAX + 1];
 	struct passwd		*pws;
-	char			*conffile;
 	const char		*missingkey;
 	time_t			 now;
 	size_t 			 i;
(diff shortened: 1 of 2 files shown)


gperciva:
I'm going to work on a simpler/cleaner way of doing this.

gperciva:
This now consists of the first 2 commits of the original PR.  It leaves a FILE open and a 8192-char buffer, both from `tar/util.c` `process_lines()`, but I'm still looking for cleaner ways of doing that part.

tar/bsdtar.c
```diff
@@ -1315,62 +1319,65 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
-	if (conf_opt[optlen]) {
+	if (bsdtar->conf_opt[optlen]) {
 		/* NUL-terminate the option name. */
-		conf_opt[optlen] = '\0';
+		bsdtar->conf_opt[optlen] = '\0';
 
 		/* Find the start of the argument. */
-		conf_arg = conf_opt + optlen + 1;
-		conf_arg += strspn(conf_arg, " \t");
+		bsdtar->conf_arg = bsdtar->conf_opt + optlen + 1;
```
cperciva: You're setting `bsdtar->conf_arg` to a string which was not allocated by malloc.  What happens if we exit from this point?

gperciva:
Updated PR.

tar/bsdtar.c
```diff
@@ -1353,24 +1358,23 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
-		if (asprintf(&conf_arg_malloced, "%s%s",
+		if (asprintf(&bsdtar->conf_arg, "%s%s",
 		    bsdtar->homedir, &conf_arg[1]) == -1)
 			bsdtar_errc(bsdtar, 1, errno, "Out of memory");
-
-		/* Use the expanded argument string hereafter. */
-		conf_arg = conf_arg_malloced;
 	} else {
-		conf_arg_malloced = NULL;
+		bsdtar->conf_arg = NULL;
```
cperciva: If we come here...

tar/bsdtar.c
```diff
@@ -1353,24 +1358,23 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
 	}
 
 	/* Process the configuration option. */
-	dooption(bsdtar, conf_opt, conf_arg, 1);
+	dooption(bsdtar, bsdtar->conf_opt, bsdtar->conf_arg, 1);
```
cperciva: ... then what does this do?

Maybe line 1365 should be a strdup instead?

gperciva: crap.  No, it can't be strdup, since `conf_arg` can be `NULL`, and `strdup(NULL)` is undefined.  I'll go back to something closer to the original code.

gperciva:
Updated, and should be easier to read / track history now.

cperciva:
Looks good to me.  Should I merge now or did you want to rebase anything?

gperciva:
Go ahead and merge it now.  More granularity in the history isn't a bad thing.

Issue #132, opened 2016-01-30
Add --verify-config option


cperciva:
Suggestion from a Tarsnap user:

> I’m a rather new but very happy user of tarsnap, and I have started pushing backups all over the place.  To do this I’m using a git tree that holds configuration both of tarsnap proper, and that of feather that I use to schedule backups.  In this context I would like to be able to run tarsnap with the single purpose of verifying the validity of the configuration file, so I can have tools prevent broken (well, syntactically broken) configuration files from being pushed to clients.  Something as simple as: tarsnap --verify-config

We won't want this to load tarsnap keys, since we might want to check a configuration file on a system without keys installed.  So the handling of the `--keyfile` option will need to change from "load the keys" to "store the path and load the keys later".  But that should be fairly straightforward, and I don't *think* there are any other similar issues.

Then we should be able to check this option and bail out after the configuration file(s) are processed.

gperciva:
This was added in #134, and will be in 1.0.37.

PR #134, opened 2016-02-03
Verify config

gperciva:
Addresses #132.

It's not clear to me from the quoted text that the user really wants to ignore missing keyfiles (unless I'm misunderstanding something about "feather").  And if we really don't want to complain about missing keyfiles, then I'm not certain why @cperciva wanted the `--keyfile` code to be "store the path and load the keys later" since we wouldn't be loading them (?).  But this is a decent starting place for a discussion.


cperciva:
> It's not clear to me from the quoted text that the user really wants to ignore missing keyfiles

The use case here is "syntax-check a tarsnap configuration file before deploying it to the systems which will use it".  The configuration file won't necessarily be edited on a system with all the same keys in the same places.

gperciva:
From offline discussion, @cperciva wanted this to warn if there's two keyfiles given.  But `dooption()` contains:

```
optset:
        /* Option specified multiple times. */
        if (fromconffile == 0) {
                usage(bsdtar);
        }
        return;
```

so it looks like tarsnap currently deliberately ignores repeated options from config files.

cperciva:
Ok, I remember what it was that I saw earlier.  Yes, we ignore multiple options from configuration files, since that's how overriding keys works.

The issue actually has nothing to do with `--verify-config` except that it touches some of the same code; it's actually a problem with the keyless dry-run code.  If we're doing a dry run and `~/.tarsnaprc` specifies a key file which we can't key, we don't set `have_keys` and then we try to read and use the keyfile specified in `/usr/local/etc/tarsnap.conf`.  Which would (a) result in the warning being skipped (since we would have keys, albeit the wrong ones) and (b) result in bogus statistics (since using the wrong keys would break deduplication).

Can you fix that bug and then come back to this afterwards?

gperciva:
Updated; this PR depends (and includes) #138.

gperciva:
Updated with much simpler "bail early" approach.

PR #135, opened 2016-02-07
man: add --version to synopsis and description

gperciva:
I know that the C implementation doesn't use the `bsdtar->mode` variable to
handle `--version`, but as far as the user is concerned, it is incorrect to
write
    The first option to tarsnap is a mode indicator from the following list:
if that list does not include `--version`.


cperciva:
Hmm, I think I would have written "Print version number" rather than "Print version".  If you think what you have is correct I'll merge it though; you're the docs guy...

gperciva:
Updated, adding "version number" where you left off "number".  :)

cperciva:
Looks like it was actually Tim who left off "number". ;-)

PR #138, opened 2016-02-10
Only attempt to load_keys() once

gperciva:


tar/bsdtar.c
```diff
@@ -876,6 +878,20 @@ main(int argc, char **argv)
 	if (bsdtar->strip_components != 0)
 		only_mode(bsdtar, "--strip-components", "xt");
 
+	/* Attempt to load keyfile. */
+	if (bsdtar->keyfile != NULL) {
+		if (load_keys(bsdtar, bsdtar->keyfile) == 0)
+			bsdtar->have_keys = 1;
+		else {
+			if (bsdtar->option_dryrun)
```
cperciva: How about setting a `bsdtar->keyfile_from_config` option so that a command-line `--keyfile` directive doesn't get the "ignore if unavailable" dry-run handling?

gperciva:
Updated with `bsdtar->keyfile_from_config`.  PR needs rewriting history before merging.

Question: if there are two keyfiles in the config file, do you want the first to be used, or the second?  The current behaviour is to use the first `keyfile` (despite the branch name).

cperciva:
"The current behaviour is to use the first keyfile"

Yes, that's correct.  This has been the historical behaviour of tarsnap due to the way that it handles configuration files (ignoring options which were set earlier) and we shouldn't change that now.

tar/bsdtar.c
```diff
@@ -876,6 +878,21 @@ main(int argc, char **argv)
 	if (bsdtar->strip_components != 0)
 		only_mode(bsdtar, "--strip-components", "xt");
 
+	/* Attempt to load keyfile. */
+	if (bsdtar->keyfile != NULL) {
+		if (load_keys(bsdtar, bsdtar->keyfile) == 0)
+			bsdtar->have_keys = 1;
+		else {
+			if ((bsdtar->option_dryrun) &&
+			    (bsdtar->keyfile_from_config))
```
cperciva: The extra parentheses are unnecessary.

cperciva:
Looks good aside from the surplus parentheses.

gperciva:
Updated PR to remove extra parentheses, and rewrote history.

PR #142, opened 2016-03-01
pkg: Add Fedora specfile

bgilbert:


pkg/fedora/tarsnap.spec
```diff
@@ -0,0 +1,59 @@
...
+Release:        2%{?dist}
+Summary:        Secure, efficient online backup service
+
+License:        Proprietary
+URL:            http://www.tarsnap.com/
+Source0:        http://www.tarsnap.com/download/tarsnap-autoconf-%{version}.tgz
+
+BuildRequires:  gcc
+BuildRequires:  pkgconfig(bash-completion)
+BuildRequires:  pkgconfig(bzip2)
```
gperciva: Are you sure we need `bzip2`?  I think we need `zlib-devel`, but not `bzip2`.

Also, on CentOS 7-1.1503.el7 (I hope this is "similar enough" to Fedora, but please let me know if it isn't), trying to install the build dependencies:

```
sudo yum install yum-utils
sudo yum-builddep pkg/fedora/tarsnap.spec
```

gives me:

```
Getting requirements for pkg/fedora/tarsnap.spec
 --> Already installed : gcc-4.8.3-9.el7.x86_64
 --> Already installed : 1:bash-completion-2.1-6.el7.noarch
 --> Already installed : e2fsprogs-devel-1.42.9-7.el7.x86_64
 --> Already installed : 1:openssl-devel-1.0.1e-42.el7.9.x86_64
 --> xz-devel-5.1.2-12alpha.el7.x86_64
 --> Already installed : zlib-devel-1.2.7-13.el7.x86_64
 --> libacl-devel-2.2.51-12.el7.x86_64
 --> libattr-devel-2.4.46-12.el7.x86_64
Error: No Package found for pkgconfig(bzip2)
```

Could you comment on writing
```
BuildRequires:  pkgconfig(bzip2)
```
instead of
```
BuildRequires:  bzip2
```
?  if I remove the `pkgconfig()` part, it installs `bzip2` without problem.

pkg/fedora/tarsnap.spec
```diff
@@ -0,0 +1,59 @@
...
+BuildRequires:  pkgconfig(ext2fs)
+BuildRequires:  pkgconfig(libcrypto)
```
gperciva: As a general note, how did you decide on this list of requirements?  The Tarsnap website claims that

```
sudo yum install gcc make glibc-devel openssl-devel zlib-devel e2fsprogs-devel
```

works, and I checked that on a bare CentOS installation.  Does Fedora have very different package names, or should we be fixing something in the website list for CentOS users, or...?

pkg/fedora/tarsnap.spec
```diff
@@ -0,0 +1,59 @@
...
+mkdir -p $RPM_BUILD_ROOT/%{_pkgdocdir}
+mv $RPM_BUILD_ROOT/%{_sysconfdir}/tarsnap.conf.sample \
+        $RPM_BUILD_ROOT/%{_pkgdocdir}
+
+
+%files
+%{_bindir}/*
+%{_datarootdir}/bash-completion/completions/*
+%{_mandir}/man*/*
+%{_pkgdocdir}/*
```
gperciva: On CentOS 7 (again, my apologies if this is not "similar enough" to Fedora), I get:
```
[gperciva@localhost tarsnap]$ rpmbuild --version
RPM version 4.11.1
[gperciva@localhost tarsnap]$ rpmbuild -bb pkg/fedora/tarsnap.spec
...

RPM build errors:
    File must begin with "/": %{_pkgdocdir}/*
``

pkg/fedora/tarsnap.spec
bgilbert: See below as to the bzip2 dependency.  `BuildRequires: pkgconfig(foo)` [is preferred](https://fedoraproject.org/wiki/Packaging:PkgConfigBuildRequires) to `BuildRequires: foo-devel` for dependencies that are detected via pkg-config, since the former is more robust to package renames.  The CentOS 7 version of bzip2-devel doesn't have a pkg-config file, so this fails.

In any event, I misremembered the packaging guideline; it does *not* apply to packages detected by hand, even if they ship with a pkg-config file.  So, that notation should only be used for `bash-completion`.  I'll update the specfile.

pkg/fedora/tarsnap.spec
bgilbert: The `configure` script checks for a few optional dependencies not listed on the Tarsnap website.  I looked again, and it seems libattr is not actually needed for xattr support; tarsnap will use the syscall wrappers in glibc.  I'll update the specfile.  libacl does seem to be required for ACL support, though, and I'm guessing bzip2 and lzma are used for `@archive-file` support.  Either way, it'd be good to get the configure script and website in sync.

`pkg/debian/control` includes all of those dependencies except liblzma.  So, if the above is correct, liblzma should be added there and libattr should be dropped.

pkg/fedora/tarsnap.spec
bgilbert: `%{_pkgdocdir}` [was added a few Fedora releases ago](https://fedoraproject.org/wiki/Changes/UnversionedDocdirs#Detailed_Description), and apparently isn't in CentOS 7.  There's a compatibility define that can be added if desired.

bgilbert:
Updated to drop `libattr-devel` and fix erroneous `pkgconfig()` dependencies.

gperciva:
@bgilbert wrote:
> The configure script checks for a few optional dependencies not listed on the Tarsnap website. I looked again, and it seems libattr is not actually needed for xattr support; tarsnap will use the syscall wrappers in glibc.

Hmm.  Is there a standard convention for packaging with regards to optional dependencies?  My first instinct is that in this age of basically infinite storage space (at least for non-video, non-"big data" machine learning), there's no harm in using `BuildRequires` for anything that's checked by `configure`.  I mean, `libattr1-dev` takes 115 kb of disk space (on Ubuntu 14.04).  I'm particularly hesitant to start second-guessing a configure script; if libarchive want to know if there's `libattr` on the system (and yes, the latest libarchive still checks for this), then I think there's no harm in including that in `BuildRequires`.

I have a similar argument in mind for `bash-completion` (which I know that you've kept; I'm just mentioning this as another example).  Yes, Tarsnap will work just fine without `bash-completion`, but I think it's a nice extra feature, so we might as well include it in the package requirements.  If somebody wants to set up Tarsnap on an embedded system and can't afford the extra 954 kb that `bash-completion` adds, then surely they'd be compiling from source anyway, with their own compiler options like `-Os` and not `-g`.

bgilbert:
The typical practice is to enable everything, but there are limits.  The size of the build dependencies doesn't matter, but particularly unreasonable *runtime* dependencies may need special handling.  For example, bash-completion is only a build dependency (for the pkg-config file), but if a binary package had to depend on every shell for which it provided completions, those completions would probably need to be moved to subpackages.

libattr is an odd case for several reasons.  First, libarchive is actually hunting for any of several ways to get xattr support, and the functions it needs are provided by the C library.  Second, the functions are provided by *both* libattr and the C library; I'm not sure whether ELF guarantees that the linker will consistently pick one library or the other if both are linked in.  (In this specific case it doesn't matter, because the functions aren't operating on shared state -- it's not like calling one library's `malloc()` implementation and another library's `free()`.)  And third, it turns out that libacl depends on libattr (and likewise for the -devel packages), so you'll get libattr even if you don't ask for it.  Except for the third point, I'd suggest leaving it out to reduce bloat and confusion, but since libattr will be linked in anyway, it's reasonable to be explicit if that's your preference.

Re `-g`, note that debug symbols are automatically split off into a `-debuginfo` package anyway.

cperciva:
@gperciva should this be closed now that #150 has been merged?

gperciva:
I believe that this can be closed.  There's still plenty of good info in here that will be helpful for me in the future, but I have that saved elsewhere.

PR #147, opened 2016-03-12
Move C preprocessor options to CPPFLAGS

bgilbert:
By default, the MacPorts `port` command runs configure scripts with `CPPFLAGS=-I/opt/local/include`, to pick up headers from MacPorts-installed libraries in /opt/local.  This was causing Tarsnap to build against /opt/local/include/archive.h instead of libarchive/archive.h, causing warnings such as

    tar/glue/archive_multitape.c:208:18: warning: implicit declaration of function 'archive_read_get_entryleft' is invalid in C99 [-Wimplicit-function-declaration]

The cause is that `$(CPPFLAGS)` precedes `$(<module>_CFLAGS)` in the argument list passed to `$(CC)`.  Fix by moving preprocessor options to `$(<module>_CPPFLAGS)` so they will precede `$(CPPFLAGS)`.

Also make some tiny fixes to C preprocessor flags, and clean up the output of `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,30 @@ config.log
 configure
 confinc
 confmf
+
+# ignore build-generated files
+*.a
+*.o
+.deps
+.dirstamp
+/Makefile
+/a.out.dSYM
+/config.h
+/config.status
+/cpusupport-config.h
+/stamp-h1
+
+# ignore installed files
+/keygen/tarsnap-keygen.1
+/keymgmt/tarsnap-keymgmt.1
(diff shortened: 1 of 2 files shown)


gperciva:
Should I apply the `CPPFLAGS` thing to the other Tarsnap projects (scrypt, etc)?

cperciva:
Yes, please do.  To be honest, I never saw the point of separating CFLAGS and CPPFLAGS until now... I assumed they were a vestige of when the preprocessor was actually a separate program.

PR #149, opened 2016-03-22
Initialize cachedir

gperciva:
This is useful for the GUI in https://github.com/Tarsnap/tarsnap-gui/issues/45

### DIFF
--- a/tar/chunks/chunks_directory.c
+++ b/tar/chunks/chunks_directory.c
@@ -362,6 +362,47 @@ err0:
 }
 
 /**
+ * chunks_directory_exists(cachepath):
+ * Return 1 if the /directory file exists within ${cachepath}, 0 if it does
+ * not, or -1 if there is an error.
+ */
+int
+chunks_directory_exists(const char * cachepath)
+{
+	char * directory_filename;
+	struct stat sb;
+	int rc;
+
+	/* Prepare filename. */
+	if (asprintf(&directory_filename, "%s/directory", cachepath) == -1) {
+		rc = -1;
+		goto done;
+	}
+
+	/* Check if file exists. */
+	if (stat(directory_filename, &sb) == 0) {
+		/* File exists. */
+		rc = 1;
+	} else {
+		if (errno == ENOENT) {
+			/* File does not exist. */
+			rc = 0;
+		} else {
+			/* Other error. */
+			warnp("stat(%s)", directory_filename);
+			rc = -1;
+		}
+	}
+
+	/* Clean up memory. */
+	free(directory_filename);
+
+done:
+	/* Return result code. */
+	return (rc);
+}
+
+/**
  * chunks_directory_free(htab, dir):
  * Free the hash table ${htab} of struct chunkdata records, all of its
  * elements, and ${dir}.
--- a/tar/chunks/chunks_stats.c
+++ b/tar/chunks/chunks_stats.c
@@ -414,3 +414,51 @@ chunks_stats_free(CHUNKS_S * C)
 	free(C->cachepath);
 	free(C);
 }
+
+/**
+ * chunks_initialize(const char * cachepath):
+ * Initialize the chunk directory (file) in ${cachepath}.  Return 0 on
+ * success, -1 on error, and 1 if the file already exists.
+ */
+int
+chunks_initialize(const char * cachepath)
+{
+	RWHASHTAB * HT;
+	struct chunkstats stats_extra;
+
+	/* Bail if ${chunkpath}/directory already exists. */
+ 	switch (chunks_directory_exists(cachepath)) {
+	case 0:
+		break;
+	case 1:
+		goto fileexists;
+	case -1:
+		goto err0;
+	}
+
+	/* Allocate empty hash table, and zero stats. */
+	if ((HT = rwhashtab_init(0, 1)) == NULL)
+		goto err0;
+	chunks_stats_zero(&stats_extra);
+
+	/* Write empty directory file. */
+	if (chunks_directory_write(cachepath, HT, &stats_extra, ""))
+		goto err1;
+
+	/* Free memory. */
+	rwhashtab_free(HT);
+
+	/* Success! */
+	return (0);
+
+err1:
+	rwhashtab_free(HT);
+err0:
+	/* Failure! */
+	return (-1);
+
+fileexists:
+
+	/* Failure! */
+	return (1);
+}
--- a/tar/glue/tape.c
+++ b/tar/glue/tape.c
@@ -232,6 +232,32 @@ err0:
 }
 
 /*
+ * Initialize cache directory.
+ */
+void
+tarsnap_mode_initialize_cachedir(struct bsdtar *bsdtar)
+{
+
+	switch (statstape_initialize(bsdtar->machinenum, bsdtar->cachedir)) {
+	case 0:
+		break;
+	case 1:
+		bsdtar_warnc(bsdtar, 0, "Cache directory already initialized");
+		/* FALLTHROUGH */
+	default:
+		goto err0;
+	}
+
+	/* Success! */
+	return;
+
+err0:
+	/* Failure! */
+	bsdtar->return_value = 1;
+	return;
+}
+
+/*
  * Nuke all the files belonging to an archive set.
  */
 void
--- a/tar/multitape/multitape.h
+++ b/tar/multitape/multitape.h
@@ -205,6 +205,14 @@ int statstape_close(TAPE_S *);
 int fscktape(uint64_t, const char *, int, int);
 
 /**
+ * statstape_initialize(machinenum, cachedir):
+ * Initialize an empty chunk directory in ${cachedir} so that --print-stats
+ * works.  This requires the "directory" file, but no other files.  Return 0
+ * on success, -1 on error, and 1 if the cachedir is already initialized.
+ */
+int statstape_initialize(uint64_t, const char *);
+
+/**
  * recovertape(machinenum, cachedir, whichkey):
  * Complete any pending checkpoint or commit, including a checkpoint in a
  * write transaction being performed by a different machine (if any).  If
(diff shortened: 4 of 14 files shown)


tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,45 @@ chunks_stats_free(CHUNKS_S * C)
...
 }
+
+/**
+ * chunks_initialize(const char * cachepath):
+ * Initialize the chunk directory (file) in the ${cachepath}.
+ */
+int
+chunks_initialize(const char * cachepath)
+{
+	struct chunks_stats_internal * C;
```
cperciva: We're not returning a cookie to be passed back to us later, so we don't need this.

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,45 @@ chunks_stats_free(CHUNKS_S * C)
...
+
+	/* Allocate memory. */
+	if ((C = malloc(sizeof(struct chunks_stats_internal))) == NULL)
+		goto err0;
+
+	/* Create a copy of the path. */
+	if ((C->cachepath = strdup(cachepath)) == NULL)
+		goto err1;
+
+	/* Read directory (a useful way to initialize C). */
```
cperciva: Unnecessarily apologetic.

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,45 @@ chunks_stats_free(CHUNKS_S * C)
...
+	void * dir;
+
+	/* Allocate memory. */
+	if ((C = malloc(sizeof(struct chunks_stats_internal))) == NULL)
+		goto err0;
+
+	/* Create a copy of the path. */
+	if ((C->cachepath = strdup(cachepath)) == NULL)
+		goto err1;
+
```
cperciva: Check that the ${cachepath}/directory doesn't exist.  We should throw an error if it does, because the user is probably making a mistake...

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,45 @@ chunks_stats_free(CHUNKS_S * C)
...
+	/* Read directory (a useful way to initialize C). */
+	if ((C->HT = chunks_directory_read(cachepath, &dir,
+	    &C->stats_unique, &C->stats_total, &C->stats_extra, 0, 1)) == NULL)
+		goto err2;
+	C->dir = dir;
+
+	/* Write directory file to filesystem. */
+	if (chunks_fsck_end(C))
```
cperciva: Eh?  Why are we detouring through _fsck_ code here?

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,45 @@ chunks_stats_free(CHUNKS_S * C)
...
+		goto err0;
+	if (chunks_transaction_commit(cachepath))
```
cperciva: Not necessary here.  The point of the `_transaction_commit` bit is to have state kept consistent with the server even in the event of an ill-timed crash; but we're not committing anything on the server here.

tar/chunks/chunks_stats.c
gperciva: 1. because that function contains `chunks_directory_write()`, and frees the `CHUNK_S` cookie.  Despite the name, it doesn't look particularly "fsck-related" to me.

2. because I can't resist calling a function whose description contains "... , and close the fscking cookie".  :)

tar/chunks/chunks_stats.c
gperciva: if I use `chunks_fsck_end` (which might not be suitable), then I'm left with a `directory.tmp` file.  `chunks_transaction_commit` renames that to `directory`.  (in a crash-proof way)

gperciva:
Updated PR.  Since the chunk, multitape, and glue layers changed so much, I merged the new code with the old commits, instead of adding them as new commits.  Future changes will be added as new commits (unless the whole architecture changes).

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,50 @@ chunks_stats_free(CHUNKS_S * C)
...
+chunks_initialize(const char * cachepath)
+{
+	char * directory_filename;
+	RWHASHTAB * ht;
+	struct chunkstats stats_extra;
+
+	/* Bail if ${chunkpath}/directory already exists. */
+	if (asprintf(&directory_filename, "%s/directory", cachepath) == -1)
+		goto err0;
+	if (access(directory_filename, F_OK) == 0)
```
cperciva: Use `stat`, check for ENOENT.

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,50 @@ chunks_stats_free(CHUNKS_S * C)
...
+/**
+ * chunks_initialize(const char * cachepath):
+ * Initialize the chunk directory (file) in the ${cachepath}.  Return 0 on
+ * success, -1 on error, and 1 if the file already exists.
+ */
+int
+chunks_initialize(const char * cachepath)
+{
+	char * directory_filename;
+	RWHASHTAB * ht;
```
cperciva: It's `HT` everywhere else in this file.  Why different here?

tar/chunks/chunks_stats.c
```diff
@@ -414,3 +414,50 @@ chunks_stats_free(CHUNKS_S * C)
...
+	struct chunkstats stats_extra;
+
+	/* Bail if ${chunkpath}/directory already exists. */
+	if (asprintf(&directory_filename, "%s/directory", cachepath) == -1)
```
cperciva: This might make sense as a `chunks_directory_exists` function?  That's the only file where the name "/directory" appears.

tar/multitape/multitape.h
```diff
@@ -205,6 +205,14 @@ int statstape_close(TAPE_S *);
 int fscktape(uint64_t, const char *, int, int);
 
 /**
+ * initialize_cachedir(machinenum, cachedir):
+ * Initialize an empty chunk directory in ${cachedir} so that --print-stats
+ * works.  This requires the "directory" file, but no other files.  Return 0
+ * on success, -1 on error, and 1 if the cachedir is already initialized.
+ */
+int initialize_cachedir(uint64_t, const char *);
```
cperciva: This seems like an odd name as far as multitape function goes.  Something like `statstape_initialize` might make more sense?

tar/multitape/multitape_fsck.c
```diff
@@ -573,3 +573,26 @@ fscktape(uint64_t machinenum, const char * cachedir, int prune, int whichkey)
...
+ * initialize_cachedir(machinenum, cachedir):
+ * Initialize an empty chunk directory in ${cachedir} so that --print-stats
+ * works.  This requires the "directory" file, but no other files.  Return 0
+ * on success, -1 on error, and 1 if the cachedir is already initialized.
+ */
+int
+initialize_cachedir(uint64_t machinenum, const char * cachedir)
+{
+	int rc;
+
```
cperciva: `(void) machinenum; /* UNUSED */`

tar/glue/tape.c
```diff
@@ -232,6 +232,33 @@ tarsnap_mode_fsck(struct bsdtar *bsdtar, int prune, int whichkey)
...
+void
+tarsnap_mode_initialize_cachedir(struct bsdtar *bsdtar)
+{
+
+	switch (initialize_cachedir(bsdtar->machinenum, bsdtar->cachedir)) {
+	case 0:
+		break;
+	case 1:
+		bsdtar_warnc(bsdtar, 0, "Cache directory already initialized");
+		bsdtar->return_value = 1;
```
cperciva: Not necessary here since err0 sets it...

gperciva:
Updated PR, with one commit per comment.  It needs history rewriting before a merge, of course.

I'm not totally sold on the name `statstape_initialize`, but I admit that the previous name wasn't appropriate either.  I've changed it as you requested.

tar/chunks/chunks_directory.c
```diff
@@ -362,6 +362,46 @@ chunks_directory_write(const char * cachepath, RWHASHTAB * HT,
...
+	}
+
+	/* Clean up memory. */
+	free(directory_filename);
+
+	/* File does not exist. */
+	return (1);
+
+fileexists:
+	/* File exists. */
```
cperciva: memory leak?

tar/chunks/chunks_directory.c
```diff
@@ -362,6 +362,46 @@ chunks_directory_write(const char * cachepath, RWHASHTAB * HT,
 }
 
 /**
+ * chunks_directory_exists(cachepath):
+ * Return 0 if the /directory file exists within ${cachepath}, 1 if it does
```
cperciva: Seems like it would make more sense to return 1 for `_exists` if the file exists?

gperciva:
PR updated.  I think it might need another round of review after I've rewritten all this history.

Note that I did not change the returned values from `chunks_initialize`, even though `chunks_directory_exists` changed.  (I did update the `switch` statement inside `chunks_directory_exists`, though)

tar/chunks/chunks_directory.c
```diff
@@ -362,6 +362,48 @@ chunks_directory_write(const char * cachepath, RWHASHTAB * HT,
...
+		goto err0;
+
+	/* Check if file exists. */
+	if (stat(directory_filename, &sb) == 0)
+		goto fileexists;
+	else {
+		if (errno != ENOENT)
+			goto err1;
+	}
+
```
cperciva: Given the three-way replication of freeing `directory_filename`, I'm tempted to break the usual style rule of "one return() statement per meaning" here.  `int rc;` which gets set to 0/1/-1, then `/* Return result code. */` `return (rc);`...

cperciva: We should probably also `warnp("stat(%s)", directory_filename)` in the not-ENOENT error case, btw.

gperciva:
Updated PR.  Includes #151 , which will probably be merged before this one, so I can remove it from this PR when I clean up the history.

cperciva:
Looks good.  Please clean up the history, then I'll merge it.

gperciva:
Updated PR with clean history, and added a mention to the NEWS file.

PR #150, opened 2016-03-23
fedora spec

gperciva:
(this is a continuation of #142; I'm not entirely certain of the most appropriate way to add patches to somebody else's PR)

### DIFF
--- a/pkg/fedora/tarsnap.spec
+++ b/pkg/fedora/tarsnap.spec
@@ -1,6 +1,6 @@
 Name:           tarsnap
 Version:        1.0.36.1
-Release:        2%{?dist}
+Release:        3%{?dist}
 Summary:        Secure, efficient online backup service
 
 License:        Proprietary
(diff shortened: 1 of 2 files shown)


gperciva:
I found a number of incompatibilities with Centos 5 and 6.  I was trying to fix all those issues and have a unified spec file, but I've now changed my mind.  After checking with @cperciva, I've decided to add a separate spec file for centos (to deal with things like the `%make_install` vs. `%makeinstall` disaster, which I thought was only relevant for OpenSUSE, but apparently not), while keeping the fedora spec file cleaner.

I added a note to the README.md warning about this.

pkg/fedora/tarsnap.spec
```diff
@@ -51,6 +53,10 @@ mv $RPM_BUILD_ROOT/%{_sysconfdir}/tarsnap.conf.sample \
 
 
 %changelog
+* Mon Mar  7 2016 Graham Percival <gperciva@tarsnap.com> 1.0.37-1%{?dist}
```
bgilbert: `%{?dist}` is customarily omitted from changelog versions.

gperciva:
Thanks, fix proposed in #152.  I think I remember seeing centos 5 complain that the changelog version didn't match the package version, but now that we're planning a separate spec file for centos, that's not a concern.  And the fedora packaging guidelines are pretty clear about not using `%{?dist}` in `%changelog`.

PR #159, opened 2016-04-22
Free FILE* and buffer from util.c process_lines()

gperciva:
Shoving yet more variables into struct bsdtar feels a bit messy, although the
actual diff is relatively small.

The main problem is that that process_lines() allocates a FILE * and
char * buff, then calls configfile_helper(), which can invoke bsdtar_errc()
instead of returning an error value back to process_lines() so that it could
clean up after itself.

I think that a "cleaner" fix (i.e. returning an error value and cleaning up
local variables in an err1: within process_lines()) would involve a larger
change to libarchive code, but I'm open to ideas.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -170,6 +170,8 @@ bsdtar_init(void)
 	bsdtar->conffile = NULL;
 	bsdtar->conf_opt = NULL;
 	bsdtar->conf_arg = NULL;
+	bsdtar->conffile_actual = NULL;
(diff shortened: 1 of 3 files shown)


gperciva:
The whole "config file reading" code is changed in recent libarchive, with their `libarchive_fe/line_reader{.h|.c}` (libarchive_fe == "libarchive frontends").  Instead of a callback function, they simply provide lafe_line_reader_next(), which would make it much easier to clean up locally-allocated memory.

PR #160, opened 2016-04-22
Avoid concatenating multiple xattr read errors

gperciva:
This arose from running Tarsnap inside a jail:
    http://mail.tarsnap.com/tarsnap-users/msg01206.html

This patch doesn't address the root problem (FreeBSD not exposing xattr inside
a jail and/or Tarsnap not detecting it being in a jail and/or the user not
being able to specify a --no-xattr option (which was added to libarchive a few
days ago)).  But at least the warning messages no longer look ridiculous.


gperciva:
If for some reason you want to see this in action, we can force an xattr error (instead of setting up a jail within which to run Tarsnap) with:
[force-xattr-error.patch.txt](https://github.com/Tarsnap/tarsnap/files/232378/force-xattr-error.patch.txt)
(grr, github doesn't allow us to attach `.patch` files?!)

Issue #164, opened 2016-06-06
humanize-numbers doesn't affect SIGINFO output


gperciva:
The `siginfo_printinfo()` function in `tar/siginfo.c` should obey the `--humanize-numbers` command.

gperciva:
Arises from: http://mail.tarsnap.com/tarsnap-users/msg01264.html

marrie:
I hope this can be looked into as not all of us want to convert the bite count. I suck at doing that on the fly.. Thanks for looking into this.

gperciva:
Fixed in #166.

PR #166, opened 2016-06-09
Apply --humanize-numbers to SIGINFO output

gperciva:


gperciva:
Addresses #164 .  One person suggested giving the byte count as well (i.e. "2.1 GB / 3.0 GB (2100123456 / 3000123456)"), but on second thought I'm not convinced that's necessary.  I can add it easily if desired, though.

cperciva:
The code looks good, but shouldn't this be documented somewhere?

cperciva:
Err, whoops.  I didn't see the man page updates at the bottom...

cperciva:
Should probably get a mention in NEWS anyway though, since it's a clearly user-visible change.

PR #170, opened 2016-07-03
build: add -D_POSIX_C_SOURCE=200809L

gperciva:
Without this, clang-3.8 with -std=c99 on Ubuntu 16.04 fails to find SSIZE_MAX
in limits.h.  This is because SSIZE_MAX is in .../bits/posix1_lim.h, and
/usr/include/limits.h contains:

    #ifdef  __USE_POSIX
    /* POSIX adds things to <limits.h>.  */
    # include <bits/posix1_lim.h>
    #endif

and in "\__STRICT_ANSI__" mode (implied by -std=c99, amongst other things),
__USE_POSIX is only defined if we've defined _POSIX_C_SOURCE or _XOPEN_SOURCE.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -244,6 +244,7 @@ lib_libtarsnap_a_CPPFLAGS=				\
 		-I$(top_srcdir)/libcperciva/network	\
 		-I$(top_srcdir)/libcperciva/util	\
 		-DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\" \
+		-D_POSIX_C_SOURCE=200809L		\
 		-DPOSIXFAIL_MSG_NOSIGNAL \
 		-DPOSIXFAIL_CLOCK_REALTIME
 


gperciva:
This is copied from https://github.com/Tarsnap/scrypt/pull/55

cperciva:
I'm slightly worried that this may break other platforms, by convincing them to hide non-POSIX functions... I'm going to merge it for now, but be aware that there may be fallout later.

gperciva:
Hmm?  `SSIZE_MAX` is notated as being `[CX]`:

> [CX] Some of the functionality described on this reference page extends the ISO C standard. Applications shall define the appropriate feature test macro (see XSH The Compilation Environment ) to enable the visibility of these symbols in this header.
> ...
> [CX] ￼ Extension to the ISO C standard ￼
The functionality described is an extension to the ISO C standard. Application developers may make use of an extension as it is supported on all POSIX.1-2008-conforming systems.
> http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

So it makes sense to me that we need to use `-D_POSIX_C_SOURCE=200809L`.  We use it in cperciva software... I feel like I'm missing something here?

PR #175, opened 2016-08-04
Require that a --configfile exists

gperciva:
We (continue to) silently ignore a non-existent "default" config file that is
not explicitly passed via --configfile.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -1237,10 +1237,20 @@ configfile(struct bsdtar *bsdtar, const char *fname, int fromcmdline)
 	bsdtar->option_no_config_include_set =
 	    bsdtar->option_no_config_include;
 
-	/* If the file doesn't exist, do nothing. */
 	if (stat(fname, &sb)) {
-		if (errno == ENOENT)
-			return;
+		/* Missing file. */
+		if (errno == ENOENT) {
+			if (fromcmdline) {
+				bsdtar_errc(bsdtar, 1, errno,
+				    "Cannot read config file: %s", fname);
+			} else {
+				/*
+				 * If the file wasn't specified on the
+				 * command-line, do nothing.
+				 */
+				return;
+			}
+		}
 
 		/*
 		 * Something bad happened.  Note that this could occur if
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -106,7 +106,8 @@ time_t get_date(time_t, const char *);
 static struct bsdtar	*bsdtar_init(void);
 static void		 bsdtar_atexit(void);
 
-static void		 configfile(struct bsdtar *, const char *fname);
+static void		 configfile(struct bsdtar *, const char *fname,
+			     int fromcmdline);
 static int		 configfile_helper(struct bsdtar *bsdtar,
 			     const char *line);
 static void		 dooption(struct bsdtar *, const char *,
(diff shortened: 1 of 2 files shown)


tar/bsdtar.c
```diff
@@ -1251,7 +1252,7 @@ configfile(struct bsdtar *bsdtar, const char *fname)
 		 * than failing to report an error if there really is a
 		 * configuration file.
 		 */
-		bsdtar_errc(bsdtar, 1, errno, "stat(%s)", fname);
+		bsdtar_errc(bsdtar, 1, errno, "configfile stat(%s)", fname);
```
gperciva: I added the "configfile" because I found the error message unclear (even when I was expecting to see it because I was testing that case).  Non-programmers (or more specifically, non-posix C programmers) don't know what "stat" means.

cperciva: I'd keep the `stat(%s)` error message here but add a more user-friendly one for the fromcmdline && ENOENT case.

gperciva:
You're probably going to tell me that you'd rather see
```
if (!fromcmdline && (errno == ENOENT))
        return;
if (fromcmdline && (errno == ENOENT))
        bsdtar_errc(...
```
?  Despite the fact that it's not as readable as nested `if` statements?  :)

(also, I wasn't certain whether the "if there's only one line, omit the `{...}`" applied if there was a multi-line comment in there as well.)

cperciva:
On 08/07/16 10:43, Graham Percival wrote:
> You're probably going to tell me that you'd rather see
> 
> |if (!fromcmdline && (errno == ENOENT)) return; if (fromcmdline && (errno ==
> ENOENT)) bsdtar_errc(... |
> 
> ? Despite the fact that it's not as readable as nested |if| statements? :)

No, nested if is the right answer here.

> (also, I wasn't certain whether the "if there's only one line, omit the
> |{...}|" applied if there was a multi-line comment in there as well.)

If there's a multiline comment, then there's more than one line.

-- 
Colin Percival
Security Officer Emeritus, FreeBSD | The power to serve
Founder, Tarsnap | www.tarsnap.com | Online backups for the truly paranoid

gperciva:
ok, rebased history.

tar/bsdtar.c
```diff
@@ -1236,10 +1237,20 @@ configfile(struct bsdtar *bsdtar, const char *fname)
...
 
-	/* If the file doesn't exist, do nothing. */
 	if (stat(fname, &sb)) {
-		if (errno == ENOENT)
-			return;
+		/* Missing file. */
+		if (errno == ENOENT) {
+			if (fromcmdline) {
+				bsdtar_errc(bsdtar, 1, errno,
+				    "Missing --configfile %s", fname);
```
cperciva: This prints `tarsnap: Missing -configfile foo: No such file or directory`, I think?  Maybe changing the errmsg string here to "Cannot read configuration file: %s" would be better?

cperciva:
This should be documented in `NEWS.md` (and thus in the release notes) since it's a change in behaviour: If someone had `--configfile /nosuchfile` then tarsnap will stop working for them.

Also, let's call this a $50 bug bounty; I'd call it non-harmless since silently ignoring an absent configuration file could result in behaviour significantly different from what the user wanted in the event that they typo the command line.

gperciva:
Updated PR with NEWS, more granular history (which you may or may not like), and somewhat altered error string.

It wasn't clear whether you were suggesting:

    tarsnap: Cannot read config file: /nonsuchfile: No such file or directory

or

    tarsnap: Cannot read config file: /nonsuchfile

I think we should print the errno string (`No such file or directory`), since users should recognize that as a standard error string.  This does mean that the first line has 3 colons in it, which seems excessive, but so far I've been unable to come up with a convincingly better alternative.

cperciva:
Yes, I meant the three-colon version.  It's verbose but consistent with unix conventions.

PR #176, opened 2016-08-04
Man extra programs

gperciva:


keygen/tarsnap-keygen.1-mdoc
```diff
@@ -10,9 +10,9 @@
 .Xr tarsnap 1
 .Sh SYNOPSIS
 .Nm
-.Fl -keyfile Ar key-file
-.Fl -user Ar user-name
-.Fl -machine Ar machine-name
+.Fl -keyfile Ar filename
```
gperciva: This is a draft PR for discussion.

I've gone back and forth about "--keyfile key-file" vs. "--keyfile filename" a few times.  I like the look of the paragraph description better with "filename":
> The --keyfile **filename** option specifies the **filename** in which to write the newly-generated keys.

However, I must admit that you're pretty consistent in using `key-file` in other user-facing info (man pages, `-h` output, etc.), so it wouldn't be good to change it in just this one place.  Maybe we could compromise on using `key-filename` ?

(I must admit that a great deal of my enthusiasm for `filename` comes from when I was writing a lot of python in academia, and I was fed up with getting files and filenames confused, so I resolved to always use "filename" (or "foo_filename" when I was referring to a string, and "foo_file" for file objects.)

cperciva: I'd like to stick with `key-file`.  And writing

> The --keyfile **key-file** option specifies the name of the file in which to write the newly-generated keys.

seems entirely grammatical...?

cperciva:
Forgot to add, I'm going to call this a $1 bounty for each unique command-line-option-clarification.  (So if you copy and paste text between man pages, that's just one bounty, but otherwise one bounty per option you're documenting.)

gperciva:
Updated PR

keyregen/tarsnap-keyregen.1-mdoc
```diff
@@ -11,7 +11,7 @@
 .Sh SYNOPSIS
 .Nm
 .Fl -keyfile Ar key-file
-.Fl -oldkey Ar old-key-file
+.Fl -oldkey Ar old-key-filename
```
cperciva: I thought we were sticking with `key-file` rather than `key-filename`?

gperciva: whoops, editing fail.

gperciva:
Updated PR.

Issue #181, opened 2016-08-13
fails to build on freebsd due to _POSIX_C_SOURCE


gperciva:
@cperciva did warn of fallout from #170, but I didn't imagine it would be like this...

freebsd 10.3-RELEASE-p4

```
$ make
...
In file included from ../lib/util/getfstype.c:7:
In file included from /usr/include/sys/mount.h:36:
/usr/include/sys/ucred.h:83:2: error: unknown type name 'u_int'
        u_int   cr_version;             /* structure layout version */
```

This is because `u_int` is only defined in `sys/types.h` if `__BSD_VISIBLE` is defined, and _POSIX_C_SOURCE stops __BSD_VISIBLE from being defined.

There's an existing PR and discussion from 2011:
https://lists.freebsd.org/pipermail/freebsd-bugs/2011-April/044044.html
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=156637

wollman wrote:
> I'd say it's the application code that is in error.  It should not be defining _XOPEN_SOURCE and then including (implementation private) header files which are not defined in the relevant standard.
>
> Do we seriously need to start writing our headers like:
>
> #include <sys/cdefs.h>
> #ifndef __BSD_VISIBLE
> #error "This is a non-standard header, but you have specified strict standard compliance."
> #endif

(in our case, we're definined _POSIX_C_SOURCE rather than _X_OPEN_SOURCE, but it's the same issue)

gperciva:
Thinking about this a bit more, I'm confused.  I view "`-D_POSIX_C_SOURCE`" as saying "I need to use POSIX features, not just standard C features".  But it seems that some people view "`-D_POSIX_C_SOURCE`" as saying "I promise that I'm not using anything other than POSIX"?

One point towards the "I promise I'm not using anything other than POSIX" is this:

> When an application includes a header described by POSIX.1-2008, and when this feature test macro is defined to have the value 200809L:
> ...
> Additional symbols not required or explicitly permitted by POSIX.1-2008 to be in that header shall not be made visible, except when enabled by another feature test macro.
> http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_02_01_01

However, `sys/mount.h' isn't in POSIX, nor is `statfs`, so I guess this part of the spec doesn't apply?

I notice that I am confused.

cperciva:
[...]

PR #182, opened 2016-08-14
build: getfstype.c, memlimit.c requires non-POSIX

gperciva:


gperciva:
Fixes #181 with `#undef _POSIX_C_SOURCE`.

It seems plausible that getting info about filesystems and memory limits requires non-portable code.

cperciva:
Hmm, "We use" seems like it would be more idiomatic than "We require"?

gperciva:
If we could get the filesystem type with POSIX functions, wouldn't we do that?

Anyway, updated with "We use".

cperciva:
Well, we *should* use only POSIX functions if we can do what we need that way.  But it's best to be clear about what we're saying.

PR #183, opened 2016-08-14
man: use imperative mood

gperciva:


gperciva:
This arose from #174.

gperciva:
There are still a few imperative forms in bsdtar man-page material, such as https://github.com/Tarsnap/tarsnap/blob/master/tar/tarsnap.1-mdoc.in#L365 .  I left those alone because I figured we didn't want "cosmetic" changes to libarchive source, but I could change those if desired.

cperciva:
Thanks.  I agree, don't reword text we inherited from libarchive.

PR #185, opened 2016-08-16
Add --no-keep-going option

gperciva:


gperciva:
Based on your comment about `--no-iso-dates`, I think we should have `--no-keep-going` as well?

gperciva:
Should this also be added to the NEWS file?  I'm leaning towards "no".

gperciva:
As a happy accident, this now allows `--keep-going` to exist in a configfile, instead of only being accepted as a command-line argument.  That has let me understand more about how we process options.

These commands (mostly from libarchive) are also "command-line only, not config file", and it's not clear to me that they shouldn't be allowed in a config file:
- `--check-links`
- `-H`, `-L`
- `--noisy-warnings`
- `--numeric-owner`
- `--one-file-system`

gperciva:
Updated with `--no-iso-dates`

cperciva:
So, my thought when we added the --keep-going option was that it wasn't something which should be included in a configuration file.  And of course if it can't be in a configuration file then we don't need a --no- version of it.

Can you see a situation where it would make sense to have this in a configuration file?

gperciva:
I can't think of a great reason to have it in the config file, but in that case...

... huh.  I was about to complain that it should be documented that --keep-going can't be in the config file, but I just discovered `tar/tarsnap.conf.5*.in`.  I haven't been updating that for things like `iso-dates` or `retry-forever`, but I'm doing that now.

PR #188, opened 2016-08-17
Fix warnings

gperciva:


gperciva:
There are many more warning fixes to come; this is a subset which I'm fairly confident about.

lib/netpacket/netpacket_write.c
```diff
@@ -80,7 +80,7 @@ netpacket_write_file(NETPACKET_CONNECTION * NPC,
 	packetbuf[8] = class;
 	memcpy(&packetbuf[9], name, 32);
 	memcpy(&packetbuf[41], nonce, 32);
-	be32enc(&packetbuf[73], buflen);
+	be32enc(&packetbuf[73], (uint32_t)buflen);
```
gperciva: Would you rather that this was

    buflen & 0xffffffff

instead?  There's a whole bunch of `be32enc()` and `le32enc()` places that currently use implicit casting.

lib/network/tsnetwork_bwlimit.c
```diff
@@ -199,9 +199,9 @@ network_bwlimit_get(int op, size_t * len)
 
 	/* Return the appropriate value. */
 	if (op == NETWORK_OP_READ)
-		*len = limit_read.bucket;
+		*len = (size_t)(limit_read.bucket + 0.5);
```
gperciva: Not certain if you wanted to round it, or simply truncate.  It's unlikely that anybody would notice the difference ("I wanted 100,000 bytes per second, not 99,999!  And my network connection is so powerful that I'm sure the difference is due to floating-point representation inexactness, not anything else!"), but I figured that rounding would be safer.

lib/datastruct/patricia.c
```diff
@@ -144,7 +144,7 @@ compare(const struct pnode * n, const uint8_t * key, size_t keylen,
 	for (i = 0; i < n->slen; i++) {
 		/* Is the key a prefix of the node? */
 		if (keylen == i) {
-			*mlen = i;
+			*mlen = i & 0xff;
```
cperciva: Hmm, why not just make `i` a `uint8_t`?

lib/netpacket/netpacket_write.c
cperciva: Given that we've validated `buflen` a few lines earlier, I'd go with the cast to `uint32_t`.

lib/network/tsnetwork_bwlimit.c
cperciva: Just stick to truncation.  You can't actually get large enough values to have floating-point roundoff, and even if you did it wouldn't result in non-integer values...

cperciva:
Get rid of the `+0.5` bit, and switch the patricia code to use a uint8_t I think... then figure out why github thinks I can't merge this.  It never used to be a problem...

lib/datastruct/patricia.c
gperciva: in other circumstances, you've told me that `size_t i` should remain as such.  For example,

> So, the idea here is that a "straightforward" reading of the code should match what it actually does. i is an array index, so it should be a size_t; where we set a byte equal to i or 0x11 * i what we really mean is "set this byte to the lowest byte of..." and & 0xff is a way of making that explicit rather than simply relying on the implicit type conversion.
https://github.com/Tarsnap/libcperciva/pull/23#discussion_r68013973

I am *entirely* in favour of `uint8_t i`.  :)

cperciva: Hey, I never claimed that I was consistent!  In this case we're iterating between 0 and a `uint8_t` value though, which makes it more natural to use a `uint8_t` value... even though we're eventually using it as an array index (which would normally suggest making it a `size_t`).

*shrug*

gperciva:
Updated PR

PR #190, opened 2016-08-24
Update DEBIAN.md

multiwebinc:
This prevents the error:

    W: There is no public key available for the following key IDs:
    1397BC53640DB551

This was taken from http://askubuntu.com/questions/766883/there-is-no-public-key-available-for-the-following-key-ids-1397bc53640db551. I really understand next to nothing about it, so I have no idea of the security implications. 

Or maybe `gpg --no-default-keyring --keyring ~/.gnupg/trustedkeys.gpg --import tarsnap-signing-key-2016.asc` needs to be done using sudo? No idea. The keyserver command worked for me though.

### DIFF
--- a/pkg/DEBIAN.md
+++ b/pkg/DEBIAN.md
@@ -22,6 +22,9 @@ To use the official Tarsnap Debian source packages, set up:
         sudo apt-key add tarsnap-signing-key-2016.asc
         gpg --no-default-keyring --keyring ~/.gnupg/trustedkeys.gpg \
           --import tarsnap-signing-key-2016.asc
+        sudo apt-key adv --keyserver keyserver.ubuntu.com \
+          --recv-keys 1397BC53640DB551
+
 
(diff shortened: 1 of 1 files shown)


gperciva:
Thanks for the patch!  I don't think this is an issue with Tarsnap, though.

Searching for `1397BC53640DB551` suggests that this is related to google chrome.  I just tested the debian package instructions on Ubuntu 16.04 and Debian 8 (from a fresh system each time), and didn't encounter any such problem.

multiwebinc:
Yeah, you're probably right. I could swear this wasn't there before I added the tarsnap repository, but I guess I just didn't notice it.

gperciva:
How often do you run `apt-get update`?  Maybe this was just the first time you ran that command after installing chrome?

multiwebinc:
I usually do `apt-get udpate && apt-get dist-upgrade` all in one shot, so it's possible that I have been missing it for a long time. I've had Chrome installed for probably 2 years.

PR #191, opened 2016-08-26
News: metion --iso-dates and --no-iso-dates

gperciva:


cperciva:
Can you re-push this without the typo in the commit message?

gperciva:
I changed it to "add" since "mention" is evidently too dififcult -- I mean, too hard.  :)

PR #192, opened 2016-08-26
man: add (no-)iso-date and (no-)retry-forever to .conf

gperciva:


gperciva:
Turns out we were only missing these 2 (or 4) options.

PR #194, opened 2016-08-27
Explicit cast to (uint32_t)

gperciva:


gperciva:
I'm not happy with this particular warning fix, and I'm *really* unhappy to be not happy when proposing any changes to crypto stuff, but I think that I've done as much as I can without your review.

gperciva:
last-minute thought: although the MGF1 algorithm doesn't specify an absolute limit on the size, I just noticed that the I20SP algorithm converts the counter to a 4-octet representation, and that algorithm contains:
> If x >= 256^xLen, output "integer too large" and stop.
where xLen is 4 in this case.

Why on earth couldn't they include that absolute limit (because xLen=4 for MGF1) in the MGF1 algorithm?!  :(

PR #197, opened 2016-08-27
Storage whichkey

gperciva:


cperciva:
My basic philosophy here was that I'd use `uint8_t` for the "network-facing" bits, since uint8_t is what actually goes into the network protocol packet; and `int` for the "user-facing" bits, since I use `int` basically as `enum` to represent "small numbers, especially ones we can count on our fingers".  Where exactly the switch between `int` and `uint8_t` should happen isn't immediately clear, of course.

So I'm not going to decide what to do with this immediately; take a look with the benefit of the above philosophy and if you still think this is appropriate I'll merge it, but I don't want to merge it now and then have you say "oh, in that case we should stick to int". :-)

gperciva:
That sounds fine.  I assume that you didn't have any sneaky reason to have both `uint8_t whichkey` and `int whichkey` in `storage.h`?  It looks like a simple mistyping, but you might have had some weird reason why `storage_fsck_start()` is different from `storage_transaction_checkpoint()`, `storage_transaction_commit()`, and `storage_transaction_commitfromcheckpoint()`.

As for "when to switch", I'm comfortable with `recovertape()` which currently has:
```
uint8_t key = (whichkey == 0) ? 0 : 1;
```
as such, I thought it likely that `fscktape()`'s
```
int key = (whichkey == 0) ? 0 : 1;
```
was a typo.

gperciva:
I've updated the commit messages in this PR to explain the changes in more detail.

PR #199, opened 2016-09-03
Fix warnings

gperciva:


tar/chunks/chunks_stats.c
```diff
@@ -335,8 +335,8 @@ chunks_stats_addchunk(CHUNKS_S * C, const uint8_t * hash,
 
 	/* Record the lengths if necessary. */
 	if (ch->d.nrefs == 0 && ch->ncopies_ctape == 0) {
-		ch->d.len = len;
-		ch->d.zlen_flags = zlen | (ch->d.zlen_flags & CHDATA_FLAGS);
+		ch->d.len = (uint32_t)len;
```
cperciva: Maybe add an assertion here?  In general assertions are good for "this is safe because all the users of this code behave" situations since they catch if a new use of the code misbehaves.  (Or, put another way, we should use assertions to check that APIs are being used correctly, and chunks_* is an API.)

gperciva: ok, added.  Did you want one for `zlen` as well?  I added one, although it seems a bit weird to be checking `C->zbuflen` (but that's where the value comes from, so if we want to check it anywhere, that's the place to check... although to be extra-safe, we'd want to check that `C != NULL` as well?  That strikes me as overkill, though.)

PR #211, opened 2016-10-27
libcperciva: put CPU macros inside relevant #ifdef

gperciva:
These macros are only used in code that is hidden by the same #ifdef.


gperciva:
This was part of https://github.com/Tarsnap/scrypt/pull/67

PR #212, opened 2016-10-31
Fix warnings

gperciva:


lib/keyfile/keyfile.c
```diff
@@ -384,15 +384,15 @@ keyfile_read(const char * filename, uint64_t * machinenum, int keys)
 		warn0("Key file has unreasonable size: %s", filename);
```
cperciva: Rather than adding so many casts, how about adding `size_t keylen`, moving the "unreasonable size" check up a couple lines with a new `/* Validate file size */` comment, and following it by `keylen = (size_t)(sb.st_size);`?

lib/keyfile/keyfile.c
```diff
@@ -372,27 +372,31 @@ keyfile_read(const char * filename, uint64_t * machinenum, int keys)
...
 	/* Stat the file. */
 	if (stat(filename, &sb)) {
 		warnp("stat(%s)", filename);
 		goto err0;
 	}
 
-	/* Allocate memory. */
-	if ((sb.st_size == 0) || (sb.st_size > 1000000)) {
+	/* Validate keyfile size. */
+	keyfilelen = (size_t)(sb.st_size);
```
gperciva: Do you want the parentheses around `(sb.st_size)` ?  My mental model of your normal style suggested that they shouldn't be there, but perhaps this clarifies that the `(size_t)` refers to the `.st_size` portion of the structure, not the `sb` structure itself?

lib/keyfile/keyfile.c
```diff
@@ -372,27 +372,31 @@ keyfile_read(const char * filename, uint64_t * machinenum, int keys)
...
+	if ((keyfilelen == 0) || (keyfilelen > 1000000)) {
```
gperciva: right, because it's `off_t`, not `size_t`.  I should have remembered that.

lib/keyfile/keyfile.c
cperciva: Parentheses are good.  The order of operations between `.` and `(type)` operators isn't something most people have memorized.

gperciva:
Updated PR

PR #214, opened 2016-11-07
Import warning fixes from scrypt

gperciva:


gperciva:
These all came from scrypt.

Commit https://github.com/Tarsnap/tarsnap/pull/214/commits/7f94539db2569e21254e316449bad87acaaad386 really belongs as part of https://github.com/Tarsnap/tarsnap/pull/214/commits/9b5591b65402217c429ad4d32273671b6954df41 , but I left them separate because that's how they appeared in scrypt.  Also, I just realized that the hash in the commit message for that "oops, I should have included this" commit is incorrect -- it refers to the hash in the original scrypt commit, but of course the same change has a different hash in Tarsnap.  Shall I merge those commits?

cperciva:
Yes, go ahead and merge those commits.  There's no need to mirror the history.

gperciva:
Updated PR.

PR #218, opened 2016-11-19
Typos

gperciva:


tar/cmdline.c
```diff
@@ -364,7 +364,7 @@ bsdtar_getopt(struct bsdtar *bsdtar)
 			/* Short-circuit if first chars don't match. */
 			if (popt->name[0] != opt_word[0])
 				continue;
-			/* If option is a prefix of name in table, record it.*/
+			/* If option is a prefix of name in table, record it. */
```
gperciva: Oops, that commit was included as part of https://github.com/Tarsnap/tarsnap/pull/220 which was merged.  Do you want a PR to revert that change?

cperciva: Oops.  Never mind then; even if we revert the change it will show up in "blame" as having been modified.

BTW, is there a way to avoid having overlapping PRs?  It's really confusing...

gperciva:
Updated PR

Issue #113, opened 2016-01-07
Tarsnap stats get mangled with recovered checkpoints


cperciva:
There's a glitch in the handling of the "all archives" statistics relating to recovered checkpoints.  Specifically, if an archive is recovered from a checkpoint the "extra" statistics for the metadata and metaindex blocks aren't included; but they're deducted off when that recovered archive is deleted.  This shows up when users delete _all_ of their archives, and then get told that they have 18.4 exabytes of data stored (aka. a small negative number interpreted as an unsigned 64-bit value).

Some code in the multitape layer needs to be fixed -- I think in creating checkpoints we'll need to increment the "extra" statistics before writing out the directory file and then adjust back.

gperciva:
Note to self: this can be illustrated without "recovered" or "deleted":

- (set up `checkpoint-bytes 1M`)

```
$ ./tarsnap -cvf a1 ~
... cancel once a checkpoint is written
$ ./tarsnap -cvf a1 ~
tarsnap: An archive already exists with the name "a1.part"
tarsnap: Error creating new archive
tarsnap: Error exit delayed from previous errors.
$ ./tarsnap --print-stats
                                       Total size  Compressed size
All archives                             22358736          1039660
  (unique data)                          12436176          1017924
$ ./tarsnap --fsck
...
$ ./tarsnap --print-stats
                                       Total size  Compressed size
All archives                             22359194          1040710
  (unique data)                          12436634          1018974
```

(the sizes increased after the `fsck`)

If we now delete `a1.part`, the size is 0 (as it should be).

cperciva:
> illustrated without "recovered"

Eh?  Tarsnap is recovering a partial archive from a checkpoint in your example.

gperciva:
I hate jargon.

gperciva:
This was fixed in #219.

Issue #217, opened 2016-11-17
No. No I don't have 18EB of data; go home tarsnap, you're drunk.


larryhynes:
tarsnap built from deb package, newly registered machine, first run of tarsnap:

```
$ tarsnap -cf foo ~
tarsnap: Removing leading '/' from member names
....
^C

(Set nodump flag on really big dir that I didn't want to back up)

$ tarsnap -cf foo ~
tarsnap: An archive already exists with the name "foo.part"
tarsnap: Error creating new archive
tarsnap: Error exit delayed from previous errors.

$ tarsnap --list-archives
foo.part

$ tarsnap -df foo.part
					Total size  Compressed size
All archives				18 EB            18 EB
  (unique data)				18 EB            18 EB
This archive				1.0 GB           997 MB
Deleted data				1.0 GB           997 MB

$ tarsnap -cf bar ~
tarsnap: Removing leading '/' from member names
tarsnap: Not adding cache directory to archive: /home/user/tarsnap_cache
					Total size  Compressed size
All archives				124 MB           100 MB
  (unique data)				123 MB           100 MB
This archive				124 MB           100 MB
New data					123 MB           100 MB

$ tarsnap --print-stats
					Total size  Compressed size
All archives				124 MB           100 MB
  (unique data)				123 MB           100 MB
```

Is 18EB just a placeholder value for when we don't know just how much data we have, or am I about to get a really big bill? :smiley:

gperciva:
Nice!  I can reproduce it, and should have a fix in a few minutes.

Since @cperciva doesn't think there should be legal action for an unpaid $0.00 bill (https://mobile.twitter.com/cperciva/status/799080890286542848), I think you'll be safe.  :)

cperciva:
This is issue #113, which I diagnosed years ago but never got around to fixing.  Hopefully @gperciva can take care of this. :-)

larryhynes:
All's well that ends well.

PR #219, opened 2016-11-19
Properly account for checkpoint metadata

gperciva:


gperciva:
The final 2/5 commits here are not necessary to fix #113 and #217, but they clean up the API slightly.

Note that instead of subtracting the metadata-related stats, I save ("back up") the original stats, modify the stats & write the directory file, then load ("restore") the original stats.  This isn't exactly what we discussed offline, but I think it's a cleaner solution than trying to get metadata stats info from `multitape_meta*_put()` into `writeape_checkpoint()` so that we could subtract those bytes from `C->stats_extra`.

tar/multitape/multitape_write.c
```diff
@@ -843,16 +843,33 @@ writetape_checkpoint(TAPE_W * d)
 		goto err0;
 
 	/*
+	 * Back up statistics about the metadata and metaindex;
```
cperciva: Back up archive set statistics before adding the metadata and metaindex.

tar/multitape/multitape_write.c
```diff
@@ -843,16 +843,33 @@ writetape_checkpoint(TAPE_W * d)
...
+	 */
+	chunks_write_extrastats_copy(d->C, 0);
+
+	/*
 	 * Flush data through and write the metaindex and metadata;
 	 * checkpoints are partial archives, so mark it as such.
+	 * This also adds the metadata and metaindex to the
+	 * "extra" statistics; we need these when we write the
+	 * directory file, but we will restore the original
+	 * statistics later.
```
cperciva: ... since the metadata and metaindex from this checkpoint will be discarded if/when another checkpoint is created or the archive is completed.

gperciva:
update PR with temp commits which fix the issues you raised.  If it looks good, I'll rewrite the history.

cperciva:
Looks good except that you seem to have an odd predisposition for early wrapping of lines in multiline comments.  Maybe you can fix that at the same time as you rewrite the history?

tar/chunks/chunks_write.c
```diff
@@ -249,6 +250,21 @@ chunks_write_extrastats(CHUNKS_W * C, size_t len)
...
+ * chunks_write_extrastats_copy(C, size_t direction):
+ * Make a copy of the extra stats; if ${direction} is 0, copy from
+ * the real stats to a copy; if 1, set the real stats to the copy.
+ */
+void
+chunks_write_extrastats_copy(CHUNKS_W * C, size_t direction)
+{
+
+	if (direction == 0)
+		chunks_stats_copy(&C->stats_extra, &C->stats_extra_copy);
```
cperciva: No, it was in C89.  It's just not used very often.

gperciva:
Updated PR.  The final 2/4 commits are not technically necessary to resolve this issue, but they clean up the API slightly by removing now-unnecessary arguments.

PR #220, opened 2016-11-29
Comments syntax

gperciva:


gperciva:
I'd like to draw your attention to the third patch, namely changing the 3 instances of

```
        /*-
         * blah blah
```

into `/*`.  I vaguely wondered if `/*-` was a convention for "here's a chatty explanation of what's happening".

(I retained `/*-` when it was used for the Copyright statement.)

Since I'm sure you're dying to know, I used this:
```
# Regex for valid beginnings of comments.
COMMENT_BEGIN = ["/\* ", "/\*\n", "/\*\*\n", "/\*-\n \* Copyright .*?\n"]
```

cperciva:
The `/*-` comments are correct -- that marks a multiline comment to which word-wrapping should not apply (because there's multiple paragraphs or ascii art or something else).

The same function is served by `/**` in the first column; comments marked with `/*-` in the first column are license text.

gperciva:
huh.  Ok, revised this PR to not include that third patch.

PR #221, opened 2016-12-10
build: add --with-conf-no-sample for tarsnap.conf

gperciva:
This will install "tarsnap.conf" instead of "tarsnap.conf.sample". Not recommended for end-users, as it would be easy to accidentally destroy one's tarsnap config.  However, this could be useful for a package manager (like apt-get) which has built-in handling for not overwriting user-edited config files.


cperciva:
I'm merging this on the basis of "looks plausible".  I assume you've tested both with and without this option and made sure that the autoconf gunk does what you want it to do.

PR #225, opened 2016-12-30
Openssl1.1

gperciva:


gperciva:
First draft.  There's still a warning about the deprecated `RSA_generate_key()`, but I think that belongs in a separate PR.

freddedotme:
Oh, I was just going to install tarsnap and I assume that's why I get this error?

    lib/crypto/crypto_keys_subr.c:178:23: error: incomplete definition of type 'struct rsa_st'

OpenSSL version:

    OpenSSL 1.1.0c  10 Nov 2016

gperciva:
Yes.  If you're on Debian testing (jessie), you can work around it with:
```
sudo apt-get install libssl1.0-dev
... install Tarsnap...
sudo apt-get install libssl-dev
```

Jessie includes libssl1.0.2, so once it's compiled, tarsnap will the 1.0.2 library.  But you can't have libssl1.0-dev and libssl-dev installed at the same time.  (I don't know why the Debian developers chose to do this, but presumably they have reasons.)

cperciva:
On 12/31/16 12:06, Graham Percival wrote:
> But you can't have libssl1.0-dev and libssl-dev installed at the same
> time. (I don't know why the Debian developers chose to do this, but presumably
> they have reasons.)

I'm guessing that they install header files into the same place?

-- 
Colin Percival
Security Officer Emeritus, FreeBSD | The power to serve
Founder, Tarsnap | www.tarsnap.com | Online backups for the truly paranoid

freddedotme:
I'm on a Macbook (MacOS 10.12.1) so I chose the latest version I could find on their website. I assume I have to downgrade it to 1.0 for now or when is this update supposed to occur?

gperciva:
At the moment, Tarsnap only works with with openssl < 1.1.0.  We are actively looking into what changed in openssl 1.1.0, and how best to support that version; however we cannot give an estimate on when the next release of Tarsnap will be.

lib/crypto/crypto_compat.c
```diff
@@ -26,3 +26,37 @@ crypto_compat_RSA_valid_size(RSA * rsa)
...
+int
+crypto_compat_RSA_import(RSA ** key, BIGNUM * n, BIGNUM * e, BIGNUM * d,
+    BIGNUM * p, BIGNUM * q, BIGNUM * dmp1, BIGNUM * dmq1, BIGNUM * iqmp)
+{
+
+	/* Sanity checks. */
+	assert(key != NULL);
+	assert((n != NULL) && (e != NULL));
+	/* All the private-key-related variables are NULL, or they're not. */
+	assert( ((d == NULL) && (p == NULL) && (q == NULL) && (dmp1 == NULL)
```
gperciva: Not certain what formatting you'd prefer here.  I personally would split up the nesting even more to make it more obvious what's happening, but I suspect you might tell me to remove all semantic spacing entirely.

gperciva:
Updated PR with a different way of organizing the compatibility layer.

lib/crypto/crypto_compat.c
```diff
@@ -0,0 +1,28 @@
...
+#include <openssl/rsa.h>
+
+#include "crypto_compat.h"
+
+/**
+ * crypto_compat_RSA_valid_size(rsa):
+ * Check that the RSA ${rsa} has the correct size.
+ */
+int
+crypto_compat_RSA_valid_size(RSA * rsa)
```
cperciva: const?

lib/crypto/crypto_compat.c
```diff
@@ -0,0 +1,28 @@
...
+{
+
+	/* Sanity checks. */
+	assert(rsa != NULL);
+	assert(rsa->n != NULL);
+
+	if ((RSA_size(rsa) != 256) || (BN_num_bits(rsa->n) != 2048))
```
cperciva: For such a trivial function, I think I'd write this as
`return ((RSA_size(rsa) == 256) && (BN_num_bits(rsa->n) == 2048));`

lib/crypto/crypto_compat.c
cperciva: How about
```
if (d != NULL)
    assert(stuff is not NULL);
else
    assert(stuff is NULL);
```
?  That seems easier to understand than a mess `&&`s and `||`s.

lib/crypto/crypto_compat.c
```diff
@@ -26,3 +26,37 @@ crypto_compat_RSA_valid_size(RSA * rsa)
...
+ */
+int
+crypto_compat_RSA_import(RSA ** key, BIGNUM * n, BIGNUM * e, BIGNUM * d,
+    BIGNUM * p, BIGNUM * q, BIGNUM * dmp1, BIGNUM * dmq1, BIGNUM * iqmp)
+{
+
+	/* Sanity checks. */
+	assert(key != NULL);
+	assert((n != NULL) && (e != NULL));
+	/* All the private-key-related variables are NULL, or they're not. */
```
cperciva: Blank line before comment.

lib/crypto/crypto_compat.c
```diff
@@ -60,3 +60,40 @@ crypto_compat_RSA_import(RSA ** key, BIGNUM * n, BIGNUM * e, BIGNUM * d,
...
+int
+crypto_compat_RSA_export(RSA * key, const BIGNUM ** n, const BIGNUM ** e,
+    const BIGNUM ** d, const BIGNUM ** p, const BIGNUM ** q,
+    const BIGNUM ** dmp1, const BIGNUM ** dmq1, const BIGNUM ** iqmp)
+{
+
+	/* Sanity check. */
+	assert(key != NULL);
+	assert((n != NULL) && (e != NULL));
+	/* All the private-key-related variables are NULL, or they're not. */
```
cperciva: See earlier comments.

lib/crypto/crypto_compat.c
```diff
@@ -62,9 +63,29 @@ crypto_compat_RSA_import(RSA ** key, BIGNUM * n, BIGNUM * e, BIGNUM * d,
 		(*key)->dmq1 = dmq1;
 		(*key)->iqmp = iqmp;
 	}
+#else
+	/* It's ok for d to be NULL here. */
```
cperciva: I'd be inclined to put `if (d == NULL)` first and have two `RSA_set0_key` calls, one of which takes an explicit NULL (with a comment indicating that NULL means we have only the public key).

gperciva:
Rebuilt on top of the new crypto-pre #229 PR.  As usual, the "REBASE" commits are fixing problems that you found.

cperciva:
Is there a way for me to look at this without the commits I've already merged as part of a different PR?  Git is confusing.

gperciva:
OK, I rebased this on top of master.  You've seen the commits without "REBASE" in front of them; the 3 commits with "REBASE" are in response to your comments on the non-REBASE ones.

Or, at this point, perhaps it would be best for me just rewrite the history, and you can look at it from scratch?

lib/crypto/crypto_keys_subr.c
```diff
@@ -208,14 +209,8 @@ crypto_keys_subr_import_RSA_priv(RSA ** key, const uint8_t * buf, size_t buflen)
...
-	(*key)->n = n;
-	(*key)->e = e;
-	(*key)->d = d;
-	(*key)->p = p;
-	(*key)->q = q;
-	(*key)->dmp1 = dmp1;
-	(*key)->dmq1 = dmq1;
-	(*key)->iqmp = iqmp;
+	if (crypto_compat_RSA_import(key, n, e, d, p, q, dmp1, dmq1, iqmp))
+		goto err1;
```
gperciva: Incidently, adding this `err1` fixes a compile warning in current Tarsnap master: At the moment, we skip from err0 to err2 and never use err1, which causes clang to complain about an unused label.

I'm choosing to claim that I intentionally wrote the commits this way, since it minimizes the diffs.  That said, in the future I'll make sure that such things are mentioned in the commit messages, are resolved within a single PR, and I'll enable stricter checking in `.travis.yml`.

TimWolla:
> As usual, the "REBASE" commits are fixing problems that you found.

Tip for your git usage that might save you some time: If you commit using `git commit --fixup <commit hash>` git will automatically create commits that are compatible with the `--autosquash` option of `git rebase --interactive`.

lib/crypto/crypto_compat.h
```diff
@@ -0,0 +1,27 @@
+#ifndef _CRYPTO_COMPAT_H_
+#define _CRYPTO_COMPAT_H_
+
+/**
+ * crypto_compat_RSA_valid_size(rsa):
+ * Check that the RSA key ${rsa} has the correct size.
```
cperciva: This is really a true-or-false question, not a success-or-failure question.

lib/crypto/crypto_compat.c
```diff
@@ -0,0 +1,147 @@
...
+int
+crypto_compat_RSA_valid_size(const RSA * const rsa)
+{
+
+	/* Sanity checks. */
+	assert(rsa != NULL);
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+	assert(rsa->n != NULL);
+	if ((RSA_size(rsa) != 256) || (BN_num_bits(rsa->n) != 2048))
```
cperciva: How about `return ((RSA_size(rsa) != 256) || (BN_num_bits(rsa->n) != 2048));` ?  Or given the name of the function, maybe the negation of that...

lib/crypto/crypto_rsa.c
```diff
@@ -81,7 +82,7 @@ crypto_rsa_sign(int key, const uint8_t * data, size_t len,
 		goto err0;
 
 	/* Make sure the key and signature buffer are the correct size. */
-	if ((RSA_size(rsa) != 256) || (BN_num_bits(rsa->n) != 2048)) {
+	if (crypto_compat_RSA_valid_size(rsa)) {
```
cperciva: This would make more sense if the function returned nonzero for valid sizes and zero for invalid sizes, so that the code could be `if (!crypto_compat_RSA_valid_size(rsa)) { ...`.  Right now it reads as "if rsa has a valid size, warn that the size is incorrect".

lib/crypto/crypto_compat.c
gperciva: I thought that functions were supposed to have a single `return` statement?  Or is this function small enough to be an exception to that rule?

gperciva:
@TimWolla Thanks for the tip about `git commit --fixup`!  I'm not wild about the commit message, though -- I'd feel better if it identified a commit by hash rather than subject, and I'd like to have my own message.  For example, addressing Colin's latest request, I could name the commit either:

```
fixup! crypto_compat: begin, add crypto_compat_RSA_valid_size()
```
or
```
REBASE: invert return values in _RSA_valid_size()
```

I think the second is more informative, so I've opted to use that technique.  Rebasing commits later isn't a big task.  :)

gperciva:
Updated PR

lib/crypto/crypto_compat.c
cperciva: Yeah, if the entire content of the function fits into one line, there's no need for the successful-return / failure-return structure.  (And all the more so given that what we're doing here is returning a *value* rather than a *status*.)

gperciva:
Updated PR with two `return` statements.

gperciva:
re OPENSSL_VERSION_NUMBER: it looked ok on Ubuntu 12.04, but I figured we might as well add an `#ifndef` in there just to make sure it's defined.

cperciva:
Looks good.  Please tidy up the commit history, and I'll merge it.

gperciva:
Updated PR with tidied history.

gperciva:
Updated PR with some nitpicks; not ready for merging yet.

cperciva:
Looks good, please clean up the commit history.

gperciva:
Rebased and ready for merging.

Just so you don't get confused, compiling with `clang -Weverything -Werror` with openssl 1.1.0 will still fail with:
```
../lib/crypto/crypto_keys_subr.c:461:10: error: 'RSA_generate_key' is deprecated
      [-Werror,-Wdeprecated-declarations]
        *priv = RSA_generate_key(2048, 65537, NULL, NULL);
                ^
/usr/include/openssl/rsa.h:193:25: note: 'RSA_generate_key' has been explicitly
      marked deprecated here
DEPRECATEDIN_0_9_8(RSA *RSA_generate_key(int bits, unsigned long e, void
                        ^
```

Without `-Werror`, it's only a warning (of course).  Adding RSA key generation to the openssl_compat file will come once this PR is merged.

cperciva:
Did you forget to push?  I'm still seeing REBASE commits.

gperciva:
Whoops.  Updated PR for real.

PR #226, opened 2017-01-02
zsh completion for tarsnap 1.0.37

vapniks:
archive names may optionally be completed by setting the `archivefile` variable at the top of the file

### DIFF
--- a/_tarsnap
+++ b/_tarsnap
@@ -18,19 +18,19 @@ files=("$matchany" ':file:file:_files')
 # completes directories
 directories=("$matchany" ':dir:directory:{_path_files -/}')
 # completes --keyfile option
-keyfile=(/$'--keyfile\0'/ ':opts:options:(--keyfile)' "$matchany" ':file:file:_files')
+keyfile=(/$'--keyfile\0'/ ':opts:options:((--keyfile\:"specify key file"))' "$matchany" ':file:file:_files')
 # completes --cachedir option
-cachedir=(/$'--cachedir\0'/ ':opts:options:(--cachedir)' "$matchany" ':file:file:{_path_files -/}')
+cachedir=(/$'--cachedir\0'/ ':opts:options:((--cachedir\:"specify cache directory"))' "$matchany" ':file:file:{_path_files -/}')
 # completes --dry-run option
 typeset -a dryrunopt
-dryrunopt=(/$'--dry-run\0'/ ':opts:options:(--dry-run)')
+dryrunopt=(/$'--dry-run\0'/ ':opts:options:((--dry-run\:"dont really create an archive, just simulate doing so"))')
 # completes -f (archive) option 
 function _tarsnap_archives() {
     local -a archivesarr
     archivesarr=( "${(f)$(cat ${1})}" )
     compadd -a archivesarr
 }
-archives=(/$'-f\0'/ ':opts:options:(-f)' /$'[^\0]##\0'/ ":archive:archive:{_tarsnap_archives $archivelistfile}")
+archives=(/$'-f\0'/ ':opts:options:((-f\:"specify name of archive to operate on"))' /$'[^\0]##\0'/ ":archive:archive:{_tarsnap_archives $archivelistfile}")
 # completes patterns
 patterns=("$matchany" ':patterns:glob pattern matching files to be viewed/extracted:')
 # complete tar file with @ at beginning (for -c command). Need to make sure it doesnt match @@<FILE>.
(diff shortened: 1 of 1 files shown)


_tarsnap
```diff
@@ -0,0 +1,163 @@
+#compdef tarsnap
+
+# zsh completion function for tarsnap 1.0.37
```
gperciva: Could you add some hint for zsh users as to how they'd use this file?  i.e. should it be copied into `/usr/share/zsh/completion/` , or `/etc/zsh/`, or `$HOME/.zsh/` ?

_tarsnap
```diff
@@ -0,0 +1,163 @@
+#compdef tarsnap
```
gperciva: Could this file be put into `misc/zsh_completion/_tarsnap` ?  Or some other filename, but definitely within `misc/` and beginning with `zsh`.

_tarsnap
```diff
@@ -0,0 +1,163 @@
...
+
+# zsh completion function for tarsnap 1.0.37
+
+# Set the following variable to the path of a file containing the output from "tarsnap --list-archives" 
+# i.e. one archive name per line. If left blank then archive names will not be completed.
+local archivefile=
```
gperciva: A little nitpicky, but could this variable be renamed to something like `archivelist_file` ?  (or something similar, if underscores aren't allowed in zsh)
My first impression was that this was supposed to be a list of archives; I had to read the comment twice before I understood that it was supposed to point to a text file which contains a list of archive names.

_tarsnap
vapniks: I'll leave that to you :)

_tarsnap
vapniks: done

_tarsnap
vapniks: done

gperciva:
Hmm, if I enter zsh and type:

```
% tarsnap --hu[TAB]
```

(trying to get `--humanize-numbers`), it turns into

```
% tarsnap --nuke
```

Does zsh autocompletion work very differently than bash's?  I also tried

```
% tarsnap -c --hu[TAB]
```

and couldn't get it to complete `--humanize-numbers` (or `--dry-run`) in there.

vapniks:
That's because I used the SYNOPSIS in the man page for tarsnap as a guide. The `--humanize-numbers` option is only available for the `-c`, `-d`, `-t` and `-x` commands, after `--keyfile`, `--cachdir`, and `-f` have been specified.

gperciva:
Hmm, I see what you mean; the SYNOPSIS might be a bit misleading.  Far below, in the EXAMPLES section, it states

```
Note that the --keyfile and --cachedir options can be specified via the tarsnap.conf(5) configuration file, in which case they may be omitted from the command line.
```

In addition, the `-f` option can be mixed in with the other `[options]`.  It can even be omitted if `--dry-run` is specified.

Yes, I definitely think that the man page is misleading.  I'll revise it and discuss with @cperciva.

vapniks:
I did make `--keyfile` and `--cachedir` optional, but `-f` is currently required. I could easily change this though.

gperciva:
`-f` is not actually mutually exclusive with `--dry-run`; there's nothing wrong with doing

```
tarsnap --humanize-numbers -f foo --dry-run -c --print-stats ~/backup
```

I'm still concerned about not being able to put options in whatever order I want.  Again, writing
```
tarsnap --hu[TAB]
```

comes up with `tarsnap --nuke`, which is not at all what I was typing!  I wonder if the auto-correct script is trying to be "too smart".

vapniks:
It's because it needs to see a mode option first. I can easily change it so that the order doesn't matter, but personally I prefer it that way since it makes it easier for me to remember the modes and important options (there are a lot of options that I will probably never use).

vapniks:
Actually, just realised that the first option needs to be a mode so that it knows what other options to offer (unless you don't mind all options being available for all modes).

cperciva:
@gperciva Is there a reason this is still open?  I don't know if this was included in other changes.

gperciva:
There's kind-of a reason.  I have a branch which parses the man page and automatically adds args to the bash completion file (which prompted the recent #403 and #404).  I'm now working on extending that to cover the zsh completion file.  (Although I think I'll just have it give a warning, rather than attempting to automatically edit the file, since the zsh one also includes short doc strings.)

I do hope to resolve this within a couple of PRs.

gperciva:
Hi @vapniks,

We now have a zsh completion file for tarsnap, added in #518.  It's generated automatically from our man-page, so it should always be up-to-date.

I'm terribly sorry about the 5-year delay.  This rarely made it to the top of my todo list, and I'm not very familiar with zsh so it took a while until I was happy with a solution.

I used your descriptions of options in this PR and rewrote them into 086c670f05dcffb1a8c8772956d6456bffa19cc4.  Thank you for your contribution!  And sorry again for the delay.

PR #229, opened 2017-02-13
Crypto pre 1.1-compat

gperciva:


gperciva:
These are the "pre-crypto_compat" changes that are part of #225, in case you wanted to evaluate them separately.

lib/crypto/crypto_keys_subr.c
```diff
@@ -32,7 +32,7 @@
  */
 
 static int import_BN(BIGNUM **, const uint8_t **, size_t *);
-static int export_BN(BIGNUM *, uint8_t **, size_t *, uint32_t *);
+static int export_BN(const BIGNUM *, uint8_t **, size_t *, uint32_t *);
```
cperciva: This is problematic because old versions of OpenSSL declare BN_bn2bin as taking a `BIGNUM *` argument instead of a `const BIGNUM *` argument.  Admittedly, it looks like "old versions" in this case means pre-0.9.4, but I wouldn't be surprised if other crypto libraries copied the broken API.

How about deconsting bn in the call to BN_bn2bin?

lib/crypto/crypto_keys_subr.c
```diff
@@ -176,27 +177,37 @@ crypto_keys_subr_import_RSA_priv(RSA ** key, const uint8_t * buf, size_t buflen)
 	}
 
 	/* Load values. */
-	if (import_BN(&(*key)->n, &buf, &buflen))
+	if (import_BN(&n, &buf, &buflen))
 		goto err1;
-	if (import_BN(&(*key)->e, &buf, &buflen))
+	if (import_BN(&e, &buf, &buflen))
 		goto err1;
```
cperciva: If we `goto err1` here, what happens to `n`?

lib/crypto/crypto_keys_subr.c
gperciva: I'm confused.  OpenSSL 0.9.6 certainly uses `const BIGNUM *` for `BN_bn2bin()` [1], so what's the fuss?  I mean, if we ever switch to a different library that has a `BN_bn2bin` which takes a non-const value, couldn't we change at that time?

[1] https://github.com/openssl/openssl/blob/OpenSSL_0_9_6-stable/crypto/bn/bn.h#L344

gperciva: and yes, SSLeasy didn't use `const` here, at least as of 0.9.1b.  But is that a concern for us?

gperciva:
Updated PR with a fix to the memory, but not the de-`const`ing, pending clarification.  Also needs rebasing before a merge.

lib/crypto/crypto_keys_subr.c
cperciva: Yes, the `const` was added between 0.9.3 and 0.9.4.

The reason I was thinking that being compatible with other libraries would be good is that some (libressl, boringssl) are designed as "drop-in replacements" and users might want to drop in such a replacement themselves.  But on further consideration... well, it's *probably* safe to assume that nobody is going to resurrect a broken API from a 17-year-old version of OpenSSL at this point.  So don't worry about it.

lib/crypto/crypto_keys_subr.c
```diff
@@ -180,23 +183,23 @@ crypto_keys_subr_import_RSA_priv(RSA ** key, const uint8_t * buf, size_t buflen)
 	if (import_BN(&n, &buf, &buflen))
 		goto err1;
```
cperciva: Is there a reason why this is `err1` while all the rest are `err2`?

gperciva: It's the first one, so if it failed, no memory was allocated?  I'm happier to make it `err2` for consistency, but I figured you'd object to that.

cperciva: No memory allocated, and no memory will be freed (I assume you've checked that).  The only reason to have `err1` and `err2` is if you're going to continue with `err3`, `err4`, et cetera... the whole point of setting them to NULL is so that we can always jump to the same place.

gperciva: Double-checking BN_free(NULL)... the docs for 1.1 says it's fine [1], but the docs for 1.0.2 doesn't mention it [2].  However, the code for 1.0.2 [3] and even 0.9.6 [4] shows that it's ok.  Should I do it manually anyway, just to be safe?

[1] https://www.openssl.org/docs/manmaster/man3/BN_free.html
[2] https://www.openssl.org/docs/man1.0.2/crypto/BN_free.html
[3] https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/bn/bn_lib.c#L247
[4] https://github.com/openssl/openssl/blob/OpenSSL_0_9_6-stable/crypto/bn/bn_lib.c#L278

gperciva:
Updated PR.  If it looks good, I'll rebase the history.

Question reminder: Openssl 1.1 promises that `BN_free(NULL)` is safe.  Openssl 1.0 gives no such promise, but the code shows that it's ok.  Can we rely on it being safe, or should I check for `NULL` myself?
https://github.com/Tarsnap/tarsnap/pull/229#discussion_r101214101

cperciva:
I think this is good; please clean up the commit history.

gperciva:
Rebased and ready for merging.

PR #233, opened 2017-02-23
take archive names from stdin when the -f option is '-'

db48x:
I had to delete a number of daily backups, and my mind rebelled at the thought of writing a command to convert a nice list of archive names into the required command-line options. I just couldn't do it; it was too dumb.

Instead I modified tarsnap so that it can read archive names from stdin. I've tested both -d and --print-stats; both work great. Of course, there might well be edge cases I haven't considered.

Enjoy


cperciva:
This use case has been addressed through the addition of the --archive-names option in 7ddbb9d4, so I'm closing this PR.  Thanks for the idea!

PR #238, opened 2017-03-16
Openssl generate key

gperciva:


lib/crypto/crypto_keys_subr.c
```diff
@@ -458,11 +458,9 @@ crypto_keys_subr_generate_RSA(RSA ** priv, RSA ** pub)
 	if (*pub != NULL)
 		RSA_free(*pub);
 
-	*priv = RSA_generate_key(2048, 65537, NULL, NULL);
-	if (*priv == NULL) {
-		warn0("%s", ERR_error_string(ERR_get_error(), NULL));
+	*priv = crypto_compat_RSA_generate_key();
+	if (*priv == NULL)
```
gperciva: This follows the existing style in this function, but I wonder if it should be

```
if ((*priv = crypto_compat_RSA_generate_key()) == NULL)
```
instead.

lib/crypto/crypto_compat.c
```diff
@@ -158,7 +158,7 @@ crypto_compat_RSA_generate_key()
 {
 	RSA * key;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x00908000L
```
gperciva: Not certain when you want this to kick in.  In 1.0.2 (and even 0.9.8), `RSA_generate_key()` was inside an `# ifndef OPENSSL_NO_DEPRECATED`:
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/rsa/rsa.h#L326
https://github.com/openssl/openssl/blob/OpenSSL_0_9_8-stable/crypto/rsa/rsa.h#L276

but it only started giving compiler warnings / errors in OpenSSL 1.1.0

lib/crypto/crypto_keys_subr.c
cperciva: Yes, I'm not sure why I didn't use the normal `if ((foo = bar()) == NULL) {` idiom here.  Maybe just because I wrote this a long time ago and hadn't settled on it at that point.

lib/crypto/crypto_compat.c
cperciva: Might as well start with 0.9.8.  Is this why we've had deprecation warnings from OS X, or was that a matter of OpenSSL being completely deprecated rather than specific functions?

gperciva: I _think_ that the OS X warnings are from the (then-upcoming) complete deprecation, rather than this specific function.

gperciva:
Updated with rebased history:
- added a commit to use the modern style for `if ((a=b())==NULL`
- merged the 0.9.8 version number

PR #239, opened 2017-03-18
Fix warnings

gperciva:


lib/util/getfstype.c
```diff
@@ -62,7 +62,7 @@
 struct ftyperec {
 	unsigned int f_type;
 	const char * typename;
-} ftypes[] = {
+} static ftypes[] = {
```
cperciva: Normally we put `static` before the type... is there some reason to do otherwise here?

gperciva:
I was confused by the struct definition.  Updated with rebased PR, fixing the `static`.

lib/network/tsnetwork_select.c
```diff
@@ -46,9 +46,6 @@ network_select(int blocking)
 		return (events_run());
 	}
```
cperciva: This should have `/* UNREACHABLE */` added, just for the sake of clarity.  Sorry I didn't notice this on the first pass.

gperciva:
Updated PR, rebased onto current master, and adding `/*  UNREACHABLE */`

PR #240, opened 2017-03-28
Siginfo printinfo fixes

gperciva:


cperciva:
Huh, I guess I never noticed this because I don't use --humanize-numbers...

gperciva:
I only noticed because I was adding the casts to that file.

PR #241, opened 2017-03-30
Fix warnings 2

gperciva:


gperciva:
This is a different set of fixes than #239 "Fix warnings".

PR #247, opened 2017-04-22
News updates

gperciva:


gperciva:
I'm not completely certain that you'll want to mention `./configure --with-conf-no-sample`, but since it will overwrite an existing `tarsnap.conf`, I think it's worth it for the warning.

Issue #249, opened 2017-04-25
'INT_MAX' undeclared


jungle-boogie:
Hi All,

Attempting to build `tarsnap` from the git repo on FreeBSD 11-release.

```
% gmake
gmake  all-am
gmake[1]: Entering directory '/usr/home/jungle/bin/tarsnap'
gcc -std=gnu99 -DHAVE_CONFIG_H -I.  -I./tar -I./libarchive -I./lib/crypto -I./lib/datastruct -I./lib/keyfile -I./lib/netpacket -I./lib/netproto -I./lib/network -I./lib/scryptenc -I./lib/util -I./libcperciva/alg -I./libcperciva/cpusupport -I./libcperciva/crypto -I./libcperciva/datastruct -I./libcperciva/events -I./libcperciva/network -I./libcperciva/util -DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\" -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -DPOSIXFAIL_MSG_NOSIGNAL -DPOSIXFAIL_CLOCK_REALTIME   -g -O2 -Wall -MT lib/crypto/lib_libtarsnap_a-crypto_keys_subr.o -MD -MP -MF lib/crypto/.deps/lib_libtarsnap_a-crypto_keys_subr.Tpo -c -o lib/crypto/lib_libtarsnap_a-crypto_keys_subr.o `test -f 'lib/crypto/crypto_keys_subr.c' || echo './'`lib/crypto/crypto_keys_subr.c
lib/crypto/crypto_keys_subr.c: In function 'import_BN':
lib/crypto/crypto_keys_subr.c:60:12: error: 'INT_MAX' undeclared (first use in this function)
  if (len > INT_MAX) {
            ^
lib/crypto/crypto_keys_subr.c:60:12: note: each undeclared identifier is reported only once for each function it appears in
gmake[1]: *** [Makefile:2288: lib/crypto/lib_libtarsnap_a-crypto_keys_subr.o] Error 1
gmake[1]: Leaving directory '/usr/home/jungle/bin/tarsnap'
gmake: *** [Makefile:1153: all] Error 2
```

I see this commit from a little more than a month ago:
https://github.com/Tarsnap/tarsnap/commit/95132525ab533cd6d52ddcd83ccad2d397a2c751

thanks!

gperciva:
[...]

PR #250, opened 2017-04-25
Fix strict header inclusion for INT_MAX

gperciva:


gperciva:
Reported by https://github.com/Tarsnap/tarsnap/issues/249

PR #251, opened 2017-04-30
Fix strict header inclusion for INT_MAX (take 2)

gperciva:


gperciva:
This fixes compiling `lib/crypto/crypto_rsa.c` on OpenBSD using libressl.

I can't complete the linking stage (with libressl) because there's nastiness relating to OPENSSL_VERSION_NUMBER.  For a quick peak, take a look at:
https://wiki.freebsd.org/LibreSSL/Ports#OpenSSL_1.1.0
I'm still investigating libressl and how other projects manage this.  But this PR is still good; we shouldn't rely on other headers to provide necessary definitions.

cperciva:
Should this PR be closed now?  It conflicts and you said changes are in #252 anyway?

gperciva:
Whoops.  It was also part of #253 Explicit includes, which was merged.  But that one was missing the bug bounty in the commit message.  Sean already received the bounty, though -- so I suggest we just close this PR and not worry about it.

PR #252, opened 2017-04-30
Libressl

gperciva:


gperciva:
This includes PR #251 (explicitly including `<limits.h>`), and enables compiling with libressl, using what appears to be the "standard" way of handling OpenSSL 1.0, OpenSSL 1.1, and LibreSSL.

We may want to add some mention of LibreSSL somewhere -- either to say that we support it, or that tarsnap can compile with it but it's not supported, or something.  Note that OpenBSD switched to LibreSSL for release 5.6 on 2014-11-01, so we've probably already had people using tarsnap with libressl without realizing it.

lib/crypto/crypto_compat.c
```diff
@@ -23,7 +23,7 @@ crypto_compat_RSA_valid_size(const RSA * const rsa)
 	/* Sanity checks. */
 	assert(rsa != NULL);
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER)
```
cperciva: How about putting

```
/* LibreSSL claims to be OpenSSL 2.0, but (currently) has APIs compatible with OpenSSL 1.0.0. */
#ifdef LIBRESSL_VERSION_NUMBER
#undef OPENSSL_VERSION_NUMBER
#define OPENSSL_VERSION_NUMBER 0x100000000L
#endif
```

at the top of the file?  Simpler than checking for LibreSSL in four different places...

gperciva:
Redefining a FOO_VERSION_NUMBER seems wack to me, but a few other projects have done this so I guess I don't have much room to complain.  PR updated accordingly, and it works on openbsd with libressl.

lib/crypto/crypto_compat.c
```diff
@@ -13,6 +13,15 @@
 #error "OPENSSL_VERSION_NUMBER must be defined"
 #endif
 
+/*
+ * LibreSSL claims to be OpenSSL 2.0, but (currently) has APIs compatible with
+ * OpenSSL 1.0.7.
```
cperciva: 0x1000107fL means 1.0.1g, not 1.0.7 (which doesn't exist).

cperciva:
I agree that redefining a FOO_VERSION_NUMBER is a bit odd; a "cleaner" approach would be to define our own SSL_API_VERSION and set that based on OPENSSL_VERSION_NUMBER and LIBRESSL_VERSION_NUMBER.  If we have to cope with versioning issues in other libraries we might want to do that in the future; but for now redefining OPENSSL_VERSION_NUMBER seems more concise and more easily understood than defining an entirely new SSL_API_VERSION macro.

gperciva:
Updated PR, changing the comment to refer to 1.0.1g.

PR #254, opened 2017-05-07
Fix warnings ssize buffer

gperciva:


gperciva:
I have an extra check that
```
len = le32dec(d->c.ch.len);
if (len <= SSIZE_MAX)
    bail;
```
in `multitape_read.c`, but in `multitape_write.c` I just assumed that `ch->len` would be <= MAXCHUNK.  I'm a bit iffy about this discrepancy, but it would have been awkward to add the checks to `multitape_write.c` based on how the existing code was structured.

PR #256, opened 2017-05-09
Headers

gperciva:


lib/crypto/crypto_keys.c
```diff
@@ -7,6 +7,7 @@
 
 #include <openssl/crypto.h>
 #include <openssl/err.h>
+#include <openssl/ossl_typ.h>
```
cperciva: This seems like an internal OpenSSL header.  The RSA_* functions are documented only as needing <openssl/rsa.h>, and in fact that's guaranteed to provide the type definition we need since it provides functions which declare the type.

cperciva:
Get rid of the <openssl/ossl_typ.h> includes and this looks good.

gperciva:
Ok, updated PR after removing `ossl_typ.h` from the first commit, and rebasing on master.

PR #259, opened 2017-05-14
Verbose deletion of a single archive

gperciva:


gperciva:
This an offshoot of #233 which I think can/should be discussed separately.  Based on the principle of least surprise, I agree that

    ./tarsnap -v -d -f foo

should print `Deleting archive "foo"`.  Currently, this only happens if there are more than one `-f` archive specified.  This becomes more important when we have other ways of supplying archive names (e.g., when debugging a script which is supposed to rotate archives), but even now I think it's better to avoid any surprises when it comes to deleting archives.

NEWS.md
```diff
@@ -15,6 +15,8 @@
   passphrase-encrypted key file to proceed even if it is anticipated to require
   an excessive amount of memory or CPU time.
 - tarsnap now supports OpenSSL 1.1.
+- tarsnap now displays 'Deleting archive "foo"' when deleting a single archive
+  in verbose mode.
```
cperciva: I'd add "(The former behaviour was to print that message only when multiple archives were being deleted by a single tarsnap command.)" for clarity.  Since we're making a user-visible change to behaviour, clarity is important.

gperciva:
Updated PR, rebasing onto master and adding the second sentence to `NEWS.md`.

PR #261, opened 2017-05-18
Archive names

gperciva:


gperciva:
First draft of an alternative way of doing #233.

gperciva:
I'm not wild about the name of `bsdtar->tapenames_setup`, but after staring at the screen for a minute, I figured I should just pick *something* and get back to useful coding.  :)

I can also spin off the elasticarray change into a separate PR if it's easier to review / track as an independent change.

cperciva:
I haven't looked at the code yet, but based on commit messages alone: Can you think of any scenario where 'archive-names foo' makes sense in a configuration file?  I can't.  If not, there's no reason to have --no-archive-names as an option.

gperciva:
> Can you think of any scenario where 'archive-names foo' makes sense in a configuration file?

Sure!  Well, kind-of.  Depends on how much of a "if you have a hammer" somebody is when it comes to pipes.  :)

Assuming `archive-names -` in the config,
```
echo "bak-`date -I`" | tarsnap -c ~
```

or if somebody has a delete-only key on a machine (or separate user account), they could set up `archive-names rotate-delete.txt` then run
```
tarsnap -d
```

Or just possibly some kind of monitoring script -- a read-only key, which print stats on a daily basis just to check that the archive sizes hasn't changed (and implicitly, still exist)?  Granted, that's a rather weaksauce attempt at checking that the archives are valid, but it's something that ~~a reasonable person~~ I might do.

Granted, I can't think of any scenario which can't trivially be rewritten to use `--archive-names foo.txt` directly.  So ok, I'll remove `--no-archive-names`.

gperciva:
Revised PR

tar/bsdtar.c
```diff
@@ -416,7 +417,9 @@ main(int argc, char **argv)
 			optq_push(bsdtar, "exclude", bsdtar->optarg);
 			break;
 		case 'f': /* multitar */
-			bsdtar->tapenames[bsdtar->ntapes++] = bsdtar->optarg;
+			tapename_cmdline = (char *)bsdtar->optarg;
```
cperciva: How does this even compile?  `bsdtar->optarg` is a `const char *`... you can't just cast away `const` like that!

tar/bsdtar.h
```diff
@@ -35,6 +35,10 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#include "elasticarray.h"
+
+ELASTICARRAY_DECL(TAPENAMES_SETUP, tapenames_setup, char *);
```
cperciva: I would call this STRLIST.  More generally, elastic arrays should have names appropriate to what they contain, not what they're used for.

tar/bsdtar.c
```diff
@@ -828,6 +837,12 @@ main(int argc, char **argv)
 		configfile(bsdtar, ETC_TARSNAP_CONF, 0);
 	}
 
+	/* Extract tapenames from tapenames_setup. */
+	if (tapenames_setup_export(bsdtar->tapenames_setup, &bsdtar->tapenames,
+	    &bsdtar->ntapes))
+		bsdtar_errc(bsdtar, 1, 0, "foo");
```
cperciva: foo?

tar/bsdtar.c
```diff
@@ -1399,6 +1414,48 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
 }
 
+static void
+archive_names(struct bsdtar *bsdtar, const char *fname)
+{
+	struct stat sb;
+
+	/* If it's a file, does it exist? */
+	if (strcmp(fname, "-") != 0) {
+		if (stat(fname, &sb))
```
cperciva: Why?  The only test which matters is whether `process_lines` can read the file, and it handles errors.

tar/bsdtar.c
```diff
@@ -1399,6 +1414,48 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
+	process_lines(bsdtar, fname, archive_names_helper, 0);
+}
+
+/* Process a line of configuration file. */
+static int
+archive_names_helper(struct bsdtar *bsdtar, const char *line)
+{
+	char * name;
+
+	/* Skip any leading whitespace. */
```
cperciva: Why ignore leading whitespace but not trailing whitespace?

tar/bsdtar.c
```diff
@@ -1399,6 +1414,48 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
+	while ((line[0] == ' ') || (line[0] == '\t'))
+		line++;
+
+	/* Ignore comments and blank lines. */
+	if ((line[0] == '#') || (line[0] == '\0'))
```
cperciva: ... for that matter, someone might have an archive named `#foo`.  I'd be inclined to simply take the archive names as provided.

NEWS.md
```diff
@@ -15,6 +15,8 @@
   passphrase-encrypted key file to proceed even if it is anticipated to require
   an excessive amount of memory or CPU time.
 - tarsnap now supports OpenSSL 1.1.
+- tarsnap now accepts --archive-names filename, which load a list of
```
cperciva: s/load/loads/ ? Although I think "reads" might be clearer anyway.

gperciva:
Updated PR with fixes with messy history.  Will rebase if/when you want.

tar/bsdtar.c
```diff
@@ -80,6 +80,7 @@ __FBSDID("$FreeBSD: src/usr.bin/tar/bsdtar.c,v 1.93 2008/11/08 04:43:24 kientzle
 #include "bsdtar.h"
 #include "crypto.h"
 #include "dirutil.h"
+#include "elasticarray.h"
```
cperciva: Is this necessary?  I don't think we use any elasticarray functions directly in here?

tar/bsdtar.h
```diff
@@ -35,6 +35,10 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#include "elasticarray.h"
+
+ELASTICARRAY_DECL(STRLIST, tapenames_setup, char *);
```
cperciva: I was sort of thinking that the functions would be prefixed `strlist_` as well...

tar/bsdtar.c
```diff
@@ -187,9 +193,23 @@ static void
 bsdtar_atexit(void)
 {
 	struct bsdtar *bsdtar;
+	size_t i;
 
 	bsdtar = &bsdtar_storage;
 
+	/* Free temporary archive names (if an error occurred before export). */
+	if (bsdtar->tapenames_setup != NULL) {
+		for (i = 0; i < bsdtar->ntapes; i++)
```
cperciva: why not replace `bsdtar->ntapes` with `strlist_getsize(bsdtar->tapenames_setup)` here?  That way you won't need to adjust `bsdtar->ntapes` every time you add a string to the array.

tar/bsdtar.c
gperciva: `clang -Weverything` is quite happy with it...?  Anyway, I split the de-consting into #267 and tweaked the intermediate commits.

gperciva:
Updated PR, including #269 .

tar/bsdtar.c
```diff
@@ -1397,6 +1411,29 @@ configfile_helper(struct bsdtar *bsdtar, const char *line)
...
+		return (0);
+
+	/* Duplicate line. */
+	if ((name = strdup(line)) == NULL)
+		bsdtar_errc(bsdtar, 1, errno, "Out of memory");
+
+	/* Record archive name. */
+	if (strlist_append(bsdtar->tapenames_setup, &name, 1))
+		bsdtar_errc(bsdtar, 1, errno, "Out of memory");
+	bsdtar->ntapes++;
```
cperciva: Not needed.

cperciva:
Looks good.  Yes, please rebase.

gperciva:
Updated PR, rebased onto master with fixed history.

PR #266, opened 2017-06-27
Fix warnings final

gperciva:


gperciva:
This is the last PR required to compile with `clang -Weverything <-Wno- a few specific exceptions>` on linux amd64 and i386 (debian wheezy).  It's possible that I'll discover more warnings in a more recent linux i386, and of course `clang` and `gcc` are always adding more warnings... but I won't go looking for those until the after the next stable release.

You might find d91b9ed a bit iffy; there's a few ways to address that issue, so I tried to guess what you'd prefer.

tar/chunks/chunks_stats_internal.c
```diff
@@ -35,9 +35,9 @@ chunks_stats_add(struct chunkstats * stats, size_t len, size_t zlen,
     ssize_t copies)
 {
 
-	stats->nchunks += copies;
-	stats->s_len += (uint64_t)(len) * (int64_t)(copies);
-	stats->s_zlen += (uint64_t)(zlen) * (int64_t)(copies);
+	stats->nchunks += (uint64_t)copies;
```
cperciva: I think this needs an explanatory comment:
```
/*
 * The value ${copies} may be negative, but since nchunks, s_len, and s_zlen
 * are all of type uint64_t, casting a negative value to uint64_t will still yield the
 * correct results, thanks to modulo-2^64 arithmetic.
 */
```

tar/ccache/ccache_read.c
```diff
@@ -65,7 +68,17 @@ read_rec(void * cookie)
 	ccr->ino = (ino_t)le64dec(ccre.ino);
 	ccr->size = (off_t)le64dec(ccre.size);
 	ccr->mtime = (time_t)le64dec(ccre.mtime);
+#if SIZE_MAX < UINT64_MAX
+	ccr_nch = le64dec(ccre.nch);
```
cperciva: Rather than checking this here, I'd just adjust the check under `/* Sanity check some fields. */` to use the original 64-bit value instead of the `size_t` version.

gperciva: Does that indicate a corrupt cache file?  In `ccache_entry.c` `callback_addchunk()`, I don't see any 32-bit-related maximum size for `->nch`.  So as far as I can tell, a cache file could be perfectly valid on a 64-bit platform, but not valid on a 32-bit platform.

If I adjust the existing sanity check, people might get a "Cache file is corrupt" message when actually it would still be valid on a 64-bit platform.  I thought it would be nice to have a separate error message for this case.

gperciva:
Updated PR, rebasing onto master, adding the comment about `(uint64_t)`, but using a slightly different approach to `ccre.nch` 64-bit vs. 32-bit.  I'll obviously wrap it up into the existing sanity check if you're certain that's what you want.

tar/ccache/ccache_read.c
cperciva: Technically using a large cache from a 64-bit system on a 32-bit system could trigger the "Cache file is corrupt" message, yes... but that's far less likely than the cache actually getting corrupted, so I'd prefer to have a false-positive warning in a very unusual use case than fail to warn about something which is probably a problem.

... especially since the solution in both cases is the same.

tar/ccache/ccache_read.c
```diff
@@ -79,6 +79,11 @@ read_rec(void * cookie)
 	ccr->flags = 0;
 
 	/* Sanity check some fields. */
+	if (le64dec(ccre.nch) > (uint64_t)SIZE_MAX) {
+		warn0("Cache record has too many chunks for this platform;"
```
cperciva: Let's say "Cache file is corrupt or too large for this platform: %s".

gperciva: err, what's the `%s` for?  `read_rec()` doesn't have the filename.  Or is that a placeholder for the "Please run `--fsck`" message?

gperciva: never mind, I see it now.  My eyes focused on `        FILE * f;       /* File handle. */` and totally skipped over the `        char * s;       /* File name. */` right above it.  :(

gperciva:
Updated PR

PR #267, opened 2017-06-28
Tapenames deconst

gperciva:


gperciva:
Split the first two commits off from #261.

PR #269, opened 2017-07-02
Tapenames elasticarray

gperciva:


gperciva:
This another one split from #261.  Not certain if I should rebase the history or not, so I left it undone.

tar/bsdtar.c
```diff
@@ -783,14 +790,15 @@ main(int argc, char **argv)
 	 * few bytes off anyway since the command line, including "--dry-run"
 	 * is included in the metadata.
 	 */
-	if (bsdtar->option_dryrun && (bsdtar->ntapes == 0)) {
+	if (bsdtar->option_dryrun &&
+	    (strlist_getsize(bsdtar->tapenames_setup) == 0)) {
```
gperciva: This is a bit ugly, but if we're not incrementing `bsdtar->ntapes`, I guess we don't have an alternative.

cperciva:
The final result looks good.  I think this can probably just be a single commit...

gperciva:
Updated PR with rebased history.  Yes, I intended it to be a single commit.

Issue #277, opened 2017-07-15
tarsnap 1.0.38 crashes with an assertion failure


jkseppan:
```
tarsnap: tar/ccache/ccache_write.c:75: callback_write_rec: Assertion `(ccr->mtime >= 0) && ((uintmax_t)ccr->mtime <= (18446744073709551615UL))' failed.
Aborted (core dumped)
```

This is on Ubuntu 14.04.5 LTS, Tarsnap 1.0.38 compiled from sources. Here is some of the stack trace from the core file:

```
#0  0x00007f00a7bcbc37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f00a7bcf028 in __GI_abort () at abort.c:89
#2  0x00007f00a7bc4bf6 in __assert_fail_base (
    fmt=0x7f00a7d19018 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x44a500 "(ccr->mtime >= 0) && ((uintmax_t)ccr->mtime <= (18446744073709551615UL))", file=file@entry=0x44a42a "tar/ccache/ccache_write.c", 
    line=line@entry=75, 
    function=function@entry=0x44a5b0 <__PRETTY_FUNCTION__.6609> "callback_write_rec")
    at assert.c:92
#3  0x00007f00a7bc4ca2 in __GI___assert_fail (
    assertion=assertion@entry=0x44a500 "(ccr->mtime >= 0) && ((uintmax_t)ccr->mtime <= (18446744073709551615UL))", file=file@entry=0x44a42a "tar/ccache/ccache_write.c", 
    line=line@entry=75, 
    function=function@entry=0x44a5b0 <__PRETTY_FUNCTION__.6609> "callback_write_rec")
    at assert.c:101
#4  0x000000000040fe0b in callback_write_rec (cookie=<optimized out>, 
    s=<optimized out>, slen=40, rec=<optimized out>) at tar/ccache/ccache_write.c:75
#5  0x000000000043615f in foreach_internal (n=<optimized out>, 
    func=func@entry=0x40fa20 <callback_write_rec>, 
    cookie=cookie@entry=0x7ffea7724e30, 
[...]

PR #278, opened 2017-07-15
Handle negative mtime

gperciva:


gperciva:
Not certain this is the best fix, but it's certainly the quickest one to test.

Attempts to address #277.

tar/ccache/ccache_write.c
```diff
@@ -83,6 +87,10 @@ callback_write_rec(void * cookie, uint8_t * s, size_t slen, void * rec)
 	if (ccr->age > MAXAGE)
 		goto done;
 
+	/* Don't write an entry if it has negative mtime. */
```
cperciva: We should check this before the assertions, since if `sizeof(uintmax_t) > sizeof(uint64_t)` the assertion will still fail for an mtime of -1.

cperciva: Actually, we might as well have all of the "Don't write an entry if" checks before the assertions.

gperciva:
Updated PR

PR #279, opened 2017-07-16
Refactor ccache_write: Add skiprecord predicate

cperciva:
In callback_count, callback_write_rec, and callback_write_data we're
making identical checks to see if the cache entry is one we want to
write out to disk.  Merge these into a 'skiprecord' function which
gets used in all three places.


gperciva:
Looks good to me.

PR #282, opened 2017-07-30
Iwyu

gperciva:


cperciva:
I just found this really old PR which I had somehow not noticed...

Can you remove the `asprintf` bits from `tar/bsdtar_platform.h` (and make sure nothing breaks in the process)?  That's why I didn't have `#include "asprintf.h"` scattered everywhere -- I was too lazy to find everywhere it was needed so I just included it via `bsdtar_platform.h`.

gperciva:
First glance looks good without `aspritnf`... but then again, I just noticed that my platform includes it in `<stdio.h>` and Tarsnap uses `_GNU_SOURCE`.  I'll need to do more testing.

Is this something you want in the current PR?  I think there's about another half-dozen changes related to headers that iwyu was suggesting; my policy was to pick out "obviously good" stuff so that the PRs could be dealt with quickly.  :)

cperciva:
I guess I was just a bit wary about merging something which adds `#include "asprintf.h"` when that's already potentially being included.  Maybe we can merge the rest and then tidy up the asprintf situation later?

gperciva:
err, isn't that exactly why we have
```
#ifndef _ASPRINTF_H_
#define _ASPRINTF_H_
```

?  but sure, I'll put the asprintf stuff in a separate PR.

gperciva:
Rebased to remove the asprintf-adding commit.

gperciva:
Rebased onto master so we have the "travis clang-3.9" fix.

PR #284, opened 2017-08-02
Add the packaging metadata to build the tarsnap snap

elopio:
This package will let you publish the latest tarsnap in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,35 @@
+name: tarsnap
+version: master
+summary: Tarsnap is a secure, efficient online backup service.
+description: "Online backups for the truly paranoid"
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: devmode
+
+apps:
+  tarsnap:
+    command: tarsnap
+  keygen:
+    command: tarsnap-keygen
+    aliases: [tarsnap-keygen]
+  keymgmt:
+    command: tarsnap-keymgmt
+    aliases: [tarsnap-keymgmt]
+  keyregen:
+    command: tarsnap-keyregen
+    aliases: [tarsnap-keyregen]
+  recrypt:
+    command: tarsnap-recrypt
+    aliases: [tarsnap-recrypt]
+
+parts:
+  tarsnap:
+    source: .
+    plugin: autotools
+    build-packages:
+      - gcc
+      - libc6-dev
+      - make
+      - libssl-dev
+      - zlib1g-dev
+      - e2fslibs-dev


elopio:
Hello!

Earlier today I was using tarsnap as a test subject for the new cross-compile feature of autotools projects in snapcraft. I thought it might be useful for you to publish the project in the Ubuntu store, so I made this PR.

snapcraft is the new packaging system that we are writing in Ubuntu. You can read more about it here: https://snapcraft.io

To build this snap locally in Ubuntu 16..04:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/tarsnap
    $ cd tarsnap
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous --devmode

As you can see, the snap is in devmode. It still requires some work to work fully confined and secure, but it's a start.

If you have any questions, please let me know.

pura vida.

shinnok:
@elopio Leo, I was planning on doing a snapcraft for Tarsnap GUI but I thought it would be best to have one for the CLI first. Any idea how easy would it be to:
1. Package Tarsnap GUI with a single dep. on Qt SDK as a snap?
2. Package both GUI and CLI in a snap? Is this a feasible/recommendable approach?

gperciva:
Thanks for your interest!  `snap` is something that we've considered, but there's a number of issues that we'd like to be resolved first.  If you can shed some light on these, great!  Otherwise these are notes to myself.

In no particular order,

- can we put the directory inside `pkg/snap/`?  We don't want to complicated the top-level directory.

- how does package signing work?  I poked around for a few minutes on https://snapcraft.io/docs/ but couldn't find mention of using a GPG key.

- our Debian package uses
    ```
    dh_auto_configure -- --with-bash-completion-dir --with-conf-no-sample
    ```
    so it would be nice if snap could deliver the same behaviour.  (I expect that we'd need to add `bash-completion` to the dependencies)

- note to self (since I didn't see this in the docs): you can investigate the package contents with:
    ```
    sudo mount tarsnap_master_amd64.snap /mnt
    ```
    For some reason it contains an empty `snap/` directory, and an empty `meta/gui/` directory.  The rest looks like what I'd expect, although the actual binaries differ from what was compiled manually on the system.  (definitely something to investigate further!)

elopio:
@shinnok, we have a plugin for qmake. You can run `snapcraft help qmake`. We also have a part to help with qt5 desktop apps, take a look at  `snapcraft define desktop-qt5`. With that, I think the tar GUI should be simple to snap, but I haven't tried.

The snap is yours, if you want to have the CLI and the GUI on the same one, that's OK. You can put the snapcraft.yaml in one of the repos, or make a separate repo for packaging. And instead of `source: .`, you can use in there the URL of the github repo.

@gperciva, you can put the snap in a different directory. But if you want to use https://build.snapcraft.io, it has to be in the root. If you prefer a different location, you can use your own CI, travis, or launchpad for continuous deliver: https://snapcraft.io/docs/build-snaps/ci-integration

The snap is signed by the store, and the signature is verified upon install. Our security team has been talking about allowing more signature options, but so far this hasn't been a top priority for our users. If you comment about this in the forum, it will help to raise the priority in our backlog: https://forum.snapcraft.io

Tab completion is a feature in progress. Some things are already implemented and releasing, but we are still polishing it. You can find some discussions in the forum, like https://forum.snapcraft.io/t/tab-completion-out-of-the-box/1395

About inspecting the contents of the snap, if you are using snapcraft, it will leave a `prime` directory. That is the directory that will be compressed to generate the .snap. Your mount solution is good, and also, after install, the snap will be uncompressed in /snap/tarsnap/current.

The empty dirs in the `snap` dir is clearly a bug. I'll fix it.

Thanks for your reviews. Let me know if you have any more doubts. At this point, I would suggest you to just give it a try. Once the snapcraft.yaml is landed, you go to build.snapcraft.io and configure there the automated continuous delivery. That will leave the snap in the edge channel of the store, where it won't be visible on user searches. You can test it there, take a look at the store, tell some trusted users to give it a try and see how they like it. Then you can make any changes you want before publishing in the `stable` channel, where it will be visible even in the software center. Or, if you see no value from the snap, you can always drop it instead of moving it to stable.

pura vida.

gperciva:
Hi @elopio,

Thanks for the updates!  Just to clarify one point: using a build system that is not under our control is not possible with Tarsnap.  Our motto is "online backups for the truly paranoid", after all!

When we build our `deb` packages, we transfer source files written directly to a USB stick with `tar` (i.e. no filesystem on the stick) to an air-gapped system which has never been used for anything else.  We're open to building `.snap` packages ourselves, signing them with our GPG key, and uploading them to an appropriate location -- but only once we fully trust the system.

I suspect that Tarsnap is not an ideal candidate for snap's design.

elopio:
@gperciva well, it's for you to choose, of course. Maybe, you will decide to keep with traditional deb packaging, and I'm totally fine with that.

I'm biased, obviously :) That's why I suggested you to give it a try, and experiment on your own. But I will give my thoughts anyway. 

First, you can build the .snap in any way you want. You can even use the same binaries already generated that you put in the deb, and just use the dump snapcraft plugin to put them in the snap. You can use snapcraft on that isolated build machine, or you can make the .snap in any other way you want, it is just a squashfs. The idea is that you can follow your current release process, instead of having to adapt to a release cycle of 6 months and complex rules imposed by the intermediaries between you and your users, that won't always help your productivity.

Then, the biggest downside of .debs, in my opinion, is that they execute maintainer scripts with root permissions. If the deb is in the archive, we as users give a lot of trust to all the debian and ubuntu developers, which is reasonable. If the deb is in a PPA, then I think installing a deb is way too risky. Effectively, I would be giving root access to my machine to the person who registered that PPA. snaps instead are just mounted read-only. They don't execute anything during installation, and I always have full control of the plugs that I decide to allow.

Another important point is that https://build.snapcraft.io uses the launchpad builders. These are the same machines that we use to build the ubuntu archive. It's one of the safest systems I know, and I fully trust it. It builds my kernel, and my file system. It doesn't matter how careful I am with the other parts of my system if these critical packages are not handled with care. This part is handy and secure, but as I mentioned, it's optional.

I invite you to put all your concerns in https://forum.snapcraft.io. Our security team hangs in there, so you will see all the things they have already implemented to make sure that this store is at the same level of the deb archive, all the things they added after years of learning about security problems in traditional packaging, and all the things that they have in progress.

pura vida.

cperciva:
@elopio If I understand this correctly, 'snap' packages bundle dependencies; is that correct?  How does the update process work?  In particular, tarsnap uses OpenSSL; if there's an OpenSSL security vulnerability, what's the process which results in people running tarsnap with the updated OpenSSL code?

elopio:
@cperciva you have to take care of your own dependencies. If there is an openssl vulnerability, you have to update your snap, push it to the store, and all your users will get an automatic update.

We are working on a feature that will record a build manifest of the snaps. So when there is a cve, we can know which snaps are affected and notify the developers. This is being discussed in our forum: https://forum.snapcraft.io/search?q=manifest

cperciva:
@elopio Just to be clear, the case I'm talking about here is where you're doing the building, not us.  There's no way that we're going to be re-running our own builds every time there's an OpenSSL update.  As @gperciva says, we can't exactly endorse binaries built by third parties; but like FreeBSD users who make use of FreeBSD's tarsnap packages, if people are already using snapcraft there's not necessarily much increased exposure if they use snapcraft tarsnap packages as well.

This PR doesn't specify versions of dependencies; so if https://build.snapcraft.io/ is building snaps, there must be some out-of-band mechanism for having it get an updated OpenSSL when needed...?

elopio:
This pr declares `libssl-dev` as build-package. Because it doesn't specify a version, it will get the latest from the ubuntu archive. The ubuntu archive will patch any vulnerabilities in libsss-dev as soon as possible, so you will have to retrigger a rebuild in build.snapcraft.io.

We could, with this new feature that we have in progress, notice that you built with an ssl version with a vulnerability, and retrigger the snap build as soon as libssl-dev is updated in the archive. However, we have no way to know if your project will work with this updated library. For that, we are also working on a feature called health-check, that will run some tests on the snap installed, defined by upstream.

So this new world that we are proposing is actually more secure. We have found that it doesn't scale to have a centralized repository. If we want to update a critical vulnerability, we have to update every single package in the archive that uses it, and test every single package and the chain of dependencies.

Now we are leaving the responsibility to upstreams. We update the core system, make sure that the basic applications work. And we could update all the snaps with healthchecks. And if there's a healthcheck that fails, we can notify the upstream developer to take a look, maybe even put the snap in quarantine. But we don't have to wait for all the applications to be ready, in order to apply an update.

Many of these features are in discussion in the forum. We have the chance to write a better, safer and sustainable system, and your input will be very appreciated there.

gperciva:
As long as an updated package only changes its minor version number (or patch level), it should be source-compatible.  If you can't trust that libraries will do this and automatically update packages accordingly, then you're leaving a huge security risk for your users.  This sounds like an attempt to reinvent the wheel without being aware of all the problems that wheels are supposed to solve.

(and before anybody says "oh, this is only theoretical"... security flaws in openssl are not at all rare!)

cperciva:
@elopio I don't want to get a notification.  I don't want to manually trigger a rebuild.  I don't want to run health checks or test new packages.

When there's a new `libssl-dev`, I want tarsnap users to get it.  Immediately.  If it makes tarsnap stop working... well, better that tarsnap stops working than people keep on running known vulnerable code.

Does that make sense?

elopio:
We have been maintaining patches for security vulnerabilities for 10 years. Some of my teammates even for more time because they are debian maintainers before ubuntu devs. We are even backporting security fixes to precise, a system released 5 years ago. We understand the problem, and all the benefits that a centralized archive gives us. But we also know all the limitations, for example, that there is no secure way to let the millions of users in trusty install tarsnap. That's the problem we are trying to solve. Not depending on a centralized archive for every library brings duplication, and requires a little more work from upstreams, but opens so many doors. We know it's not just a theoretical or infrequent possibility, we have to support many huge companies relying on ubuntu and requiring prompt updates, so we are dealing with CVEs each day. The impossibility to find a balance between a secure, stable and updated system with the current wheels is actually what led us to design snaps.

In the case of openssl, and some other very critical components, we have even discussed about keeping them inside core. But there's only so much we can do before we get back to the slow 6 month cycle, complex packaging and strong freeze that centralization requires.

As I said, this actually opens many options that we didn't have before, like quarantine an application that will fail after applying a security update. This is not yet implemented, but finally there's nothing stopping us from doing it, we can now add new applications to the store without having to sync them with thousands of other existing packages. We are very happy when upstream developers that have been excluded from the archive tell us about their needs, and their priorities, because that's what we use to sort our backlog. So let me insist that it would be awesome if you push your snap to edge, give it a try, and bring your feedback to the forum.

pura vida.

cperciva:
Closing this since it doesn't seem that snap supports automatically updating dependencies which are rolled into the package, which I consider to be essential for security.

Feel free to re-open once you have the necessary tooling in place.

elopio:
@cperciva oh, that's totally doable now.

Look here:

https://github.com/elopio/ipfs-snap/blob/759fdd1e4777af4c7a28d5f4cff223ecb65bca34/snapcraft.yaml#L22

If there is a security vulnerability in make or wget, it will be patched in the distro. This snap is built everyday and pushed to the store by [travis](https://snapcraft.io/docs/build-snaps/ci-integration#using-travis). So it will be ready for testing less than 24 hours after the patch is landed, and your users will get it automatically in less than 4 hours after you mark it good. 

That's a huge improvement over the current situation. Our stats show that most of our users don't run apt upgrade, so they have outdated software always. And we have in progress reducing those times for very critical bugs, with push updates that will trigger a refresh on user machines almost immediately.

But, I won't insist anymore. It's ok if you don't want this distribution mechanism. What we need at this point are projects interested in guiding and contributing with our new features. Thanks a lot for your comments, that we are already taking into account.

pura vida.

Issue #286, opened 2017-08-21
1.0.39 clock_gettime problem on OS X 10.11 with Xcode 8.x


ilovezfs:
```
==> brew test tarsnap --verbose
==> FAILED
Testing tarsnap
/usr/bin/sandbox-exec -f /tmp/homebrew20170821-9565-xa46nn.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/tarsnap.rb --verbose
==> /usr/local/Cellar/tarsnap/1.0.39/bin/tarsnap -c --dry-run /tmp/tarsnap-test-20170821-9566-xa8gv4
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/tarsnap/1.0.39/bin/tarsnap
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/tarsnap/1.0.39/bin/tarsnap
  Expected in: /usr/lib/libSystem.B.dylib
```

The test block runs
```
  test do
    system bin/"tarsnap", "-c", "--dry-run", testpath
  end
```

https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/6974/version=el_capitan/console
Duplicated here since that log will eventually be deleted: https://gist.github.com/ilovezfs/8ee666cbb4a5ec36e615771f423db925

I can work around this issue with the following silly hack:
```
    if MacOS.version == :el_capitan && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
      inreplace "libcperciva/util/monoclock.c", "CLOCK_MONOTONIC",
                                                "UNDEFINED_GIBBERISH"
    end
```

The immediate problem here is that even though there are configure time checks for the availability of clock_gettime, the code in `libcperciva/util/monoclock.c` is guarded with `CLOCK_MONOTONIC` rather than `HAVE_CLOCK_GETTIME`:
```
#ifdef CLOCK_MONOTONIC
        if (clock_gettime(CLOCK_MONOTONIC, &tp) == 0) {
                tv->tv_sec = tp.tv_sec;
[...]

PR #287, opened 2017-08-23
Monoclock gettime

gperciva:


gperciva:
This is a draft fix for https://github.com/Tarsnap/tarsnap/issues/286, mainly so that the libcperciva change can be accepted and I can link to that commit hash for the "import from libcperciva" commit.

I'm looking at better-integrating the libcperciva POSIX approach (i.e. defining `-DPOSIXFAIL_*` as appropriate) with autotools, but this PR is a good first start.

configure.ac
```diff
@@ -498,5 +498,9 @@ fi
 AC_SUBST([BASH_COMPLETION_DIR])
 AM_CONDITIONAL([INSTALL_BASH_COMPLETION],[test "x$with_bash_completion_dir" != "xno"])
 
+if test "x$ac_cv_func_clock_gettime" != "xyes" ||
+    test "x$ac_cv_lib_rt_clock_gettime" != "xyes"; then
```
cperciva: Is `ac_cv_lib_rt_clock_gettime` set to `yes` even if the function is exposed via libc?

I can completely believe that it would be, but I just want to double-check that autoconf is insane in the way you're assuming here.

gperciva: Looks like it.  `configure` contains:

```
# Check for clock_gettime.  On some systems, this is provided via librt.
{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for clock_gettime in -lrt" >&5
$as_echo_n "checking for clock_gettime in -lrt... " >&6; }
if ${ac_cv_lib_rt_clock_gettime+:} false; then :
  $as_echo_n "(cached) " >&6
else
  ac_check_lib_save_LIBS=$LIBS
LIBS="-lrt  $LIBS"
cat confdefs.h - <<_ACEOF >conftest.$ac_ext
/* end confdefs.h.  */

/* Override any GCC internal prototype to avoid an error.
   Use char because int might match the return type of a GCC
   builtin and then its argument prototype would still apply.  */
#ifdef __cplusplus
extern "C"
#endif
char clock_gettime ();
int
main ()
{
return clock_gettime ();
  ;
  return 0;
}
_ACEOF
if ac_fn_c_try_link "$LINENO"; then :
  ac_cv_lib_rt_clock_gettime=yes
else
  ac_cv_lib_rt_clock_gettime=no
fi
{ $as_echo "$as_me:${as_lineno-$LINENO}: result: $ac_cv_lib_rt_clock_gettime" >&5
$as_echo "$ac_cv_lib_rt_clock_gettime" >&6; }
if test "x$ac_cv_lib_rt_clock_gettime" = xyes; then :
  cat >>confdefs.h <<_ACEOF
#define HAVE_LIBRT 1
_ACEOF
```

I mean, assuming that adding a potentially-unnecessary `-lrt` to the linker won't break it.  Which seems to be an assumption that autotools makes.

cperciva: Hmm.  The case of "adding `-lrt` breaks things" is basically what our `POSIX/posix-l.sh` checks for... can you check if this breaks on OS X?  IIRC it now has clock_gettime in libc but does not have a librt?  In FreeBSD I added some special handling for libraries which POSIX says should exist but don't... I'm not sure if OS X handles that correctly.

It might be safer to tell autoconf "link to librt if we can" followed by "check if clock_gettime works" without specifying where to find clock_gettime.

gperciva: I did check that it works on OS X (https://github.com/Tarsnap/tarsnap/pull/287#issuecomment-334334811), with both the native headers and the XCode 8 headers.  But I'll poke around and look at why `POSIX/posix-l.sh` is needed for some things but apparently not with autotools.

cperciva: Oh, I see what the autoconf bits are doing.  They're doing exactly what I suggested: Checking "does linking to librt make clock_gettime work", and then checking "does clock_gettime work".

So I think the answer here is just to keep the `ac_cv_func_clock_gettime` check but drop the `ac_cv_lib_rt_clock_gettime` check -- the former tells us whether we have a working `clock_gettime`, and the latter just says where to find it (which we don't care about).

cperciva: BTW on the libraries issue, autoconf and libcperciva take opposing approaches: Autoconf says "check if we need library foo in order to get symbol bar; if yes, link to it", while the libcperciva code says "bar might be defined by library foo, and POSIX says that foo should exist, so link to it unless it's broken".  In other words, autoconf defaults to not linking to libraries unless they're determined to be needed, while libcperciva defaults to linking to libraries unless they're determined to be broken.

gperciva:
Nope, `ac_cv_func_clock_gettime` is not enough.  I admit that `ac_cv_lib_rt_clock_gettime` only worked by accident on my system, though.

```
for ac_func in clock_gettime
do :
  ac_fn_c_check_func "$LINENO" "clock_gettime" "ac_cv_func_clock_gettime"
if test "x$ac_cv_func_clock_gettime" = xyes; then :
  cat >>confdefs.h <<_ACEOF
#define HAVE_CLOCK_GETTIME 1
_ACEOF

fi
done
```

It's true that `ac_fn_c_check_func` includes the `ac_fn_c_try_link` command, but that still succeeds.  Maybe due to the "weak linking"?

```
Grahams-Mac-mini:build gperciva$ gcc -o conftest -I/usr/local/opt/openssl/include  -I/usr/local/opt/openssl/include  -L/usr/local/opt/openssl/lib  conftest.c
Grahams-Mac-mini:build gperciva$ echo $?
0
Grahams-Mac-mini:build gperciva$ ./conftest 
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /Users/gperciva/src/tarsnap/build/./conftest
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /Users/gperciva/src/tarsnap/build/./conftest
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5
Grahams-Mac-mini:build gperciva$ 
```

I'll do more digging.

gperciva:
Incidentally, this looks like a big failure of communication within the open-source community.  Tons of projects are encountering this, and it looks like most people are reinventing the wheel each time.  :(

cperciva:
This is obsoleted by #290, right?

gperciva:
Right.  I kept it open because I wasn't certain if you'd like using `libcperciva/POSIX/` in principal, but now that's ok, I'll remove this PR.

PR #290, opened 2017-10-05
import latest libcperciva util/monoclock.c

gperciva:


gperciva:
Draft alternate solution to autotools + OSX 10.11 + XCode 8's problem with `clock_gettime()`: just re-use the existing `libcperciva/POSIX/` code.  Works on my system.

- I doubt that this method plays well with a read-only source directory (which is not something we worry about with other software, but the autotools method doesn't normally touch the source, so if you like this approach in theory, I'd like to get this working again).
- this draft doesn't use `LDADD_POSIX` anywhere.  If we do, we'll likely end up with two `-lrt` in the library list -- not necessarily a huge problem, but it's annoying.  Not certain what to do about this.

cperciva:
I was avoiding mixing the libcperciva POSIX code with autoconf due to impedance-mismatch issues, but I'm inclined to say that it might be the best solution here.  I think we can let autoconf take care of the libraries, since AFAIK it manages to do that correctly.

BTW we can drop the hard-coded `POSIXFAIL_MSG_NOSIGNAL` if we're picking it up from the POSIX code.

configure.ac
```diff
@@ -498,5 +498,11 @@ fi
 AC_SUBST([BASH_COMPLETION_DIR])
 AM_CONDITIONAL([INSTALL_BASH_COMPLETION],[test "x$with_bash_completion_dir" != "xno"])
 
+# Run the libcperciva POSIX compatibility tests and pass values to the Makefile.
+AC_MSG_NOTICE([checking POSIX compatibility...])
+LIBCPERCIVA_DIR=${srcdir}/libcperciva
+CFLAGS_POSIX=`export CC="${CC}"; command -p sh ${LIBCPERCIVA_DIR}/POSIX/posix-cflags.sh "$PATH"`
```
gperciva: Note that we no longer need the `cd ${LIBCPERCIVA_DIR}/POSIX &&` line in there.

gperciva:
This works on linux and OSX.

- it now plays nice with a read-only source tree.
- it doesn't import `posix-l.sh`, nor use `${LDADD_POSIX}` at all.

cperciva:
Looks good, but can you make the out-of-tree-build changes in libcperciva first?  (I think I can see what you changed, but it's easier if there's an actual diff...)

Also, POSIX/README is very out of date -- it only mentions the LDADD bits (which you're not even importing here).  Can you add a paragraph about the POSIXFAIL defines?

gperciva:
Err, sure, I'll put the out-of-tree-build stuff into libcperciva after yoga.  But, y'know, the commit *does* provide an actual diff:
https://github.com/Tarsnap/tarsnap/pull/290/commits/99376e62744c6735d4c3ecf603cbb2443038aabc
(since I know that you tend to look at the all-in-one "files changed" view in github, rather than the actual logical grouping of changes I carefully curate in the form of individual commits.)

cperciva:
Err, yeah.  The out-of-tree-build bits look fine.  Have I mentioned that I hate github's UI?

gperciva:
The requested libcperciva changes are now #295.

(github doesn't have a "this PR depends on this other PR, so I'm using the `needs-rebase` label to indicate that you shouldn't bother looking at this PR right now.)

gperciva:
These two commits come directly from libcperciva.  I'm testing it on OSX right now just to double-check.

(this PR is still open because otherwise I'd ruin github's tracking of branches and PRs.)

gperciva:
Yep, looks good on 10.11 with Xcode 8.

PR #292, opened 2017-10-13
Improve manuals

mptre:
Couple of patches related to the manuals and especially improving them on OpenBSD.


cperciva:
LGTM.  @gperciva, I assume this won't conflict with anything you're touching documentation or autoconf wise?

gperciva:
LGTM.  I'll run `sh release-tools/mdoc2man.sh` after this is merged.  (that gets the `.mdoc` changes into `-man.in`).

PR #295, opened 2017-10-29
Libcperciva posix

gperciva:


gperciva:
This is split from #290 (and includes the latest updates to `libcperciva/POSIX/`).

PR #297, opened 2017-11-10
Dump config

gperciva:


gperciva:
Sample output (with a silly example):
```
$ ./tarsnap --print-stats --dump-config --configfile ~/.tarsnaprc --configfile ../tar/tarsnap.conf.sample 
Command-line:
  ./tarsnap --print-stats --dump-config --configfile /home/td/.tarsnaprc --configfile ../tar/tarsnap.conf.sample
Reading from config file: /home/td/.tarsnaprc
  print-stats
  humanize-numbers
  cachedir ~/.test-tarsnap/backup.cache
  keyfile ~/.test-tarsnap/backup.keys
  checkpoint-bytes 1M
Reading from config file: ../tar/tarsnap.conf.sample
  cachedir /usr/local/tarsnap-cache
  keyfile /root/tarsnap.key
  nodump
  print-stats
  checkpoint-bytes 1G
Reading from config file: /home/td/.tarsnaprc
  print-stats
  humanize-numbers
  cachedir ~/.test-tarsnap/backup.cache
  keyfile ~/.test-tarsnap/backup.keys
  checkpoint-bytes 1M
Reading from config file: /usr/local/etc/tarsnap.conf
  nodump
  print-stats
  no-humanize-numbers
                                       Total size  Compressed size
All archives                               5.4 MB           1.8 MB
  (unique data)                            2.4 MB           918 kB
td@mac: ~/src/tarsnap/build (dump-config)
```

gperciva:
One quibble is that the config lines are printed in order that `tarsnap` reads them, but later options don't invalidate earlier ones.  For example, `/usr/local/etc/tarsnap.conf` is the last file that's read, and it contains `no-humanize-numbers`, but the earlier `humanize-numbers` takes priority.

cperciva:
Ok, I can see why this could be useful for debugging user errors... but why make it an *option*?  This seems like it should simply be a *command* -- I mean, the reading and parsing of configuration files won't depend on whether someone is going to be creating or extracting or deleting an archive.

gperciva:
Good point; if I was printing out the values after parsing, then I admit it would be more useful as an option -- it would confirm whether or not `tarsnap` was understanding the command-line options in the way the user expected.  But it's true that this PR doesn't do that; the only unexpected thing it might expose (as an option rather than a command) is shell expansion.

I think it's still worth keeping as an option, for future expansion possibilities.

- Longer-term, there's a possibility that we might use the magic getopt, in which case we can easily print the parsed command-line options.  (I'd like to move `getopt_dummy` out of the "internal interfaces" section of `getopt.h` to avoid printing `(dummy)`, but that's a trivial side issue.)
  
  Or, if this option receives a lot of enthusiasm from users, I could try to find a less hack-y way of getting the parsed info from the existing `bsdtar_getopt()`.

- I'd also like to add a `--dump-include-exclude` option (or some other name), which spams info about `--include` or `--exclude` options.  That would definitely be an option rather than a command.

Issue #112, opened 2016-01-07
Users get confused by not seeing anything happening


cperciva:
Like all good UNIX utilities, tarsnap runs quietly unless it has a reason to be noisy (e.g., an error occurred, or the user asked for verbosity).  Unfortunately this is provoking a FAQ: "Tarsnap doesn't seem to be doing anything, did it hang?"

I think the solution to this might simply be better documentation, but I'm creating the issue here in case anyone wants to chime in with suggestions.

gperciva:
I personally don't think it would be terrible if an abbreviated version of `--print-stats` happened automaticaly.  Namely, something like:

```
$ tarsnap --dry-run -c ~/backup
tarsnap: Removing leading '/' from member names
                                       Total size  Compressed size
This archive                               2.7 MB           915 kB
New data                                   1.5 MB           9.7 kB
```

I figure there's no way that you would accept that, though.  :|

Another option could be making `-v` imply `--print-stats` (or `--print-short-stats` for the above material?).  I mean, `-v` by itself doesn't have any kind of "completed" message.  Sure, experienced UNIX people know that the lack of an error message means that it completed (or they could even check `echo $?`), but not everybody knows that.

cperciva:
Adding `--print-stats` wouldn't help.  The problem is that people are pointing tarsnap at many GB of data, and seeing nothing happen, when in fact tarsnap is running smoothly uploading data.  They think it got stuck because it hasn't output anything and hasn't exited either.  Then they hit ^C.

gperciva:
Hmm.  What about a `--upload-progress` flag which printed how much data was uploaded (and perhaps also how much data was read from disk / in cache) ?  As discussed previously, we can't tell them "uploaded 1.2 / 7.5 GB", but at least they could see "uploaded xyz MB" updating in realtime.

shinnok:
Any progress parameter would be useful for the GUI too. I'd be happy to incorporate an extra piece of useful info along with the loading gif.

On the original FAQ, personally I have nothing against the command line client running silently by default. I think it works well like this in true allegiance to UNIX utils. Any command line user should be apt enough to check on its status on his own or ask for verbosity. How about also printing network progress on SIGINFO along with current file.

zatricky:
I see isatty() is used elsewhere in the project. Not uncommon to use that to decide whether or not to print something useful.

cperciva:
On 03/16/18 00:17, Brendan wrote:
[...]

PR #298, opened 2017-11-19
Add --upload-progress-bytes

gperciva:


gperciva:
First draft of addressing #112 (lack of "progress" messages).  No man-page docs or bash completion yet.

This adds a `--upload-progress-bytes`, similar to `--checkpoint-bytes`, which allows users to specify how often they'd like to see an "Uploaded 123 kB" message (where "often" is measured in bytes uploaded).

Notes:
- It's true that users with a lot of data that doesn't change could still be confused by not seeing any progress messages, but if somebody has TB of data, they'll (hopefully) test it on a smaller batch first anyway.
- I considered doing the fancy "write to console, then erase what I wrote and write new numbers" trick, but a) I wasn't certain if that was doable with POSIX-only, and b) that wouldn't play nicely if the user added `-v`.
- Yes, renaming `raisesigs()` to `raisesigs_progress()` is yucky.  I considered putting it in a separate function, but when I realized that I'd have to re-compute `totalout`, I decided to keep this functionality in the same function.  At least for this first draft.

recrypt/recrypt.c
```diff
@@ -36,6 +36,7 @@ int tarsnap_opt_noisy_warnings = 0;
 int tarsnap_opt_humanize_numbers = 0;
 uint64_t tarsnap_opt_checkpointbytes = (uint64_t)(-1);
 uint64_t tarsnap_opt_maxbytesout = (uint64_t)(-1);
+uint64_t tarsnap_opt_uploadprogressbytes = (uint64_t)(-1);
```
gperciva: Not certain if you want underscores in the second part of this name.  Sometimes you do:
```
extern int tarsnap_opt_retry_forever;
extern int tarsnap_opt_humanize_numbers;
```
but sometimes you don't:
```
--checkpoint-bytes
extern uint64_t tarsnap_opt_checkpointbytes;
```

In the end, I decided that `--upload-progress-bytes` was more similar to `--checkpoint-bytes` then `--retry-forever`.

cperciva: That spelling is fine.

tar/storage/storage_write.c
```diff
@@ -111,6 +114,20 @@ raisesigs(struct storage_write_internal * S)
...
+		    tarsnap_opt_uploadprogressbytes) {
+			lastuploadprogress = totalout;
+			if (tarsnap_opt_humanize_numbers)
+				fprintf(stderr, "Uploaded %s\n",
+				    humansize(totalout));
+			else
+				fprintf(stderr, "Uploaded %" PRIu64 "\n",
+				    totalout);
+		}
+	}
```
cperciva: Hmm, should we do this, or just raise a SIGUSR1 to ourself?

gperciva: SIGUSR1 isn't hugely valuable from a "progress" standpoint:

```
$ tarsnap -c --dry-run ~
adding home/td/texmf/tex/generic/pgf/libraries/shapes/pgflibraryshapes.geometric.code.tex (0 B / 132 kB bytes)
adding home/td/texmf/doc/generic/pgf/pgfmanual.pdf (655 kB / 8.0 MB bytes)
adding home/td/texmf/doc/generic/pgf/images/pgfmanual-mindmap-1.pdf (0 B / 147 kB bytes)
adding home/td/texmf/doc/generic/pgf/text-en/pgfmanual-en-library-external.tex (0 B / 47 kB bytes)
```

Granted, it would accomplish the main goal of reassuring users that something is happening.  But the filenames are essentially arbitrary (and many will be skipped entirely!), and the `0 B`, `655 kB`, `0 B`, `0 B` could be confusing (they doesn't convey a huge amount of "progress"!).  If we wanted to go that route, I think it would be preferable to heavily promote `-v` in the docs instead of changing any code.

I considered trying to print something like
```
adding home/td/texmf/tex/generic/pgf/libraries/shapes/pgflibraryshapes.geometric.code.tex (0 B / 132 kB bytes) [uploaded 8 kB total]
adding home/td/texmf/doc/generic/pgf/pgfmanual.pdf (655 kB / 8.0 MB bytes) [uploaded 14 kB total]
```

but getting the info from `tar/storage/storage_write.c` into `tar/siginfo.c` looked like it would be messy.  I could take another look at that if you want, though.

cperciva: Hmm, take a look at what the latest bsdtar code does.  When I hit ^T on FreeBSD I get 

```
In: 8516 files, 1845319680 bytes; Out: 1845319680 bytes, compression 0%
Current: usr/local/llvm38/bin/llc (13434880/32473040 bytes)
```

which seems like it has more functionality than what I wrote in tarsnap and merged to bsdtar.

PR #305, opened 2018-02-19
progress-bytes

gperciva:


gperciva:
New version of printing progress (formely #298).  No man page or autocompletion yet.

This version is cleaner, and applies to `-x` and `--dry-run -c` as well:

```
td@mac: ~/src/tarsnap/build/foo (print-progress)
$ ../tarsnap --dry-run -c --progress-bytes 2M ~
adding home/td/backup-tests/backup-backup/zeros-900k-b.txt (0 B / 900 kB bytes) [processed 2.1 MB bytes on disk]
adding home/td/backup-tests/backup/zeros-900k-b.txt (0 B / 900 kB bytes) [processed 4.8 MB bytes on disk]
adding home/td/backup-tests/backup-tiny/lorem-1000.txt (0 B / 1.0 kB bytes) [processed 7.6 MB bytes on disk]
^C
td@mac: ~/src/tarsnap/build/foo (print-progress)
$ ../tarsnap -x -f a3 --progress-bytes 100k
extracting tar/tarsnap-read.o [processed 115 kB bytes on disk]
extracting tar/tarsnap-matching.o [processed 291 kB bytes on disk]
extracting tar/chunks/ [processed 404 kB bytes on disk]
extracting tar/tarsnap-siginfo.o [processed 568 kB bytes on disk]
```

I anticipate some nitpicking over the `processed XXX bytes on disk]` message.  My thoughts:
- keeping a single progress message on a single line is good, even if that line doesn't fit into 80 chars.  (otherwise scripting becomes more complicated)
- I wanted to be clear that the `[]` messages referred to the local disk, not the encoded network traffic.

gperciva:
IIRC you suggested off-line that the `[processed XXX bytes on disk]` messages should go on a second line.  Is that still the case?

There was a suggestion to use `isatty()` https://github.com/Tarsnap/tarsnap/issues/112#issuecomment-373624520.  I'd rather not, for precisely the reason you suggested caution: it would be too easy for users to get confused between what they see on the console, and what they might see if they redirect to a file.  Since `--progress-bytes` is an optional command, which sends info to `stderr`, I don't think we need to worry about trying to "hide" that info.  Users can choose if they want it or not.

cperciva:
> IIRC you suggested off-line that the [processed XXX bytes on disk] messages should go on a second line. Is that still the case?

Yes.  And maybe that text could be a bit clearer -- to say that we've processed XXX bytes of *archive data*, not just XXX bytes from the file we mentioned on the previous line...

> There was a suggestion to use isatty() #112 (comment). I'd rather not, for precisely the reason you suggested caution: it would be too easy for users to get confused between what they see on the console, and what they might see if they redirect to a file. Since --progress-bytes is an optional command, which sends info to stderr, I don't think we need to worry about trying to "hide" that info. Users can choose if they want it or not.

Yeah.  Using isatty to quiet output automatically is generally an odd thing to do, but it's especially silly when users have explicitly asked for output...

gperciva:
Do you mean *compressed* archive data, or *uncompressed* archive data?  The `XXX bytes from disk` referred to uncompressed archive data -- IMO, "archive data" isn't a clear term.  I thought that "bytes on disk" was easier to understand.

gperciva:
Updated with another DRAFT.

- adding the global variable feels really icky.
- I thought it would be more useful to track the "new data", instead of "uploaded", so that it works well with `--dry-run`.

Sample:
```
$ ./tarsnap --dry-run -c --progress-bytes 300k lib
adding lib/scryptenc
Processed 3 files, 953 kB; New data: 273 kB 
adding lib/network/lib_libtarsnap_a-tsnetwork_select.o (0 B / 5.0 kB bytes)
Processed 46 files, 1.2 MB; New data: 358 kB 
adding lib/keyfile/.deps
Processed 86 files, 1.5 MB; New data: 439 kB 
adding lib/crypto/.deps/lib_libtarsnap_a-crypto_keys.Po (0 B / 7.3 kB bytes)
Processed 116 files, 1.8 MB; New data: 528 kB 
                                       Total size  Compressed size
All archives                               4.2 MB           1.1 MB
  (unique data)                            3.9 MB           1.1 MB
This archive                               1.9 MB           545 kB
New data                                   1.9 MB           545 kB
```

Extracting:

```
$ ./tarsnap -x -f a3 --progress-bytes 300k 
extracting tar/tarsnap-cmdline.o
Processed 11 files, 319 kB
extracting tar/.deps/tarsnap-bsdtar.Po
Processed 23 files, 632 kB
extracting tar/storage/tarsnap-storage_delete.o
Processed 45 files, 957 kB
^C
```

cperciva:
```
$ ./tarsnap --dry-run -c --progress-bytes 300k lib
adding lib/scryptenc
Processed 3 files, 953 kB; New data: 273 kB 
adding lib/network/lib_libtarsnap_a-tsnetwork_select.o (0 B / 5.0 kB bytes)
Processed 46 files, 1.2 MB; New data: 358 kB 
adding lib/keyfile/.deps
Processed 86 files, 1.5 MB; New data: 439 kB 
adding lib/crypto/.deps/lib_libtarsnap_a-crypto_keys.Po (0 B / 7.3 kB bytes)
```

Is there a reason the first line is at 953 kB?

tar/bsdtar.c
```diff
@@ -1770,6 +1774,20 @@ dooption(struct bsdtar *bsdtar, const char * conf_opt,
 
 		bsdtar->option_print_stats = 1;
 		bsdtar->option_print_stats_set = 1;
+	} else if (strcmp(conf_opt, "progress-bytes") == 0) {
```
cperciva: And handle --no-progress-bytes?

tar/chunks/chunks_write.c
```diff
@@ -13,11 +13,14 @@
 
 #include "chunks_internal.h"
 #include "hexify.h"
+#include "tarsnap_glob.h"
 #include "rwhashtab.h"
 #include "warnp.h"
 
 #include "chunks.h"
 
+uint64_t tarsnap_glob_newbytes = 0;
```
cperciva: Rather than having a global variable, can we pass the chunks cookie to the siginfo code so that it can pull the data out of the chunks layer statistics which we're already keeping?

gperciva:
On Wed, Apr 11, 2018 at 02:13:57AM +0000, Colin Percival wrote:
>    Is there a reason the first line is at 953 kB?

It depends on what order files are processed, since siginfo_setinfo() is called
after finishing each file.  After adding
```
printf("siginfo_setinfo: %s %zu\n", path, total_uncompressed);
```

I get:

```
$ ./tarsnap --dry-run -c --progress-bytes 300k lib
siginfo_setinfo: lib 0
siginfo_setinfo: lib/crypto 512
siginfo_setinfo: lib/libtarsnap.a 1024
siginfo_setinfo: lib/scryptenc 953856
adding lib/scryptenc
Processed 3 files, 953 kB; New data: 273 kB 
```

tar/chunks/chunks_write.c
gperciva: The issue is how to link siginfo code and the chunks layer.

- siginfo is all part of `bsdtar.h`.
- `CHUNKS_W` is owned by the multitape layer.

I've sketched a version of what I think you're suggesting, but it has a few WTF moments.  I kind-of hope that I misunderstood something.

- `chunks_write_newzbytes()` refers to chunkstats `.s_zlen` directly (where previously that was only used in `chunks_stats_internal.c` and `chunks_directory.c`.  [I could fix this by adding a `chunks_stats_calcznew()` function, but that seemed like overkill at this stage.]

- `multitape_write.c` now includes `bsdtar.h`, for access to the new `siginfo_setchunks()`.

- `siginfo.c` now includes `chunks.h`.

- `siginfo.c` now contains a `static CHUNKS_W * chunks_write_cookie`.

It might be cleaner to put `CHUNKS_W` in `struct bsdtar` directly?

gperciva:
Updated PR with a really questionable (IMO) design.  Not for merging.

tar/siginfo.c
```diff
@@ -222,3 +225,24 @@ siginfo_done(struct bsdtar *bsdtar)
...
 	free(bsdtar->siginfo);
 }
+
+/**
+ * siginfo_raisetotal(total):
+ * Look at how many bytes on disk have been processed since the last update,
+ * and trigger a siginfo_printinfo() if it exceeds tarsnap_opt_processbytes.
+ */
+static void
+siginfo_raisetotal(int64_t total)
```
cperciva: Why not take the `bsdtar` structure here (or possibly `bsdtar->siginfo`)?  And put `lastprogress` into said structure?

cperciva:
Hmm.  Maybe it would be cleaner to pass `struct siginfo_data *` from `tarsnap_mode_c` to `archive_write_open_multitape` to `writetape_open` to `chunks_write_start`, and then have `chunks_write_chunk` make a call to `siginfo_blahblahblah` to update its statistics?

It's probably a fairly large diff given how many layers the siginfo_data cookie needs to get passed through, but it would avoid the global variable and layering violation.

gperciva:
Ignore the previous comment.  (I still think that `bsdtar.h` is a mess, but it's not a serious problem here.)

gperciva:
I'm happier with this design.

The one part I'm not sold on is passing `struct siginfo_data *` as `(void **)`.  I think that we need the extra indirection because when write.c calls `archive_write_open_multitape()`, `bsdtar->siginfo` is still `NULL`, and I didn't think it was worth trying to play weird games in `write.c`.

tar/siginfo.c
```diff
@@ -191,3 +260,23 @@ siginfo_done(struct bsdtar *bsdtar)
...
+siginfo_raisetotal(uint64_t total)
+{
+	static uint64_t lastprogress = 0;
+
+	if (tarsnap_opt_progressbytes != (uint64_t)(-1)) {
+		if (total > lastprogress + tarsnap_opt_progressbytes) {
+			lastprogress = total;
+
+			/* Fake a SIGINFO (no need for an actual signal). */
+			siginfo_received = 1;
```
cperciva: Maybe a dumb question, but is there a reason to not call `siginfo_printinfo` here?

tar/siginfo.c
```diff
@@ -51,6 +51,15 @@ struct siginfo_data {
 	/* How large is the archive entry? */
 	int64_t size;
 
+	/* How many files in total? */
```
cperciva: s/files in total/files have we handled in total/ ?

Or is there a reason this comment is different from the other two?

tar/siginfo.c
cperciva: Oh, we don't have the `struct bsdtar` here.  But that just raises another question, why isn't this just part of `siginfo_setinfo`? 

And if we do that, `lastprogress` can be part of the siginfo structure rather than being a random static variable here...

tar/chunks/chunks_write.c
```diff
@@ -33,16 +33,20 @@ struct chunks_write_internal {
...
 };
 
 /**
- * chunks_write_start(cachepath, S, maxchunksize):
+ * chunks_write_start(cachepath, S, maxchunksize, siginfo, siginfo_newbytes):
  * Start a write transaction using the cache directory ${cachepath} and the
  * storage layer cookie ${S} which will involve chunks of maximum size
- * ${maxchunksize}.
+ * ${maxchunksize}.  If ${siginfo_newbytes} is not NULL, call it with
+ * ${siginfo_cookie} and the number of new bytes.
```
cperciva: That last sentence feels like it ended too soon.  "... after writing each chunk"?

tar/siginfo.c
gperciva: I used `static uint64_t lastprogress` because that's what happened in `raisesigs()` from `tar/storage/storage_write.c' (with `lastcheckpoint`).  I thought it was a nice solution, actually -- it keeps the variable scope isolated to the code that actually needs it.  Are you sure that you'd rather have it inside `struct siginfo_data`?

Good point about moving it into `siginfo_setinfo()`.

gperciva:
Should I rebase these fixes?

tar/siginfo.c
```diff
@@ -115,6 +114,7 @@ void
 siginfo_setinfo(struct bsdtar *bsdtar, const char * oper, const char * path,
     int64_t size, int file_count, uint64_t total_uncompressed)
 {
+	static uint64_t lastprogress = 0;
```
gperciva: I used `static uint64_t lastprogress` because that's what happened in `raisesigs()` from `tar/storage/storage_write.c` (with `lastcheckpoint`).  I thought it was a nice solution, actually -- it keeps the variable scope isolated to the code that actually needs it.  Are you sure that you'd rather have it inside `struct siginfo_data`?

cperciva: It's nice for scoping reasons; but it's bad because aside from scope it's a global variable.  Singletons should be avoided except when there's a very good reason for them...

cperciva:
It has been too long since I looked at this for me to have any cached context, so please go ahead and tidy up the commit history -- that way it will be clearer when I look at it ab initio.

gperciva:
Rebased PR.

The handling of "new bytes" still feels overly complicated, but given that we're passing info between radically different layers of the program, it might be unavoidable.  If you'd like to narrow the scope, I could make a PR that adds a `--progress-bytes` which prints the number of files and total data processed; printing the number of new bytes could be handled in a separate PR.

gperciva:
This is waiting until #313 is accepted first.

PR #306, opened 2018-04-13
Avoid tautological comparison

gperciva:


gperciva:
This is the only issue with clang-6.0.0.

PR #307, opened 2018-05-03
Cleanup

gperciva:


cperciva:
Are those header guards now in libarchive?  I don't normally make non-functional changes to files I get from there.

gperciva:
No; updated PR accordingly.

PR #309, opened 2018-05-18
Man "(-α mode only)"

gperciva:


cperciva:
I like this, but not enough at I'd want to be gratuitously different from libarchive.  See if they want the change.

PR #312, opened 2018-07-02
Avoid singleton in storage_write's raisesigs()

gperciva:


gperciva:
Arose from #305; I figured it would be easier to talk about a potential code change if it was an actual change that we can see easily.

I'm quite prepared to accept that there's a good reason to keep this as a singleton, but I'm curious about that reason.  This was added in 1.0.20 (the first release with `--checkpoint-bytes`)
https://github.com/Tarsnap/tarsnap/commit/5dcba8998666f31ad331e0b6edeece6e91d59401#diff-d600af067c43593acc67da4456c428fdR92

PR #315, opened 2018-08-20
build: fix OpenSSL paths for darwin

shawwn:
I noticed that `autoreconf -i` doesn't set the OpenSSL paths on OS X, causing
`./configure` to fail. After applying this patch, I was able to build Tarsnap
without problems.

PR #316, opened 2018-08-20
Building osx

gperciva:


shawwn:
FWIW, it's convention on OS X to let homebrew do as much configuration management as possible. Build instructions tend to look more along the lines of "Run `brew install openssl && ./configure && make -j4`".

To put it another way, it would be very strange if an OS X dev wanted to build against an openssl located anywhere other than `/usr/local/opt/openssl`. Whereas there might be a cost associated with making a new developer read through extra documentation to build the project .

(I don't have any strong feelings either way; I was just giving the point of view of a typical macOS developer in case it was useful.)

It's true that this PR would be closer to unix traditions, so if that's important to the project then the tradeoff might be worth it.

gperciva:
True, and we're likely going with your #315.  But if somebody is compiling Tarsnap from scratch, presumably they're interested in the details (and/or don't trust brew).

I mean, if they're comfortable with brew, I assume that they'd just do:

    brew install tarsnap

shawwn:
They could, but brew usually installs binaries rather than compiling from source. Some formulas support the -HEAD option, but I’m not sure you’d get a clone of the repo, so it would be hard to do pull requests and such.

In my case, I’m interested in Tarsnap’s algorithms. I was hoping to be able to use the pointer heap and deduplication code, but I see the license doesn’t permit that.

I’m trying to create a library that uses shared memory to communicate with different languages, which is quite similar to networking in some respects and similar to hardware design in other respects. But the goal is for JS to talk to LuaJIT, and for LuaJIT to do high performance operations on NodeJS buffer memory. In that context, tarsnap has implemented a lot of the same data structures that I would have needed to write myself.

gperciva:
Good point about trying to compile from git HEAD, I hadn't thought about that.

You might be interested in https://github.com/Tarsnap/libcperciva, which contains BSD-licensed code (in particular, datastruct/ptrheap.h and datastruct/ptrheap.c).  We have other projects which are BSD-licensed, but I suspect that anything you're interested in would be in libcperciva.

The deduplication is indeed only in the Tarsnap client source code and is not under a permissive copyright license.  You *might* find something useful in bsdiff:
https://github.com/cperciva/bsdiff
(also found in the FreeBSD source tree now), but I'm not optimistic about that.  Still, that's the best direction I can point you in.

shawwn:
Thanks!

Is there anything that the tarsnap project needs in terms of technical work / contributions? (Or some other project from the company.)

gperciva:
Actually, yes!  I'm getting ready to ask for OS X testers for tarsnap-gui 1.0.2:
https://github.com/Tarsnap/tarsnap-gui/tree/version-1.0.x

The only change from 1.0.1 is that I'm trying to handle updates via brew (https://github.com/Tarsnap/tarsnap-gui/issues/194).  Brew no longer supports `brew linkapps`, so the tarsnap-gui binary is in a location like

    /usr/local/Cellar/tarsnap-gui/1.0/Tarsnap.app/Contents/MacOS/Tarsnap

That's the path that we write to `~/Library/LaunchApps/com.tarsnap.gui`, but of course that broke when people upgraded from 1.0 to 1.0.1.  I added a `--check` option to update the binary path and re-load launchd, and if this works, I'll get that added to the brew formula for tarsnap-gui.

I _think_ that works, but this is my first interaction with launchd, and I've only spent a total of ~10 hours running OS X in the past decade, so I'm a bit unfamiliar with it.  Does that solution look reasonable?

(I'll also take a look at making a `brew cask` formula for tarsnap-gui, but I wanted to get this fix out there first so that people who already have it installed as a non-cask can upgrade safely.)

shawwn:
Cool. I’m walking back to the office at the moment, but I’ll look soon.

gperciva:
Other than that, I think that portability testing & fixes for the other Tarsnap software would be the most useful thing.  We strive for strict POSIX compatibility, but sometimes we get something wrong, or sometimes platforms don't actually follow POSIX.  Our workarounds are in libcperciva/POSIX.

The main projects to check are:
https://github.com/Tarsnap/libcperciva
https://github.com/Tarsnap/spiped
https://github.com/Tarsnap/scrypt

Also, if you use any functionality of these that it's covered by our tests, please let us know (and/or add the test(s) yourself)!  The test framework is relatively new, so I'm sure that there's all sorts of useful scenarios that aren't covered yet.

shawwn:
I'm hanging out in #tarsnap on EFNet IRC while I test tarsnap-gui, FWIW.

gperciva:
Sorry, I was having lunch.  I'm on IRC now.

shawwn:
I think this will work: https://github.com/shawwn/tarsnap-gui/tree/version-1.0.x

I've also created a cask for it: https://raw.githubusercontent.com/shawwn/homebrew-cask/04c93999260431c21ddc24c37b22dbfe5087670b/Casks/tarsnap.rb

You can test it by running `brew cask install https://git.io/tarsnap` (which simply forwards to that URL).

When the cask installs, the `postflight` clause will run [tarsnap-gui-postinstall](https://github.com/shawwn/tarsnap-gui/blob/version-1.0.x/bin/tarsnap-gui-postinstall) which takes care of removing the existing service and setting it up correctly.

If you run `./build.sh` on macOS, it will create Tarsnap.app and Tarsnap.zip, and prints the sha256sum of Tarsnap.zip. When you issue a [release](https://github.com/shawwn/tarsnap-gui/releases), you have to upload that Tarsnap.zip file to the release, then update the cask's sha256sum and submit a pull request with the new cask version.

I tested this on a fresh OS X laptop as follows: I ran `brew install tarsnap`, `brew install tarsnap-gui`, enabled automatic job scheduling, `brew uninstall tarsnap-gui`, `brew cask install tarsnap`.  Afterwards, `launchctl list com.tarsnap.gui` shows that the automatic job scheduling task was registered successfully.

Everything should install fine for existing users; you can tell them to just run `brew cask install tarsnap` after doing a release.

(Your `--check` argument seemed to work ok, but I think a cask is what macOS devs prefer for GUI apps.)

If that wasn't quite what you were needing from me, let me know. :)

shawwn:
Also, cask has the option of verifying downloads using a GPG key. This should make it possible for people to have confidence in binaries signed by 
https://github.com/Tarsnap/tarsnap-public-keys

My formula above doesn’t use this feature, but I can set it up if it’d help.

shawwn:
image

Looks like the new cask properly set up the plist. tarsnap-gui is cool.

cperciva:
@gperciva Am I right in thinking that this can be closed now?

gperciva:
I'll be following up on this in https://github.com/Tarsnap/tarsnap-gui/issues/198

gperciva:
No wait, I was confused.  It's still worth adding this material (copied from spiped, BTW), although maybe I should mention brew by name as well.

cperciva:
Ok, why do we need this now that we're setting CPPFLAGS and LDFLAGS automatically?

gperciva:
I forgot about the clear system requirements at the top of `BUILDING` (especially `2. Has OpenSSL available via -lcrypto and #include <openssl/foo>,`).

Closing this one.  The typo fix is in #326.

PR #317, opened 2018-08-22
Update README.md

shawwn:
Add Tarsnap icon, slack badge, travis-CI badge, spectrum community shield, twitter follow link, and macOS installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,15 @@
-Client code for Tarsnap
-=======================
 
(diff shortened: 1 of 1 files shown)


cperciva:
Thanks, but no.  Two things I don't like here:

1. Adding gunk into the README file which renders nicely in github with a graphical web browser but is just noise to anyone reading the file using `more`.

2. I don't want to privilege macOS here by pointing people at the homebrew installation.

PR #318, opened 2018-08-24
Add --debug-network-stats

gperciva:


gperciva:
This is a sketch of adding a `--debug-network-stats`:

```
$ ./tarsnap --list-archives --debug-network-stats
Netpacket connection ended with in / out / queued:	945	592	0
z1
zeros-1000-zeros-1000
zeros-1000
z2
Netpacket connection ended with in / out / queued:	3760	893	0
$ ./tarsnap -c --debug-network-stats -f zeros-20000 zeros-20000
Netpacket connection ended with in / out / queued:	1928	4264	0
Netpacket connection ended with in / out / queued:	711	483	0
Netpacket connection ended with in / out / queued:	781	593	0
$ ./tarsnap -d --debug-network-stats -f zeros-20000
Netpacket connection ended with in / out / queued:	1658	1745	0
Netpacket connection ended with in / out / queued:	781	593	0
Netpacket connection ended with in / out / queued:	2405	747	0
$
```

I have a different (local) branch where I tried to make a nice `--print-network-stats`, which carefully passed the option to the storage layer, accumulated network traffic into a single message shown at the end of all processing, had a nicely designed distinction between stats-related functions in `storage.h` and `storage_internal.h`, etc.

However, that got way messier when dealing with `tar/multitape/multitape_delete.c`.  While I was looking at my design and 10 commits of trying to do this the "proper" way, I really started to doubt that it was worth it.  Especially since I'd only covered write/read/delete, and not things like `--list-archives`!

So I've now changed my mind and think that this type of implementation is more suitable.  It'll be enough for us to debug stuff (particularly `--resume-extract`), and it doesn't get bogged down with ugly refactoring.

(if you agree, I'll add man-page, NEWS, and bash completion for `--debug-network-stats`.  I also don't think it's worth having `--no-debug-network-stats`, but of course I could add that as well if you want it.)

PR #320, opened 2018-08-31
Give explicit warning for server signature mismatch

gperciva:


lib/netproto/netproto_keyexchange.c
```diff
@@ -304,6 +304,7 @@ dh_received(void * cookie, int status)
 		goto err2;
 	case 1:
 		/* Bad signature. */
+		warn0("Bad Diffie-Hellman signature from server");
```
cperciva: This isn't a DH signature.  This is an RSA signature on a DH parameter.

How about "Diffie-Hellman parameter from server is not validly signed"?

gperciva:
PR updated with suggested message.

PR #321, opened 2018-09-03
Siginfo cleanup

gperciva:


tar/siginfo.c
```diff
@@ -128,10 +128,12 @@ siginfo_printinfo(struct bsdtar *bsdtar, off_t progress)
 	/* Sanity check. */
 	assert(progress >= 0);
 
-	/* If there's a signal to handle and we know what we're doing... */
-	if ((siginfo_received == 1) &&
-	    (siginfo->path != NULL) &&
-	    (siginfo->oper != NULL)) {
+	/* Quit if there's no signal to handle. */
+	if (siginfo_received == 0)
```
cperciva: I would probably spell this `if (!siginfo_received)` since that reads a bit more literally as the meaning ("if not siginfo received").

tar/siginfo.c
```diff
@@ -161,11 +162,14 @@ siginfo_printinfo(struct bsdtar *bsdtar, off_t progress)
...
 			}
 		}
+		/* --verbose mode doesn't want newlines at the end of lines. */
 		if (!bsdtar->verbose)
 			fprintf(stderr, "\n");
-		siginfo_received = 0;
 	}
 
+	/* We've handled the signal. */
+	siginfo_received = 0;
```
cperciva: Why did this move?

I don't think we want it here -- if someone hits ^T before we have path and oper set, we don't want to clear `siginfo_received` without responding to it.

tar/siginfo.c
```diff
@@ -55,7 +55,9 @@ struct siginfo_data {
 #ifdef SIGINFO
 	struct sigaction siginfo_old;
 #endif
+#ifdef SIGUSR1
```
cperciva: Technically not necessary, since POSIX says that SIGUSR1 exists.  (It is silent about SIGINFO.)

I agree with having this here for the sake of consistency, but the commit message should mention that we're doing this for consistency, not because it's actually needed.

tar/siginfo.c
gperciva: Hmm.  I was thinking that it would sometimes be useful to print a message if we *don't* have `path` and `oper`, but you're quite correct that it would change the current behaviour so it doesn't belong in this PR.  Removed and I'll check this again to see whether I want to include it in #313.

tar/siginfo.c
gperciva: done

gperciva:
Updated PR (including history).

cperciva:
Did you see https://github.com/Tarsnap/tarsnap/pull/321#discussion_r217621990 and not like that idea, or did github not display it?

gperciva:
Oops, sorry.  I read it, but I got distracted by the github "review" stuff.  PR updated.

PR #322, opened 2018-09-03
Siginfo files uncompressed

gperciva:


gperciva:
This is split from #313, and includes #321.  I suggest that we merge #321 (refactoring and cleaning up signinfo) before looking at this seriously.

This PR adds printing the \# of files and \# of uncompressed bytes (including skipped bytes) to `SIGINFO` printing, but does not add `--progress-bytes`.

BTW, when testing it, I found it useful to add

    siginfo_received = 1;

to `siginfo_setinfo()`.  Although that said, if my OS let me send a `SIGINFO` with `^T` then I wouldn't have bothered editing the code.  (Until I realized I could do that, I'd start running tarsnap, then quickly switch to a different window and run `killall -SIGUSR1 tarsnap`.)

gperciva:
PR updated, ready for examination and potential merge.

tar/write.c
```diff
@@ -1102,6 +1102,7 @@ write_entry_backend(struct bsdtar *bsdtar, struct archive *a,
 			    archive_error_string(a));
 			exit(1);
 		}
+		siginfo_skipped(bsdtar, skiplen);
```
cperciva: Is this the only place we call `archive_write_skip`?

Is there some way we can make `archive_write_skip` adjust the relevant value within libarchive so that we get the right value out from `archive_position_uncompressed`?

gperciva: oops.  Si, we call `archive_write_skip` elsewhere.

My second instinct is to add `siginfo_skipped` at the relevant places because it's not immediately obvious whether messing with `archive.file_position` might break something in libarchive.

gperciva:
Updated with messy REBASE.  There's also an `archive_write_skip()` in `archive_multitape_copy()`, but I left that unchanged since it doesn't contain any `siginfo_setinfo()` so I don't think that matters for printing stats.

tar/write.c
cperciva: > My second instinct is to add siginfo_skipped at the relevant places because it's not immediately obvious whether messing with archive.file_position might break something in libarchive.

You know, we have the source code... where does `file_position` get used in libarchive when we're writing an archive?

gperciva:
Updated with long justification in 7e07e05 about why it's safe to alter `a->file_position`.

gperciva:
Ready for merge, I think.

- I added a note about the new SIGINFO behaviour to NEWS and the man page.
- the justification for using `a->file_position` is now in  2181c00

PR #325, opened 2018-09-13
Update README.md

shawwn:
With regards to the OpenSSL libraries on macOS, I ran across a project in a similar situation as Tarsnap:

https://github.com/alexcrichton/git2-rs/blob/3d61c663bb36ae4475041a7061fe47c08e8c0527/README.md#building-on-osx-1010

This PR just copies their README.md section on OpenSSL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,16 @@ The official releases should build and install using autotools:
 
 See the [BUILDING](BUILDING) file for more details.
 
+Building on OSX 10.10+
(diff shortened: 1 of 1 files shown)


cperciva:
This seems like something which belongs in `BUILDING`?

While I'm on the topic: @gperciva We could probably mention how to install OpenSSL using brew on /compiling.html on the website?

Issue #329, opened 2018-09-28
Tarsnap fails at extracting filenames with brackets


scotchi:
If you create an archive with files with brackets in it, you can't extract them by name unless you escape them ... even when bypassing the shell.

e.g. 

    $ mkdir foo
    $ touch 'foo/bar[baz]quux'
    $ sudo tarsnap -c -f example foo/
                                          Total size  Compressed size
    All archives                        2207166128045     591061010434
      (unique data)                       11405443757       2809115624
    This archive                                 2491             1409
    New data                                     2491             1409
    $ sudo tarsnap -tf example
    foo/
    foo/foo[bar]baz

But then:

    $ sudo tarsnap -xvf example 'foo/foo[bar]baz'
    tarsnap: foo/foo[bar]baz: Not found in archive
    tarsnap: Error exit delayed from previous errors.

It works if you escape the brackets with backslashes, but that doesn't seem like it should be necessary.  To make sure it wasn't any shell injected stuff that I wasn't aware of, I did it directly with `exec` to test:

    #include <unistd.h>
    #include <stdio.h>

    int main()
    {
        execlp("/usr/local/bin/tarsnap", "tarsnap", "-tf", "example", "foo/foo[bar]baz", NULL);
        perror(NULL);
        return 0;
    }

And there the error persists.  (This got reported as a redsnapper issue.)

gperciva:
Hmm.  This is a difference in behaviour between GNU tar and BSD tar:

```
$ tar -tf brackets.tar 
aa[bb]cc
$ tar -xf brackets.tar 'aa[bb]cc'
$ bsdtar -xf brackets.tar 'aa[bb]cc'
bsdtar: aa[bb]cc: Not found in archive
bsdtar: Error exit delayed from previous errors.
$ tar --version
tar (GNU tar) 1.28
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
$ bsdtar --version
bsdtar 3.1.2 - libarchive 3.1.2
td@mac: ~/src/tarsnap/build (resume-extract)
$ 
```

[...]

PR #330, opened 2018-10-02
man: mention [] characters in pattern-matching

gperciva:


gperciva:
Addresses #329.

Before working on this, I didn't realize that our man-page already mentioned shell-style globbing patterns.  However, the man-page follows libarchive in referencing `tsch(1)`, which isn't on my system so that's not so useful.

I definitely think that adding that the patterns "follows the POSIX rules for pattern-matching" is good.  I'm less certain that we should specify that `[]` are special characters, but ultimately I figured that adding a potentially-unnecessarily clause wasn't a huge problem especially since it confused an advanced user and me.

Issue #258, opened 2017-05-10
Tarsnap could accept keyfiles missing final \n


np:
It happened to me that my key files got stripped out of their final new line character. Tarsnap was then considering them as "corrupt or truncated" which is fair.

I was wondering if this would do any harm to be a bit more lenient here.

cemeyer:
Me too, although I disagree with the misleading error message being reasonable.

https://github.com/Tarsnap/tarsnap/blob/master/lib/keyfile/keyfile.c#L300

gperciva:
@cemeyer: Doesn't "truncated" accurately describe a file that's missing the final character?  I suppose you could call it "slightly truncated" instead, but either way the file is shorter than expected.

I'd consider writing a check to warn specifically about a missing final `\n`, but what's the use case?  I mean, how does one lose the final `\n` in a benign manner?  Does it occur via minor disk corruption (which fortunately doesn't disrupt anything else in the keyfile), or by manually editing the keyfile in vim with `:set noeol`, or...?

cemeyer:
@gperciva The "truncation" characterization is only accurate if you believe that the final piece of whitespace, after the ASCII trailer, is a quintessential piece of cryptographic key file material.  And "corrupt" is inaccurate.

Regardless, the error message is unhelpful as-is.  The print could simply add the text "missing newline at end of file."  Or just handle the missing newline gracefully without error.

> I'd consider writing a check to warn specifically about a missing final \n, but what's the use case? I mean, how does one lose the final \n in a benign manner?

Copy (many text applications do not consider trailing newline significant) and paste (via e.g. `xclip -selection clipboard -out > tarsnap.key`).

gperciva:
Hmm.  I see what you mean about copying from a graphical application.

For clarity, `xclip` works by itself:
```
$ xclip ~/.test-tarsnap/tarsnap.keys 
$ xclip -o > test.keys 
$ diff ~/.test-tarsnap/tarsnap.keys test.keys
$ 
```

but this does not:
```
$ gedit ~/.test-tarsnap/tarsnap.keys
       (ctrl-a  ctrl-c, switch to alternate terminal window)
$ xclip -o > test.keys 
$ diff ~/.test-tarsnap/tarsnap.keys test.keys
69c69
< # END OF TARSNAP KEY FILE
---
> # END OF TARSNAP KEY FILE
\ No newline at end of file
$ 
```

[...]

PR #334, opened 2018-11-06
keyfile: accept a missing final newline

gperciva:


gperciva:
This is one possible solution for #258.

lib/keyfile/keyfile.c
```diff
@@ -333,6 +326,10 @@ read_base64(const char * keybuf, size_t keylen, uint64_t * machinenum,
 		keybuf += llen;
 		keylen -= llen;
 
+		/* If there was a missing EOL, don't skip past the EOL. */
```
cemeyer: I would describe the situation as EOF rather than "missing EOL," but it is not very different.

gperciva:
Updated with "skip past EOL if not EOF".

PR #335, opened 2018-11-06
keyfile: warn about a missing final newline

gperciva:


gperciva:
This is one possible solution for #258.

cperciva:
I think the other option is better -- there's no security problem inherent in having a missing EOL, since if the file is truncated (by more than just the EOL character) we'll detect and complain about that.

PR #337, opened 2018-11-13
Allow users to give a passphrase on stdin

gperciva:


gperciva:
Updated.

After some thought, I'm 80% confident that you don't want me to add `--passphrase-stdin` to `tarsnap-keygen` in this PR.  It might be tricky to handle the "tarsnap password / keyfile passphrase" combo, and we don't want to suggest that people write their tarsnap password in a shell script or something like that.  If somebody finds a scenario where this is desirable, we can add it later.

PR #339, opened 2018-11-14
Sort man-page options in (mostly) alphabetical order

gperciva:


cperciva:
I think I'd prefer `-newer-mtime-than` to come after `-newer-than`.  Easier to understand when an option refers to a *previous* option than when it refers to the *next* option.

gperciva:
Updated to avoid changing `--newer-mtime-than`.

PR #340, opened 2018-11-15
Fix misc bash completion-related issues

gperciva:


gperciva:
These arose while I was working on a script to automatically update shell completion scripts by reading a man-page.

In the case of bash completion, the script may or may not break even (if I spent thirty minutes reading the bash completion script really carefully, I might have been able to find all these issues to fix).  The main point, however, is to update more complicated cases such as the zsh script in https://github.com/Tarsnap/tarsnap/pull/226 (the zsh script splits up completions based on the tarsnap mode, e.g., `-c` will offer different autocompletions than `-x`).

gperciva:
A careful reader may note that I'm not always listing the options in the bash completion file in alphabetical order.  Once #339 and this are merged, I'll run my script and have them ordered.

gperciva:
We're missing a few more options in `longopts`, so I'll take this down for now.

gperciva:
ready now

Issue #319, opened 2018-08-31
libressl 2.7 adds openbsd 1.1 API


gperciva:
LibreSSL 2.7.0 (2018-03-21) adds the openssl 1.1 API:
https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.7.0-relnotes.txt

This required a bunch of programs to update their `OPENSSL_VERSION_NUMBER` handling, such as libarchive:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=226853

As far as I know, libressl still supports the OpenSSL 1.0.1 API, so we don't need to change anything yet.  But at some point, we should probably switch over.
(in `lib/crypto/crypto_compat.c`, we assume that any libressl is using the 1.0.1g API)

```
#ifdef LIBRESSL_VERSION_NUMBER
#undef OPENSSL_VERSION_NUMBER
#define OPENSSL_VERSION_NUMBER 0x1000107fL
#endif
```

botovq:
It wold be nice if this could be reconsidered. In LibreSSL 3.5 many types will be made opaque, including the RSA type. This will break the build of tarsnap. Users on OpenBSD-current will be affected by this before the end of the year. From the compat code that LibreSSL currently reaches, only the call to ERR_remove_thread_state(NULL) in crypto_compat_free() is still needed.

I was going to make a PR that removes the compat handling, but then I noticed the ERR_remove_thread_state() in the master branch... I'm not sure how you would prefer to handle that.

gperciva:
Thanks, I didn't know about the LibreSSL 3.5 issue.  Yes, I need to move this up on my priority list.

On a related note, the OpenSSL 3.0 release on 2021-09-07 warns that the "low level" APIs are deprecated, so I've made a few forays into updating our code but nothing serious.  I guess this is a "nice" new topic to delve into on a Monday morning!

(As for "no-one should be using libressl prior to 2.7.0"... we're very reluctant to break working code at tarsnap.  What if somebody has an openbsd 5.9 system?)

For the record, I'll probably start the crypto compat updates with
https://github.com/Tarsnap/libcperciva/issues/413

gperciva:
Fix appears in #508.

@botovq, does my commentary in https://github.com/Tarsnap/tarsnap/pull/508#issuecomment-981979940 sound reasonable?  I found that I had to call `EVP_cleanup()` with LibreSSL 2.7.0+ in order to clean up all memory.

[...]

PR #341, opened 2018-11-20
Crypto compat cleanup

gperciva:


gperciva:
sigh

lib/crypto/crypto_compat.c
```diff
@@ -213,3 +213,18 @@ crypto_compat_RSA_generate_key()
 	/* Failure! */
 	return (NULL);
 }
+
+/**
+ * crypto_compat_free(void):
+ * Free the shared memory allocated by OpenSSL.
+ */
+void
+crypto_compat_free(void)
```
gperciva: I'm not certain if you'd prefer `crypto_compat_cleanup()` here instead.  Most (if not all) of our other `_free` functions take a structure, but I figured that since this was a `(void)` it would be sufficiently clear.

cperciva:
Does this DTRT with LibreSSL?  IIRC their versions map to OpenSSL versions in interesting ways.

gperciva:
Yes-ish?  Empirically yes, theoretically yes, practically we can't rely on it working in the future (because that's just how SSL projects behave :( ).

It compiles and appears to work in:
- OpenBSD 6.1, which is libressl 2.5.3 (supports openssl 1.0.2g api)
- OpenBSD 6.4, which is libressl 2.8.2 (which added support for the opsnssl 1.1 api, but still includes the openssl 1.0.2g api)

That said, Alpine linux switched from libressl back to openssl early in 2018 [1].  Amongst their reasons, they claimed that "libressl promised to retain compatibility with 1.0.1g APIs, but has failed to do so.".  (I read the first portion of the large thread discussing this and didn't find specifics about this point, but neither did I find anybody disputing that claim.)

[1] https://lists.alpinelinux.org/alpine-devel/6073.html

At the moment, the OpenBSD libressl man pages show the expected functionality (and deprecation status) of `ERR_remove_state()` and `ERR_remove_thread_state()`:

https://man.openbsd.org/ERR_remove_state.3
https://man.openbsd.org/ERR_remove_thread_state.3

So in summary: I think we're good for now.  (Also, we have a separate issue for potential problems from libressl 2.7 onwards: https://github.com/Tarsnap/tarsnap/issues/319.  As far as I know, that's a theoretical problem so it's been low on my priority list.)

cperciva:
Ok, just wanted to make sure you'd tested with LibreSSL.  Thanks!

PR #344, opened 2019-03-31
import scryptenc updates

gperciva:


gperciva:
This imports recent changes in scryptenc from scrypt.

NOTE: 684d094 introduced (very rare) memory leaks, which were fixed in 7aa3766.  Currently, this PR contains exactly the same commits as we had in scrypt.  I would suggest that I rebase that commit so that the history of this PR doesn't contain the memleak.  (I tried the rebase locally, and it applies without any problems.)

gperciva:
Updated to rebase the memleak fix into the main commits (based on offline discussion).

PR #346, opened 2019-04-05
Use crypto_verify_bytes

gperciva:


gperciva:
Imported from https://github.com/Tarsnap/scrypt/pull/161

PR #348, opened 2019-04-14
crypto_hash: remove unnecessary memset()

gperciva:


cperciva:
I think these could just go away entirely.  Doesn't sha256.c wipe the state in *_Final?

gperciva:
Updated to remove unnecessary `memset()` instead.

PR #352, opened 2019-05-07
Remove unnecessary "bsdtar_platform.h" in lib

gperciva:

PR #354, opened 2019-05-08
Add & use platform.h

gperciva:


gperciva:
Draft.  AFAIK the code is correct, but the git commit messages need some work.  The first part of this PR is #352, and I'll rework the later commit messages in that style if it's good.

gperciva:
Update PR, rebased on master.

cperciva:
I'm not convinced about this being `tar/platform.h`.  How about `lib/platform.h`?

PR #368, opened 2019-10-08
Travis-CI: switch to xenial

gperciva:


gperciva:
For clarity: trusty was 2014; xenial was 2016.

PR #369, opened 2019-10-15
Allow siginfo to print a final message

gperciva:


gperciva:
This is the last of the changes in #313 which affect somebody who did *not* specify `--progress-bytes`.  All of the later code changes are "hidden" under
`if (bsdtar->option_process_bytes != 0) {`

Issue #370, opened 2019-10-16
Misleading tarsnap error on negative account balance


ralish:
Assume the Tarsnap account associated with a given key has a negative balance. When attempting to perform operations against the account the following error is seen:
```
tarsnap: Network protocol violation by server
tarsnap: Error fscking archives
tarsnap: Error exit delayed from previous errors.
```

While a negative balance of course means Tarsnap should not be expected to work, the above error is extremely misleading. The [documentation on the error](https://www.tarsnap.com/network.html) also further reinforces this. In our case we missed a notification that our account needed credit added so weren't aware of the problem, and the Tarsnap error message sent us down entirely the wrong troubleshooting path.

We're using the latest Tarsnap package (1.0.39-2) from the APT repository on Ubuntu 18.04 x64. The command being run was an `--fsck` to rebuild the cache after migrating a service to a new server.

jungle-boogie:
Hi,

See issue https://github.com/Tarsnap/tarsnap/issues/328

This was testing the exit code, but you'll notice the output message indicated what exactly to check. Must be an edge case that hasn't been checked for account balance before.

ralish:
@jungle-boogie: Yeah, that sounds plausible. The output above is the entire total output but the exit code was non-zero (as it should be).

cperciva:
Wow, this is odd and definitely something we should fix.  @gperciva can you reproduce this in your test environment?

gperciva:
Confirmed:

```
$ ./tarsnap --keyfile local.keys --cachedir cache/ --fsck
tarsnap: Network protocol violation by server
tarsnap: Error fscking archives
tarsnap: Error exit delayed from previous errors.
```

Server log shows `tarsnap-server: Disallowing transaction start due to ENOMONEY`, so the problem is definitely in the client.  Fix incoming.

gperciva:
This was fixed in #371.

PR #371, opened 2019-10-16
fsck ENOMONEY

gperciva:


tar/storage/storage_transaction.c
```diff
@@ -327,16 +327,26 @@ storage_transaction_start(NETPACKET_CONNECTION * NPC, uint64_t machinenum,
...
+		switch (type) {
+		case 0:
+			/* write. */
+			warn0("Cannot start write transaction: "
+			    "Account balance is not positive.");
+			warn0("Please add more money to your tarsnap account");
+			goto err0;
+		case 2:
+		case 3:
+			/* fsck. */
```
gperciva: at first I wrote
```
/* Write transaction. */
```
but then I had a problem with fsck.  Would it be
```
/* fsck transaction. */
```
or
```
/* Fsck transaction. */
```

The latter makes me cringe, but the former looks bad in context (of other comments having an initial capital letter).

tar/storage/storage_transaction.c
```diff
@@ -327,16 +327,26 @@ storage_transaction_start(NETPACKET_CONNECTION * NPC, uint64_t machinenum,
 		warn0("Sequence number mismatch: Run --fsck");
 		goto err0;
 	case 2:
-		/* We should only get this for a write transaction start. */
```
gperciva: I'm not certain what you were thinking at the time, so I don't know if this is a decent fix.

The previous significant changes to this file were in 2010-06-27 for tarsnap 1.0.27 -- that added the `fsck` with `whichkey` being 2 or 3, for example.  This particular message was introduced in 2008-10-14 for tarsnap 1.0.14.

gperciva: oh, in case it helps, here's some git hashes:

1.0.14: fe41096be6d1ec38ea35a175f3d82605e104b604
1.0.27: 04fdb829bb369ad099e3b5bfe35086e0cb93f2c9

gperciva:
Draft fix for #370.  I'm rather vague about how the storage layer works.

You wouldn't believe how many times I typo'd `ENOMONKEY`.

tar/storage/storage_transaction.c
cperciva: Maybe `/* --fsck transaction using write key. */` and likewise s/write/delete/ ?

cperciva: I take back the second part of that.  This code path should only be triggered when using the write keys (whether fscking or not).

tar/storage/storage_transaction.c
cperciva: Yeah, I was right that this had to do with using the write keys for fsck starting at 1.0.27.  The case of using the delete key for fsck is fine already (we'll never get ENOMONEY in that case).

gperciva:
I think that this one is ready for merge.  I also added a "Progammer error" for "non-positive account balance" with any other operations.

tar/storage/storage_transaction.c
```diff
@@ -330,16 +330,27 @@ storage_transaction_start(NETPACKET_CONNECTION * NPC, uint64_t machinenum,
...
+		case 3:
+			/* --fsck transaction using the write key. */
+			warn0("Could not begin fsck: "
+			    "Account balance is not positive");
+			warn0("Please add more money to your tarsnap account");
+			goto err0;
+		default:
+			/* We don't expect account balance for other ops. */
+			warn0("Programmer error: "
+			    "Unexpected non-positive account balance");
```
cperciva: Is there a reason this isn't "... account balance error"?  It's the error which is unexpected here, not the balance.

gperciva: ok.  I fixed this via a rebase, since it's only a 2-line commit.

PR #373, opened 2019-10-24
comments: clarify fsck with the write key

gperciva:


gperciva:
These arose while I was looking at #371.

PR #375, opened 2019-11-19
Libcperciva parsenum import

gperciva:


gperciva:
After merging this, you'll need to run
```
autoreconf -i
```
because it modifies the build system.

PR #378, opened 2020-03-14
Initialize siginfo before use in 't' mode

gperciva:


gperciva:
An alternate solution would be to hide the uses of `siginfo_*` in 763fa25 with

    if (mode == 'x')

... actually, the more I think about it, the more I'm leaning towards that solution.

gperciva:
I uploaded the alternate solution to this PR; if you prefer it, I'll twiddle the history accordingly.  Do not merge this branch in the current state.

cperciva:
The revised solution looks good to me.

gperciva:
Revised history, and added extra comments explaining **why** this is only in 'x' mode, because otherwise I'll forget and be tempted to "clean it up".

I debated between `/* Siginfo... */` and `/* siginfo`; going with the latter becaues that's what the function / variable / file is called.  I could capitalize it if the grammar offends you, though.

PR #379, opened 2020-03-19
Begin tests

gperciva:


gperciva:
DRAFT for discussion.

This adds our standard `make test` framework that we use in other projects.  Tests 01 and 02 run just fine, but test 03 reveals that git master is broken on i386.  (And I'm sure 32-bit ARM as well, although I haven't tried the automatic test on it.)

This breakage can be fixed with one of:
    https://github.com/Tarsnap/tarsnap/pull/380
    https://github.com/Tarsnap/tarsnap/pull/381
I think that first (`platform.h`) is a "less surprising" solution.

CAUTION: in order to reveal the breakage, I had to add a keyfile -- even though I'm running `-c --dry-run`, the same code path isn't used with simply `--no-default-config` and no manual keyfile specified.  This PR adds a `fake-keygen` binary, which create a keyfile for machinenum 0 and claims to have all user keys, but actually can't communicate with the server (because it's not registered).

I expect that you'll reject this idea, for good reason.  It's here for brainstorming, in case you can think of a different way of reproducing the breakage.

CAUTION 2: I'm not convinced that this is entirely up to our code standards.  If/when we've sorted out the general solution desired, I'll proofread it some more.

gperciva:
PS: I've checked that both PRs fix the problem, both by running commands manually (with my real keyfile), and after rebasing on top of this `tests` PR running the automatic tests.

I know that we can't test much without a real keyfile... but I view `make test` as more of a sanity check of compilation.  For example, does our usage of autotools work on the latest HP-UX, or Solaris, or NetBSD on ARM64?  If somebody has a weird platform, it would be nice if they could check that `tarsnap` compiles and does something, without having to register (or copy) a keyfile.

One option would be to add tests 01 and 02, and postpone the `fake-keygen` thing for later.

gperciva:
PPS: yes, I'm planning on expanding the tests to use a real keyfile (but no network) if they specify a `TARSNAP_TEST_KEYFILE` environment variable.  If you really don't like `fake-keygen` then we can rely on those tests, although it makes it slightly more awkward to test tarsnap on assorted platforms.

I'm not certain whether to include a `TARSNAP_TEST_USE_MONEY` variable, or whether to keep those network-required tests in a separate repository.  I'm leaning towards keeping them separate.

gperciva:
Revised PR to only contain importing the test framework and two tests.  Everything questionable will be postponed to a later PR.

I'd like to get this in so that I don't need to re-run `autoreconf` and `configure` when I change branches (since the initial test commit modifies `Makefile.am`).

PR #380, opened 2020-03-19
tar/: add "platform.h" to all .c files

gperciva:


gperciva:
DRAFT for discussion.

See https://github.com/Tarsnap/tarsnap/pull/379#issuecomment-601003523

gperciva:
Rebased onto master, added more to the commit message, and verified that it fixes the problem on i386.

PR #381, opened 2020-03-19
build: extract ABI-changing defines

gperciva:


gperciva:
DRAFT for discussion.

See https://github.com/Tarsnap/tarsnap/pull/379#issuecomment-601003523

cperciva:
I like this more than #380 but how did you decide which definitions affected the ABI?

gperciva:
I read through the `config.h` produced on my FreeBSD 12.1 machine and went with my gut instinct for each of the ~180 values.

In other words, no rigorous method.  The list in `lib/ABI/extract-abi.sh` is easy to update, but it's likely that manual updates in response to bug reports will be necessary.

(config.h contains commented-out defines, so I'm cautiously optimistic that the list is valid for autoconf 2.69 on any platform.  But the only way to be certain would be to check the whole 10524-line `configure`.)

That's why I think that #380 is a better solution -- it avoids (partially) recreating what autotools already does, and especially re-evaluating the list for different versions of autotools.  Just adding `#include "platform.h"` means much less future maintenance!

cperciva:
Hmm.  I see your point about the `#include` avoiding problems if autoconf decides to add an ABI-changing #define which we don't know about.

On the other hand, we're just *assuming* that none of the bits in `/lib/` or `/libcperciva/` have ABIs which autoconf will screw with, right?  I mean, you're not proposing adding the `#include` to them... right now I think we're safe because they don't use `off_t` but hypothetically if autoconf does something we don't expect I don't know if that would guarantee anything.

cperciva:
Can we make the autoconf build add "-include config.h" to everything we compile?

gperciva:
> On the other hand, we're just assuming that none of the bits in /lib/ or /libcperciva/ have ABIs which autoconf will screw with, right?

I'd actually prefer to separate non-POSIX stuff into `lib-platform`, as sketched in https://github.com/Tarsnap/tarsnap/pull/382#issuecomment-610755736.  It's only 5 files (or rather, 10 files consisting of 5 .h-.c pairs).

Granted, that still leaves the question of `off_t`, but it would still help to isolate the "attack surface".

I'll look into `-include config.h` tomorrow.

gperciva:
`-include` is not supported in `c99`.
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html

cperciva:
Hmm, does it really matter though?  I mean, we're already doing lots of non-POSIX stuff in tarsnap...

gperciva:
At the moment, tarsnap compiles on FreeBSD with `c99`.  If we add `-include` to `CFLAGS`, then it doesn't.  I think that would be a big step away from our claim in `BUILDING`, namely
> Tarsnap should build and run on most IEEE Std 1003.1 (POSIX) compliant
systems which
>   1. Includes the Software Development Utilities option,
>   ...

I have other quibbles about `-include`, such as whether or not it plays nice with `libarchive/archive_platform.h` on all platforms [1], and how unintuitive it is[2].

[1] I suspect that it's ok, but if I had ten cents for every time I thought that an automake / posix compatibility / build system thing would be ok, I could buy a coffee & muffin combo at Whole Foods.

[2] unintuitive isn't quite the right word, because nothing relating to any build system is "intuitive".  Rather: it's a solution that I've never encountered before, whereas `#include "config.h"` (or some variant) is the recommended approach, which everybody does.

Adding "platform.h" to all `tar/` files as #380 has the advantage of being simple & obvious.  Build systems are messy enough without trying to be clever [3].

[3] with the possible exception of `cpusupport-config.h`, which was a very clever but nevertheless positive build-system tool.

cperciva:
Hmm, ok.  I'm kind of tempted to say that we should "compile" C files with a shell script which cats the autoconf header and the C file together and then passes the result to a compiler, but I think that's too convoluted even for my taste.

Ok, we'll go with added `#include`s -- we just need to remember be careful in case we ever add library code which uses `off_t` or `ino_t` or anything else which linux might break ABIs on.  (Possibly `time_t` as well?  I don't know what they're doing about Y2k38 on i386.)

PR #382, opened 2020-04-08
Isolate platform-specific lib/

gperciva:


gperciva:
Actually, I should call it "an engineer's handwaving of concept" instead of "proof of concept", because there's still build system stuff to work out.

The idea is this:
- `libcperciva/` is (mostly? completely?) strictly POSIX, with the quibbles provided by the `POSIXFAIL` stuff which is already integrated in autotools.
- `lib/` will become strictly POSIX (again allowing for POSIXFAIL).
- `lib-platform/` will contain the few files from `lib/` which is not POSIX.

Makefile.am
```diff
@@ -102,8 +102,6 @@ lib_libtarsnap_a_SOURCES=					\
 		lib/crypto/crypto_keys_subr.c			\
 		lib/crypto/crypto_passwd_to_dh.c		\
 		lib/crypto/crypto_rsa.c				\
-		lib/crypto/crypto_scrypt.c			\
```
gperciva, line 105: I'm less confident about this change than the other ones, because of `crypto_scrypt_smix*`.

- should they stay in `lib/` because they don't depend on `platform.h` ?
- should they move to `lib-platform` because `crypto_scrypt.*` is moving?
- or is this a sign that this whole idea needs a bit more thought and/or a different approach?

Makefile.am
```diff
@@ -247,6 +247,8 @@ lib_libtarsnap_a_CPPFLAGS=				\
 		-I$(top_srcdir)/lib/network		\
 		-I$(top_srcdir)/lib/scryptenc		\
 		-I$(top_srcdir)/lib/util		\
+		-I$(top_srcdir)/lib-platform		\
+		-I$(top_srcdir)/libarchive		\
```
cperciva, line 254: Not sure I understand why this line is here.

gperciva: Oops, mistake.

At one point, I was experimenting with sorting all these lists alphabetically, but then I decided that should be in a separate PR (if I even went through with it).  I evidently didn't finish undoing those changes.

Makefile.am
cperciva: I'm fine with leaving them in `/lib/`.

cperciva:
Looks good, please rebase.

gperciva:
Rebased, ready for merge.

PR #383, opened 2020-04-16
Treat a passphrase error in --dry-run the same regardless of keyfile origin

gperciva:


gperciva:
The overall problem is the discrepancy in these commands:

```
$ tarsnap --config conf --keyfile hunter2.key --cachedir cache/ -c --dry-run -v recrypt/
Please enter passphrase for keyfile hunter2.key: 
tarsnap: Passphrase is incorrect
tarsnap: Error decrypting key file: hunter2.key
tarsnap: Cannot read key file: hunter2.key
$ tarsnap --config conf --cachedir cache/ -c --dry-run -v recrypt/
Please enter passphrase for keyfile hunter2.key: 
tarsnap: Passphrase is incorrect
tarsnap: Error decrypting key file: hunter2.key
tarsnap: Ignoring cachedir due to missing or invalid keyfile in config file.
a recrypt
a recrypt/.deps
a recrypt/.dirstamp
a recrypt/tarsnap-recrypt.1
a recrypt/tarsnap_recrypt-recrypt.o
a recrypt/.deps/.dirstamp
a recrypt/.deps/tarsnap_recrypt-recrypt.Po
td@beastie: ~/src/tarsnap/build (keyfile-passphrase-dry-run)
$ 
```

(where `conf` contains `keyfile hunter2.key`)

I think they should behave the same way, and I favour the "bail for a bad passphrase" approach.

This PR is a sketch of how it can be achieved; I haven't proofread it or checked in valgrind yet.  It's more invasive to `keyfile.c` than I'd like, but I can't see a nicer way of returning a "failed due to passphrase" error from `keyfile_read()`.

gperciva:
(technical info here, from the commit message of 2e35bd3)

Previously,
- if we're doing a --dry-run,
- given a passphrase-protected keyfile,
- specified in a config file (but not the command-line),
- the user fails to provide the correct passphrase,
tarsnap printed the
```
    Ignoring cachedir due to missing or invalid keyfile in config file.
```
message and continued processing the --dry-run.

This commit changes that behaviour to bail instead.  Note that "bail for
a bad passphrase" is already the behaviour if the --keyfile is specified
on the command-line; the change is that now tarsnap behaves the same way
with a bad passphrase, regardless of whether the keyfile was specified
on the command-line or config file.

This arose on 2016-01-06, when we added

>    Allow dryrun with an unusable keyfile in config
>    dd4558f

so that people could generate --dry-run stats without having to modify
the default /usr/local/etc/tarsnap.conf (and without adding
--no-default-config).

gperciva:
I think I've thought of a less invasive solution, which involves no modificationes to `keyfile.c`.

- if we're doing a `--dry-run`, `bsdtar.c` checks if the keyfile exists first.
- if it doesn't, but the keyfile was specified in a config file, then it pretends that there's no keyfile.

Basically, instead of special-casing the output of `keyfile_read()`, we special-case avoiding descending into `keyfile_read()` in the first place.

gperciva:
I sketched an alternate solution in https://github.com/Tarsnap/tarsnap/pull/384.

PR #384, opened 2020-04-17
bsdtar: bail on --dry-run given a bad passphrase

gperciva:


gperciva:
Alternate solution to https://github.com/Tarsnap/tarsnap/pull/383; this one doesn't require any modifications to `lib/keyfile/`.  I'm not a fan of the `goto` solution I added, but that's easily tweaked if you prefer this approach.

tar/bsdtar.c
```diff
@@ -1938,3 +1948,14 @@ argv_has_archive_directive(struct bsdtar *bsdtar)
 
 	return (0);
 }
+
+/* Return 0 on success, -1 on failure. */
+static int
+file_readable(const char * filename)
```
cperciva, line 1954: Is there a reason you're not using access(2) here?

tar/bsdtar.c
```diff
@@ -1001,19 +1002,28 @@ main(int argc, char **argv)
 
 	/* Attempt to load keyfile. */
 	if (bsdtar->keyfile != NULL) {
+		/*
```
cperciva, line 1005: It seems that this would be better hoisted above this block, e.g.

```
if (bsdtar->keyfile && bsdtar->keyfile_from_config &&
    bsdtar->option_dryrun && access(bsdtar->keyfile, F_OK) {
        free(bsdtar->keyfile);
        bsdtar->keyfile = NULL;
        bsdtar->config_file_keyfile_failed = 1;
}
```

cperciva: (And that way it doesn't need the goto.)

gperciva:
Huh, I had no idea that `access(2)` existed.  Changed as suggested, suitable for detailed review now.

tar/bsdtar.c
```diff
@@ -999,18 +999,24 @@ main(int argc, char **argv)
 	if (bsdtar->mode == OPTION_VERIFY_CONFIG)
 		exit(0);
 
+	/*
+	 * Special case: if we're doing a dryrun and the keyfile came
+	 * from a config file, ignore a non-readable keyfile.
+	 */
+	if (bsdtar->keyfile && bsdtar->keyfile_from_config &&
+	    bsdtar->option_dryrun && access(bsdtar->keyfile, F_OK)) {
```
cperciva, line 1007: Is `F_OK` what we want here?  I'm not sure what the most sensible response is if a key file exists but is not readable.

Either way, it's not consistent with the comment...

gperciva:
Updated with a REBASE commit.
- comment: `readable` -> `existent`
- code: `access() -> (access() == -1)` because I still find it slightly intuitive that syscalls return "true" if there's an error.  I don't mind it if it's `if(syscall())`, but since there's already 3 other conditions in this `if` statement, I think the explicit `== -1` adds clarity.

I seriously considered having:
```
((access() == -1) && ((errno == ENOTDIR) || (errno == ENOENT) || (errno == EACCES))
```

in there as well, but then I got scared trying to imagine if any platform might have a different errno value that would be applicable here.

(Incidentally, three cheers for POSIX being so skimpy that they couldn't afford a second `S` for `EACCES`.  Like, wtf, man?!?)

Anyway.  I went with "existence" rather than "readability" because the overall goal of the special case is:
> tarsnap -c --dry-run can now run without a keyfile, allowing users to predict how much Tarsnap will cost before signing up.
> (our NEWS.md for 1.0.36.1)

The whole "ignore a missing keyfile" is for people who have a default config file installed as part of their package manager or `make install`.  As such, I think we should be as specific as possible for that scenario.

If somebody has a keyfile but it's not readable, then they're not in the case of "just installed, before signing up".  In that case, I think that an error is appropriate, so that they can investigate whether they already signed up but forgot about it, or there's another tarsnap user on the same shared system, or whatever else is happening.

cperciva:
Looks good.  BTW EACCES dates back to Version 7 AT&T UNIX, so it's not POSIX's fault; I vaguely recall that C preprocessors only considered the first 6 characters of a macro to be significant at the time.

gperciva:
Hmm, so `ENOTDIR` and `ENAMETOOLONG` were later inventions?

... I was going to say "ready for merge", but I see that you did it already.

PR #387, opened 2020-04-24
lib/network: add tsnetwork_cork.h

gperciva:


gperciva:
This will help the potential move to `lib-platform/`.  It's not technically required for each `.c` file to have its own `.h` file, of course, but it might be a bit confusing if `network_cork.c` was in a completely different directory than `network_internal.h` if that still had the interface.

PR #390, opened 2020-05-02
Libcperciva import

gperciva:


gperciva:
This doesn't fix it, but we'll need these changes eventually.

PR #392, opened 2020-05-17
Allow --dump-config to be a mode (like --print-stats)

gperciva:


gperciva:
For example, this would have been useful in the most recent thread on tarsnap-users: http://mail.tarsnap.com/tarsnap-users/msg01580.html

tar/bsdtar.c
```diff
@@ -1120,6 +1125,7 @@ main(int argc, char **argv)
 	case OPTION_RECOVER_DELETE:
 		missingkey = crypto_keys_missing(CRYPTO_KEYMASK_AUTH_DELETE);
 		break;
+	case OPTION_DUMP_CONFIG:
```
cperciva, line 1128: Missing `break` here?

tar/bsdtar.c
```diff
@@ -1075,7 +1080,7 @@ main(int argc, char **argv)
 
 	/* Make sure we have whatever keys we're going to need. */
 	if (bsdtar->have_keys == 0) {
-		if (!bsdtar->option_dryrun) {
+		if ((!bsdtar->option_dryrun) && (!bsdtar->option_dump_config)) {
```
cperciva, line 1083: Doesn't this mean that `tarsnap --dump-config` will generate (fake) keys?  I'm not sure I understand what you're doing here.

tar/bsdtar.c
gperciva: That's deliberate because I wanted the fallthrough into the `OPTION_PRINT_STATS`, but I should have modified the comment accordingly.

gperciva: Wait, no, you're right that a `break` is better here.

cperciva:
Looks plausible.  Please rebase.

gperciva:
Rebased.  I'm having second thoughts about whether `--dump-config` should try to load the keyfile, though.

- check keyfile: the main goal of `--dump-config` is to help confused users, most likely because the package installed a "default" config file.  So it makes sense to check as many things as possible.

- don't check keyfile: the name of the command is `--dump-config`, not `--please-help-me-i-am-lost`.  If the user's problem is that they can't read the keyfile, that error message is already pretty clear, and it states the exact filename it tried to read.

... I'm now leaning more towards "don't check", in which case I'd need to move the "bail" part 10 lines higher.  What's your take?

cperciva:
I wondered about that.  Yeah, I think I'd move it up to the same place as VERIFY_CONFIG.

For that matter, now I'm wondering if we should set the mode to DUMP_CONFIG at all, or just reuse the VERIFY_CONFIG mode...

gperciva:
Rebased again, to use the `--verify-config` mode (if suitable).

PR #393, opened 2020-05-18
libcperciva import

gperciva:


gperciva:
This imports the readpass "try to open /dev/tty; if that fails, bail" functionality.  It does not include `readpass_file`; that will be a later PR.

PR #395, opened 2020-05-19
Scryptenc partial import

gperciva:


gperciva:
This imports a few changes to `lib/scryptenc`.  I didn't include the `SCRYPT_` error codes in this PR, since I think that's a big enough change that it should go in its own.  Also, that would touch non-`lib/scryptenc` code, so again it's better to keep that separate.

PR #396, opened 2020-05-20
libcperciva: import poll-based events_network.c

gperciva:


gperciva:
This brings tarsnap in line [1] with the version of events_network.c in

> spiped 2017-02-09
> https://github.com/Tarsnap/spiped/commit/8c513a5b3e464002d04644b444e814af6edc4306

That commit was chosen because:
- the switch to poll-based network events was completed.
- no other files were modified.
- events_network.c in spiped wasn't changed for the next 2.5 years;
  and even then, the later modifications were only to comments and
  adding atexit(events_network_shutdown()).

We've had two spiped releases since this change:
> spiped 1.6.0 2017-04-29
 > spiped 1.6.1 2020-02-21
, as well as three years' worth of kivaloo development, so there's good
evidence that these changes are solid.

[1] there's one slight difference between tarsnap's events_network.c and
    the one in spiped at the specified time: the tarsnap version
    contains 78b6c17453e30495034486d773a043be662b9cd0 from 2019-06-21,
    which adds a missing colon in a function comment.

PR #397, opened 2020-05-20
Scryptenc partial import

gperciva:


gperciva:
This takes us up to just before the `SCRYPT_` error defines.

PR #398, opened 2020-05-21
Scryptenc sync SCRYPT_ error codes

gperciva:


gperciva:
This is the final set of syncing `lib/scryptenc/`.  It imports the `SCRYPT_` errors and modifies `lib/keyfile/keyfile.c` accordingly.

PR #399, opened 2020-05-21
Libcperciva import event cleanup

gperciva:


gperciva:
In spiped, we waited a month and a half between these commits and actually removing `events_shutdown`.  I think the reason was to make it easier to revert changes if necessary.

Do you want to have a similar delay here, or should I remove `events_shutdown()` from `lib/network/tsnetwork_select.c` right now?

cperciva:
Just go ahead and remove `events_shutdown` now.

gperciva:
Done, ready for merge.

cperciva:
Looks like you removed the call to `events_shutdown` but didn't actually remove `events_shutdown`?

gperciva:
Yes, because we haven't removed it from libcperciva (yet).

gperciva:
https://github.com/Tarsnap/libcperciva/blob/master/events/events.c#L245

cperciva:
Oh right, I forgot about that.

PR #401, opened 2020-05-21
Fix solaris build

gperciva:


gperciva:
This depends on #400; I'm just uploading it now so you can see where it's headed.

This fix does not appear to be necessary for scrypt; I'm guessing that since scrypt has fewer `configure` checks than tarsnap, it's not hitting whatever limit that causes it to fail with the tarsnap `configure` script.

PR #405, opened 2020-05-27
sync memlimit RLIMIT_DATA

gperciva:


gperciva:
This is the last of the `lib-platform/` syncing, and these commits come directly from scrypt.  If you want to merge as-is, go ahead.  Or I could rewrite the history if, for example, you wanted to put the "refactor" commit first.

A few caveats:
- the "don't use RLIMIT_DATA if we HAVE_MMAP" commit is required on NetBSD, but (by itself) breaks things on scrypt.  That was shown in the official scrypt 1.3.0 release.
- the "USE_RLIMIT_DATA on OpenBSD" appears to fix things, but currently this has only been verified by me running OpenBSD 6.6 inside virtualbox.

I think that these commits are simple enough, and the commit messages verbose enough, that it's plausible to merge them now.  But if you want to be cautious (say, until another scrypt release), that's also cool.

PR #407, opened 2020-06-03
tools: add check_man.py to check man-page & completions

gperciva:


gperciva:
I originally put this in `release-tools/`, but then I remembered https://github.com/Tarsnap/tarsnap/pull/80 so I dumped it in `tools/` instead.

The current script only checks that all man-page options are in a completion file.  Future work will include an `--edit-bash` option, which wipes out the existing options from a completion file and recreates them from a man-page.  That would catch any duplicate options in the completion file, and generally give a better test (i.e. if the file is unchanged, then we know it's good).  That code is already working, but I'm quibbling over the exact formatting of the bash completion file, so I figured that it would be nice to get the `--check-file` option in first, particularly because that can apply to zsh as well.

PR #410, opened 2020-06-15
import openbsd fix

gperciva:


gperciva:
This is the final update to allow the client to compile with gcc 4.2.1.

PR #411, opened 2020-06-15
Update bash completions

gperciva:


tools/check_man.py
```diff
@@ -182,11 +184,15 @@ def parse_cmdline():
                         help="man-page in mdoc format")
     parser.add_argument("-c", "--check-file",
                         help="check that all options are in the file")
+    parser.add_argument("--edit-bash",
```
cperciva, line 187: Should "edit" be "regenerate"?

gperciva: Hmm.  It certainly doesn't generate it from scratch, but I suppose that "regenerate" could work.

gperciva: ... or maybe "update", since that's what I ended up calling the branch?

gperciva:
Updated

PR #414, opened 2020-07-02
Libcperciva import

gperciva:


gperciva:
After this one, the only libcperciva import left [1] is the big "mpool to a stack" change.  That occurred on 2017-09-22, so after the "mpool for network cookies" which was on 2017-05-31.

[1] the only one left to get to current libcperciva master.

PR #416, opened 2020-07-06
Libcperciva import getopt goto

gperciva:


gperciva:
I'd forgotten that we used this in tarsnap, but it's used for keygen, keymgmt, keyregen, and recrypt.

PR #417, opened 2020-07-07
Libcperciva import mpool as stack

gperciva:


gperciva:
These are the final commits to bring the client up to date with current libcperciva.

We've had one spiped release that includes this change:
> spiped 1.6.1 2020-02-21
, as well as 18 months of kivaloo development.

PR #420, opened 2020-07-24
Libcperciva import

gperciva:


gperciva:
This includes the "fix numeric comparison for valgrind tests" commit, rebased into the "check valgrind version" commit.

PR #422, opened 2020-07-28
Don't close stdin when cleaning up after a --dry-run

gperciva:


gperciva:
This one was so much fun!

tar/multitape/multitape_write.c
```diff
@@ -416,6 +416,9 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 		goto err0;
 	memset(d, 0, sizeof(struct multitape_write_internal));
 
+	/* Initialize lockfd as "not locked". */
+	d->lockfd = -1;
```
gperciva, line 420: I did consider omitting this part and changing the "check if ok to close" to:

```
if (d->lockfd > 0)
    close(d->lockfd);
```

but then I figured that if stdin was closed somewhere else (due to a new bug), then `multitape_lock(cachedir)` could return 0, which would then be interpreted as "not locked".  It seemed safer to go with -1, which is definitely not a real descriptor.

tar/multitape/multitape_write.c
```diff
@@ -528,7 +531,8 @@ writetape_open(uint64_t machinenum, const char * cachedir,
 err5:
 	storage_write_free(d->S);
 err4:
-	close(d->lockfd);
+	if (d->lockfd >= 0)
```
cperciva, line 534: `!= -1`.  There's only one value which we're using to indicate "cache directory is not locked".  Other negative values would be invalid but we shouldn't hide bugs like that.

gperciva:
Rebased.

PR #425, opened 2020-08-31
build: move macros to m4/

gperciva:


gperciva:
Same as https://github.com/Tarsnap/scrypt/pull/272, but tarsnap has more macros than scrypt.

cperciva:
Did you want to update .gitignore?

gperciva:
Good question, but I don't think it's necessary.  I mean, empirically there's no extra files in `m4/` as a result of running `autoreconf -i`.

In the case of scrypt, we have:
```
$ ls m4/
check_libcperciva_posix.m4 ltsugar.m4
check_posix_sh.m4          ltversion.m4
libtool.m4                 lt~obsolete.m4
ltoptions.m4
```

So that's 5 extra files in `m4`.  4 of them explicitly say that they're part of libtool in their comments, while the remaining one `ltoptions.m4` mentions libtool in the help string for `--enable-shared`, and has `LIBTOOL` sprinkled throughout the file..

Since we're not using libtool in tarsnap, I believe that we shouldn't expect any extra files in `m4`.  Now, it's possible that a later version of autotools will decide "hey, let's dump extra files to `AC_CONFIG_MACRO_DIR`", but if that happens I'd be interested in knowing about it, and it's easy to add to `.gitignore` at that time.

cperciva:
Got it.  I wasn't sure what the issue was with m4 which resulted in it needing to be added in scrypt.

PR #427, opened 2020-09-05
import scryptenc 1: add scrytenc_params to public API

gperciva:


gperciva:
First PR for importing from scrypt. This one corresponds to https://github.com/Tarsnap/scrypt/pull/250.

PR #431, opened 2020-09-08
Lib scryptenc 3: return explicit params in scryptenc

gperciva:


gperciva:
The tarsnap client does not set explicit parameters, so we won't see `SCRYPT_EPARAM` in `keyfile.c`.

PR #432, opened 2020-09-08
Lib scryptenc 4

gperciva:


gperciva:
I'm less confident about this one.  The change to `scryptenc.c` arose from
https://github.com/Tarsnap/scrypt/pull/263

I think my worry comes down to:
- `static int scryptenc_setup()` is not documented as only using `force` if it's given explicit parameters.  (maybe that's ok, because this is an internal function)
- `int scryptenc_buf()` is not documented as only using `force` if it's given explicit parameters.  (Should I update the API comment in scrypt?)
- we currently do not use explicit params in tarsnap, but I can't guarantee that we never will.

cperciva:
I thought the whole point of these patches was to allow for explicit parameters?  Or were you merging this code to tarsnap in order to keep it in sync with scrypt but not intending to have any change in tarsnap's behaviour?

gperciva:
AFAIK, the whole point is to keep it in sync with scrypt.  At this point, I wouldn't hazard a guess as to whether explicit scrypt parameter would be a good idea in the tarsnap client.

cperciva:
Explicit parameters aren't something I'd *recommend* for scrypt, but it's apparently something people want, and there's at least some plausible use cases.  Since this would just be a matter of adding them to keygen and keymgmt, I don't see any harm in exposing that functionality.

gperciva:
I updated the API comments relating to `${force}` in https://github.com/Tarsnap/scrypt/pull/276, and made a note to add explicit params to keygen and keymgmt.

gperciva:
At the moment, this does not enable any ability to force using scryptenc parameters.  When we add explicit parameters to keygen and keymgmt, I imagine that we'll want to add `--force-resources` as well, but I think it makes sense for this PR to focus on importing code rather than adding new features.

PR #437, opened 2020-09-27
add snap packaging script

thefenriswolf:
Snap builds can be automated with github actions and canonical's [snapcraft.io](https://snapcraft.io/build) service.
This script adds the ability to do do this.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,93 @@
+name: tarsnap
+base: core20
+version: '1.0.39'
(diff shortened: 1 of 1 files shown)


gperciva:
Thanks for the interest!  As far as I understand it, snaps have the disadvantages of static linking: if there's a security flaw in the underlying libraries (such as libssl), updating the system library won't fix it in a snap.  We didn't accept previous PR for snaps in tarsnap (https://github.com/Tarsnap/tarsnap/pull/284) for that reason.

PR #440, opened 2020-10-10
Fix strict compile with clang-10

gperciva:


gperciva:
Yes, the only problems arose in `tsnetwork_bwlimit.c`.

PR #443, opened 2020-10-26
Fix gcc nitpicks

gperciva:


gperciva:
The first commit (the change to `crypto_scrypt_smix_sse2.c`) was imported from scrypt.

PR #447, opened 2020-11-12
Build fixes for OSX

gperciva:


gperciva:
... and yes, that function was added after 1.0.39.  Namely, in 62af7672384063cdc304f3e3f8984297db0da17d (2019-05-29).

cperciva:
Is this still blocked on another PR?

gperciva:
It's now ready for merge, although I'm expecting travis-ci to take an hour or more.  This is the last PR which involves travis-ci.  (You can see that it worked in https://github.com/Tarsnap/tarsnap/pull/448, which I combined these fixes with the "switch to github actions" PR.)

Granted, this delay is purely to satisfy an obsessive "each branch should be address single topic", and I'm considering "build on OSX" as separate from "switch to github actions and test on OSX".  From the standpoint of "ending up with good code", we could skip this PR entirely and just do #448.

The more I think about it, the more I lean towards cancelling this PR and going with #448 instead.

PR #448, opened 2020-11-12
Switch from Travis-CI to GitHub Actions

gperciva:


gperciva:
BTW, this PR shows up as "tests not completed" because github is still expecting a travis-ci test.  I'll switch that over once #447 is merged, or when we give up on that one and aim for this #448 instead.

gperciva:
I cancelled #447; we'll fix the OSX build and switch to actions at once in this PR.

PR #453, opened 2020-12-23
Valgrind freebsd

gperciva:


gperciva:
This works for the local tests.  It doesn't allow us to run valgrind on freebsd when uploading to the server, because that has a different code path in `getaddrinfo()` than merely looking up `localhost:80`.

To address that (no pun intended), I'll need to revise `potential-memleaks` and the test framework to only call certain functions if the tester deliberately wants to use functionality which requires an internet connection.  That's not hard to do, but it belongs in a separate PR.

PR #455, opened 2020-12-29
Libcperciva import aesctr cipherblock

gperciva:


gperciva:
Sometimes you want to postpone importing certain libcperciva changes to tarsnap until they've been tested in an scrypt and/or spiped release; that might apply to these.

PR #458, opened 2020-12-31
Libcperciva import

gperciva:


libcperciva/cpusupport/Build/cpusupport-HWCAP-GETAUXVAL.c
```diff
@@ -0,0 +1,11 @@
+#include <sys/auxv.h>
```
cperciva, line 1: Is there a reason to have this .c file, given that tarsnap doesn't use it?

gperciva: Oops, sorry.  I had a mental note to remove it from the non-kivaloo projects.

gperciva: ... well, actually my mental note was about not including cpusupport-ARM-CRC32.c, and I was successful at that.  I just neglected to apply that to HWCAP-GETAUXVAL.c as well.

gperciva:
Updated to not include cpusupport-HWCAP-GETAUXVAL.c.

PR #466, opened 2021-02-10
Libcperciva import

gperciva:


gperciva:
We still need to specify CPUSUPPORT CFLAGS in `Makefile.am` for now.

PR #467, opened 2021-02-12
Import -vv from libarchive

gperciva:


gperciva:
The libarchive PR was mostly https://github.com/libarchive/libarchive/pull/630, while Tim's commit was earlier in 2015.

I opted not to include the man-page change in that PR, because the grammar wasn't quite correct, and the current "addition -v options will provide additional detail" is still accurate.

PR #469, opened 2021-02-16
Add major/minor headers in tar/util.c

AlexAltea:
Otherwise I get the following in Ubuntu 20.04.2 LTS.

```
/usr/bin/ld: tar/tarsnap-util.o: in function `list_item_verbose':
util.c:(.text+0x198b): undefined reference to `minor'
/usr/bin/ld: util.c:(.text+0x19a6): undefined reference to `major'
```

PR #470, opened 2021-02-17
Add major/minor headers to tar/util.c

gperciva:


gperciva:
This is a rebased version of #469.

It's also getting the usual compile & test checks, whereas the previous PR didn't.  Maybe github actions are only triggered by a PR from members?  It's possible that's an anti-ddos technique?  I'll look into that.

PR #472, opened 2021-02-20
Libcperciva import

gperciva:


gperciva:
We should probably merge #470 before this one, so that the fix for f0e0935c0ff804d61eb63beb0d207f824a407318 is closer to the breakage.

PR #479, opened 2021-04-15
key(re)gen: sanity check the machine name

gperciva:


keygen/keygen_actual.c
```diff
@@ -19,6 +20,25 @@
...
+{
+	size_t i;
+
+	/*
+	 * Bail if there's any unprintable character in the "C" locale.  This
+	 * assumes that we haven't run setlocale(), so this process is using
+	 * the "C" locale by default (as specified in C99).
+	 */
+	for (i = 0; i < strlen(str); i++) {
+		if (!isprint(str[i]))
```
gperciva, line 34: The reason that I didn't use `isprint_l()` right away is that I'm not totally confident about library support:
- It's part of POSIX Issue 7, but `man isprint_l` doesn't produce anything, nor is it part of the `isprint` man page on FreeBSD.  I also can't find it via `grep -R isprint_l /usr/include/*` (although it appears to be ok to use in code, though??).

- On Linux, it's supported if we add `_XOPEN_SOURCE >= 700` and it's glibc 2.10+ (2009); before that, we'd have to add `_GNU_SOURCE`.  As it happens, we do both of those... but I still figure that `isprint()` is safer.

PR #481, opened 2021-05-06
fedora packaging - removed lzma-devel dependency

elagost:
lzma-devel is no longer in Fedora and the tarsnap package builds just fine without it.

### DIFF
--- a/pkg/fedora/tarsnap.spec
+++ b/pkg/fedora/tarsnap.spec
@@ -1,6 +1,6 @@
 Name:           tarsnap
 Version:        1.0.39
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Secure, efficient online backup service
 
 License:        Proprietary
@@ -12,7 +12,6 @@ BuildRequires:  bzip2-devel
 BuildRequires:  e2fsprogs-devel
 BuildRequires:  libacl-devel
 BuildRequires:  libattr-devel
-BuildRequires:  lzma-devel
 BuildRequires:  openssl-devel
 BuildRequires:  xz-devel
 BuildRequires:  zlib-devel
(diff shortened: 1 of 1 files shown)


gperciva:
Thanks for the PR, @elagost!

I think this ok, but please change the message to "lzma-devel dependency removed as it was replaced by xz-devel".

I would normally want to indicate this by using a boolean expression in the spec file, i.e.

    BuildRequires:   (xz-devel or lzma-devel)

but apparanently that capability was only added in rpm 4.13 [1]. CentOS 7 uses rpm 4.11.3, so we can't rely on that.  :(

[1] https://rpm.org/user_doc/boolean_dependencies.html

I've checked that this works with centos 7 (originally 2014-07-07, which is the oldest currently-supported version of that distro; centos 6 was EOL on 2020-11-30).  RHEL 6 is not supported any more; RHEL 7 is similar to CentOS, so it should be fine.  The oldest currently-supported version of fedora is 32 (2020-04-08), so I'm sure it also has xz-devel.

gperciva:
Oh, one more piece of reassurance about RHEL: version 6 (which is EOL'd) already supports xz-devel:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/package_manifest/chap-base-server-variant
so there's absolutely no doubt in my mind about RHEL 7+.

elagost:
Updated the wording as you mentioned. Hope that fits right.
Thanks for replying so quickly! You rock.

gperciva:
Thanks for the update!  Are you familiar with git rebase?  If not, don't worry; we can do that when we merge the PR.  If you are familiar with it, please rebase the two commits together and do a force-push to your branch.

elagost:
Not too familiar, but I believe what I've done works. Please correct me if I'm wrong.

gperciva:
Looks great!  Now I just need to figure out how to get the actions to run.

(and no, I'm not seeing the "Approve and run" box that's mentioned in https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

gperciva (approved):

cperciva:
Ok, this looks fine to me, but github won't let me merge it until checks are completed and they're not running... @gperciva can you figure this out?

gperciva:
No, apparently I can't figure it out.  :(

I've disabled the "require status checks" temporarily, so you should be able to merge it now.  I'll revisit the question of "approving status checks for first-time contributers" later.  (This is a recent change in github, so it's possible that there's still some bugs in the infrastructure.)

PR #482, opened 2021-05-21
Passphrase entry (refactor internal use only)

gperciva:


gperciva:
This is a preliminary step towards `--passphrase method:arg`.  It imports passphrase_entry.[ch] from scrypt and uses it internally in `keyfile.c`.  No program behaviour or internal APIs are changed in this PR.

PR #483, opened 2021-05-21
Passphrase entry

gperciva:


gperciva:
This is the full `--passphrase method:arg` PR.

I'm happy to split off another PR that goes up to b330b74, if you want.  That would be an intermediate step between #482 (refactor internal use only) and the current one: it does the keyfile.h API update and refactoring bsdtar.c, but would not change any program behaviour.

gperciva:
Ready for review.

This PR removes the `--passphrase-stdin` option, because it was not part of any official release (more info in the commit message of c47509a).  By contrast, we kept `-P` in scrypt because it was added in 1.2.1, and we only added `--passphrase` to scrypt in 1.3.1.

PR #484, opened 2021-05-23
Passphrase entry (refactor binaries)

gperciva:


gperciva:
This is another part of #483.  It updates the `keyfile.h` API (and adjusts binaries to match), then updates bsdtar.  There is no change of program behaviour.

PR #487, opened 2021-06-16
Continue without write cache if we can't read it

gperciva:


gperciva:
I don't think that this is important enough to list in `NEWS.md` or the release notes, but I'll raise the question in case you think it should.

cperciva:
Looks good, and I agree that it doesn't need to be in the news or release notes.

Can you fix whatever github issue is preventing me from merging this?

gperciva:
Looks like there's a "require branch to be up-to-date" option checked.  We didn't use to have this enabled... I can't see any sign that it's a new github thing.  Maybe I included it by accident the last time we had to temporarily remove protection from `master`?

I'm inclined to leave it on for now, but it's easy to disable it if you definitely don't like it.

cperciva:
Maybe I've always merged one branch at once before you uploaded a new branch before?  I'm ok with leaving that turned on for now; we can always turn it off if it continues to cause problems.

gperciva:
Nah, you've definitely approved multiple PRs in the past without me updating them.  Sometimes there's a conflict (which has always required manual attention), but this it the first time I remember seeing a non-conflict "branch is out of date".

PR #488, opened 2021-06-17
Fix valgrind with --list-archives on FreeBSD

gperciva:


gperciva:
This allows us to run

    valgrind ./tarsnap --list-archives

on FreeBSD.  No leaks detected in Linux.

PR #497, opened 2021-10-05
Nitpicks in crypto

gperciva:


lib/crypto/crypto_compat.h
```diff
@@ -11,7 +11,9 @@ int crypto_compat_RSA_valid_size(const RSA * const key);
 
 /**
  * crypto_compat_RSA_import(key, n, e, d, p, q, dmp1, dmq1, iqmp):
- * Import the given BIGNUMs into the RSA ${key}.
+ * Import the given BIGNUMs into the RSA ${key}.  If this function fails,
```
gperciva, line 14: This is the only change that I'm not completely certain about.

I mean, ultimately I think it's a good change, and I can't think of a better way to organize it.  But the API comment is a bit awkward, which makes me think that it *isn't* the best way to handle this.

cperciva: I think it's fine apart from the grammar.  "If this function fails, free any BIGNUMs which have not been imported into ${key} (but do not free ${key} itself)."  Imperative case or whatever it's called.

gperciva:
Updated.

lib/crypto/crypto_compat.c
```diff
@@ -66,31 +68,31 @@ crypto_compat_RSA_import(RSA ** key, BIGNUM * n, BIGNUM * e, BIGNUM * d,
...
+		key->dmq1 = dmq1;
+		key->iqmp = iqmp;
 	}
 #else
 	/* Do we have a public key, or private key? */
 	if (d == NULL) {
 		/* We could use d here, but using NULL makes it more clear. */
-		if (RSA_set0_key(*key, n, e, NULL) != 1)
+		if (RSA_set0_key(key, n, e, NULL) != 1)
 			goto err0;
```
cperciva, line 87: Shoudn't we be freeing `n` and `e` here?

gperciva:
Done

PR #500, opened 2021-10-29
Ensure that hexlink_read() has the correct errno

gperciva:


lib/util/hexlink.c
```diff
@@ -56,6 +57,7 @@ hexlink_read(const char * path, uint8_t * buf, size_t buflen)
 {
 	char * hexbuf;
 	ssize_t rc;
+	int saved_errno;
```
cperciva, line 60: `int saved_errno = 0;` and delete the two `= 0` lines below.

gperciva:
Rebased, ready for merge.

PR #503, opened 2021-11-20
storage_read_cache prep

gperciva:


gperciva:
Some tidbits:

- We want to reserve "storage_read_cache*" for future use.

- I find that it makes sense to have two related functions:

      storage_read_add_name_cache()
      storage_read_add_data_cache()

  Also, not part of this branch, but the idea is to also have

      storage_read_cache_add_name()
      storage_read_cache_add_data()

  I don't mind using "hash" or "file" instead of "name", but I'll note that the storage_read API uses `(char class, const uint8_t name[32])` and `uint8_t classname[33]` all over the place, so I figured that "name" was the best thing to call it.  (Or "classname", I suppose, but I don't think the "class" adds anything useful to the function name, given that `char class` is in the argument list anyway.)

- 81f4f45 shows a bunch of changes, but if you look at it with `--ignore-space-change`, we can see that the non-whitespace changes are trivial:

```
$ git show --ignore-space-change 81f4f45
...
diff --git a/tar/storage/storage_read.c b/tar/storage/storage_read.c
index 8d15152..225a397 100644
--- a/tar/storage/storage_read.c
+++ b/tar/storage/storage_read.c
@@ -549,10 +549,11 @@ callback_read_file_response(void * cookie, NETPACKET_CONNECTION * NPC,
                }
 
                /* Should we cache this data? */
+               if (sc == 0) {
                        classname[0] = C->class;
                        memcpy(&classname[1], C->name, 32);
                        if (((CF = rwhashtab_read(C->S->cache_ht, classname)) != NULL) &&
-                   (CF->inqueue != 0) && (CF->buf == NULL) && (sc == 0)) {
+                           (CF->inqueue != 0) && (CF->buf == NULL)) {
                                /* Make a copy of this buffer if we can. */
                                if ((CF->buf = malloc(C->buflen)) != NULL) {
                                        /* Copy in data and data length. */
@@ -564,6 +565,7 @@ callback_read_file_response(void * cookie, NETPACKET_CONNECTION * NPC,
                                }
                        }
                }
+       }
 
        /*
         * If the user's tarsnap account balance is negative, print a warning
```

gperciva:
I've reconsidered having the pair of:

```
storage_read_add_name_cache()
storage_read_add_data_cache()
```

We need a public API function for `storage_read_add_name_cache()` -- with potentially a different name, like `storage_read_request_to_cache()` or `storage_read_add_data_for_this_name_to_the_cache_if_you_see_it()` -- but we don't need a public function for `_add_data_cache()`.

With that in mind, I'd suggest that we name the latter function what it'll end up being inside `storage_read_cache.h`... probably `storage_read_cache_add_data()`, unless a better name comes to mind.

gperciva:
Rebased for review, updated the top comment.  PR 1/3 towards an independent `storage_read_cache.[c,h]`.

tar/storage/storage_read.c
```diff
@@ -549,18 +549,20 @@ callback_read_file_response(void * cookie, NETPACKET_CONNECTION * NPC,
...
-		    (CF->inqueue != 0) && (CF->buf == NULL) && (sc == 0)) {
-			/* Make a copy of this buffer if we can. */
-			if ((CF->buf = malloc(C->buflen)) != NULL) {
-				/* Copy in data and data length. */
-				CF->buflen = C->buflen;
-				memcpy(CF->buf, C->buf, C->buflen);
-
-				/* We've got more data cached now. */
-				C->S->cachesz += CF->buflen;
+		if (sc == 0) {
```
cperciva, line 579: How about we drop the `if (sc == 0)` and make this `case 0:` of the preceding `switch` statement?  That seems pretty self-explanatory if we successfully decrypted the data, check if we want to stuff it into the hash table.

cperciva: That would also make it possible to simplify the series of `if` conditions by replacing with a series of (commented) `if (blah) break;` statements covering all the reasons we might *not* put data into the hash table.

tar/storage/storage_read.c
```diff
@@ -249,6 +249,35 @@ storage_read_add_name_cache(STORAGE_R * S, char class, const uint8_t name[32])
 	return (-1);
 }
 
+/**
```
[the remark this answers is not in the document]
gperciva: It's going to be moved to a new file `storage_read_cache.c` file.  The planned history is:
- this PR: make it an independent function
- next PR (#505): make it accept a new `struct storage_read_cache * cache` instead of `STORAGE_R * S`
- following PR (not online yet): move all cache functions (including this one) into the new `storage_read_cache.c`

tar/storage/storage_read.c
gperciva: Yes to using `case 0:` instead of `if (sc == 0)`.

We can't do the `if (blah) break;` if this code is in a separate function (and later, separate file).  But I've rewritten that function to use `if (blah) return;`, which is the same idea.

cperciva:
Ok, LGTM.

gperciva:
Rebased; this changed the history of the last 3 commits of this PR.

PR #412, opened 2020-06-18
tests: check -c, -d with a real keyfile

gperciva:

PR #504, opened 2021-11-20
tests/04: add trivial checks for -t and -x

gperciva:


gperciva:
Add quick little tests for `-t` and `-x` to the existing set of quick little tests.  This is primarily so that we easily do a valgrind test on those code paths.

FWIW, the past (minimal) discussion about this file was #412.  I can't remember why I didn't add `-t` and `-x` to begin with, nor why you didn't ask me about it.  I have a *very* vague memory that on some operating system, anything involving openssl would produce a ton of valgrind warnings -- I even have a branch aimed at suppression those warnings -- but I've tested on freebsd 12.2/amd64, amazon linux/arm64, and linux/amd64 without seeing valgrind problems.

PR #505, opened 2021-11-21
storage_read_cache consolidate

gperciva:


gperciva:
This depends on #503.  The goal of this PR is to ensure that every `storage_read_cache_*`
function takes `(struct storage_read_cache *)` as an argument instead of `(STORAGE_R *)`; and that no non-`storage_read_cache` function accesses the cache struct's members.

tar/storage/storage_read.c
```diff
@@ -58,6 +62,42 @@ static sendpacket_callback callback_read_file_send;
 static handlepacket_callback callback_read_file_response;
 static int callback_read_file(void *, int, uint8_t *, size_t);
```
cperciva, line 64: Spurious blank line.

tar/storage/storage_read.c
```diff
@@ -653,6 +686,27 @@ callback_cache_free(void * record, void * cookie)
...
+static void
+storage_read_cache_free(struct storage_read_cache * cache)
+{
+
+	/* Behave consistently with free(NULL). */
+	if (cache == NULL)
+		return;
+
+	/* Free contents of cache. */
+	if (cache->cache_ht != NULL)
```
cperciva: Is it possible for `cache_ht` to be NULL here?  I prefer to avoid NULL checks if possible since they can mask bugs.

tar/storage/storage_read.c
gperciva: That was somewhat intended, since the `cache_free` was going to be deleted in the following PR.  But yeah, on second thought I agree that I shouldn't be adding the blank line.

tar/storage/storage_read.c
gperciva: > Is it possible for cache_ht to be NULL here?

Yes-ish.  I mean, with the current setup of `storage_read_cache_init()`, it's possible.

* `cache` is allocated
* `rwhashtab_init()` fails for some reason.  (`malloc` failing, or `crypto_entropy_read()`, or `keylength == 0`)
* we jump to `err1` in `storage_read_cache_init()`, which currently calls `storage_read_cache_free()`.

It should be safe to replace the _init function's `err1` with a `free()` instead; I've pushed a new commit to do that.

cperciva:
Much better, thanks.

gperciva:
Rebased

PR #506, opened 2021-11-23
Fixes

gperciva:


tar/chunks/chunks_directory.c
```diff
@@ -176,6 +176,13 @@ chunks_directory_read(const char * cachepath, void ** dir,
 		goto err2;
 	}
 
+	/* Make sure the directory file isn't too small (different message). */
+	if ((size_t)sb.st_size < sizeof(struct chunkstats_external)) {
```
gperciva, line 180: Arguably, this could be merge in the above

```
if ((sb.st_size < 0) ||
    ...
```

check.  I opted not to do that because the earlier check was clearly checking for "insane" sizes, and if the on-disk directory was something like 8 bytes, it's clearly an error but not an "insane" error.

gperciva:
Changed this to a draft because I meant to add more context to the final commit message.

gperciva:
Ready.

PR #507, opened 2021-11-27
tests/valgrind: pthread_atfork() leak in FreeBSD 12.2

gperciva:


gperciva:
I discovered this due to LibreSSL's arc4random4.c and arc4random_freebsd.h, but it's not their fault at all.

I don't think this matters a great deal.  POSIX notes:

    The pthread_atfork() function may be formally deprecated (for
    example, by shading it OB) in a future version of this standard.
    https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_atfork.html

while LibreSSL's arc4random_freebsd.h notes:
```
/*
 * Unfortunately, pthread_atfork() is broken on FreeBSD (at least 9 and 10) if
 * a program does not link to -lthr. Callbacks registered with pthread_atfork()
 * appear to fail silently. So, it is not always possible to detect a PID
 * wraparound.
 */
```

Still, it would be nice to clear up the valgrind warnings about the 40 bytes of "still rechable" memory.

gperciva:
Revised to only compile&run on FreeBSD.

tests/valgrind/potential-memleaks.c
```diff
@@ -9,6 +9,15 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if defined(__FreeBSD__)
+/*
+ * We only want to enable this check on FreeBSD, because that OS can use
+ * pthread_atfork() without explicitly linking to pthread.
```
gperciva, line 15: I assume you know how FreeBSD does this.

cperciva: Sure, pthread_atfork is part of libc. ;-)

Or at least a stub function is.  For same same reasons as we're (umm, once I get around to reviewing it) going to have "maybelock".

PR #508, opened 2021-11-29
Improve support for LibreSSL 2.7.0

gperciva:


gperciva:
Fixes #319.

gperciva:
First commit: with LibreSSL 2.7.0+, we need to explicitly call `EVP_cleanup()`.

(this is irrespective of whether we use the 1.0.1 API or 1.1.0 API)

Annoying, this is contrary to the documented behaviour of LibreSSL. I've explained our need to use `EVP_cleanup()` in the file, but here's a few additional quotes and comments:

LibreSSL documents that `EVP_cleanup` is part of a set of which:

>    These functions are deprecated.  It is never useful for any application program to call any of them explicitly.
> https://man.openbsd.org/OpenBSD-6.3/EVP_cleanup.3

[OpenBSD 6.3 (2018-04-15) used LibreSSL 2.7.2; that was the first OpenBSD release which used LibreSSL 2.7.0 or higher.  The same message appears in LibreSSL 3.4.2 (2021-11-25).]

OpenSSL says:
> In versions prior to 1.1.0 EVP_cleanup() removed all ciphers and digests from the table.  It no longer has any effect in OpenSSL 1.1.0.
> https://www.openssl.org/docs/man1.1.1/man3/EVP_cleanup.html

I've heard that the upcoming LibreSSL 3.5.0 will change the API a bit. Hopefully that will include making an explicit `EVP_cleanup()` call unnecessary, in which case we only need to modify the preprocessor lines inside the `#ifdef LIBRESSL_VERSION` section.

PR #510, opened 2022-01-02
Fix --newer

gperciva:


gperciva:
This addresses three issues, two of which were raised in 2016:

1) the man-page should say that `--newer` works in c, x, and t modes.
https://mail.tarsnap.com/tarsnap-users/msg01343.html

2) `tarsnap -xf blah --newer "20170101"` exits with an error if it encounters a file it should skip.
  https://mail.tarsnap.com/tarsnap-users/msg01344.html

  This bug was inherited from libarchive, and is fixed by the 3 back-ported commits.

3) `get_date()` returns -1 to indicate error; we should check that.

gperciva:
The test failure is due to brew on openssl, and openssl 3.0's "deprecated" stuff.

It appears that github is giving us openssl 3.0, even though we **specifically** are trying to link to openssl@1.1.

gperciva:
Known issue with github actions: https://github.com/actions/virtual-environments/issues/4719

gperciva:
In more details: this broke (at least) 21 days ago; they merged a fix 19 days ago, but the updated images are not yet in use.  The issue is tagged "awaiting-deployment", and they said that they'll notify when it's out.

gperciva:
Previous updates of github's macOS 10.15 image occurred on:

11-20
11-26
12-07
12-12
12-20

There's various workarounds we could use, including disabling the "require status checks" from the github interface, but it looks like they usually update their image approximately every week, other than the holidays.  So I'm hopeful that they'll deploy their fix any day now.

PR #511, opened 2022-01-03
Fix misc

gperciva:


gperciva:
(I cancelled the CI tests because they'd fail due to the same github actions bug that's blocking #510.)

PR #514, opened 2022-01-15
siginfo: clarify that we're counting all filesystem entries

gperciva:


gperciva:
Previous behaviour:

```
$ ls dir-with-1-and-2/
1 2
$ ./tarsnap -c --dry-run --progress-bytes 1k dir-with-1-and-2/
Processed 2 files, 2560 bytes
adding dir-with-1-and-2/2 (0 / 1234 bytes)
Processed 3 files, 4608 bytes
$ 
```

That looks wrong, since there's only 2 files.  New behaviour:

```
$ ./tarsnap -c --dry-run --progress-bytes 1k dir-with-1-and-2/
Processed 2 files & directories, 2560 bytes
adding dir-with-1-and-2/2 (0 / 1234 bytes)
Processed 3 files & directories, 4608 bytes
$ 
```

cperciva:
Hmm, tar files also contain symlinks.  And device nodes.  And FIFOs.  Admittedly device nodes and FIFOs aren't very useful in archives, but symlinks certainly are.

Maybe a generic "entries" would be better?

gperciva:
Rebased to say "entries" instead, ready for merge.

PR #515, opened 2022-01-15
Fix --progress-bytes vs -v newlines

gperciva:


gperciva:
There's some ugliness if somebody uses both `--progress-bytes` and `-v`.  (I thought that we'd discussed this when we added `--progress-bytes` in 2018 and decided that we didn't expect people to use both options, but I can't find any record of this in the old PRs.)

For example,
```
$ ./tarsnap --progress-bytes 1k -v -xf 1234
x 1
x 2
Processed 1 files, 2560 bytes
extracting 2
x 3
Processed 2 files, 4608 bytes
extracting 3
x 4
Processed 3 files, 6656 bytes
extracting 4

Processed 4 files, 9216 bytes$ echo "that doesn't look good"
that doesn't look good
```

The complication here is that libarchive doesn't add a `\n` to the end of "verbose" lines.  Instead, all other "printing" functions which appear after printing a verbose line start with a `\n`.  I haven't tracked down why this happens, but I'm scared of Chesterton's fence.

This solution, which adds an `(int lastline)` to `siginfo_printifo()`, seems like the safest option. Libarchive eliminated `siginfo.c` shortly after we diverged from libarchive 2.7.0, so this isn't adding any significant divergence from libarchive.

I have a different branch where I switched verbose to include a `\n`.  That leads to "cleaner" code in that it doesn't require any API change, and we can remove all of the
```
if (bsdtar->verbose)
        fprintf(stderr, "\n");

...

if (!bsdtar->verbose)
        fprintf(stderr, "\n");
```

lines in `siginfo.c`.  But that also leads to many changes in `read.c` and `write.c` to undo the special "no `\n` after a verbose line" workarounds.  I'm concerned that the alternate solution would lead to more corner-case bugs.

cperciva:
> The complication here is that libarchive doesn't add a \n to the end of "verbose" lines. Instead, all other "printing" functions which appear after printing a verbose line start with a \n. I haven't tracked down why this happens, but I'm scared of Chesterton's fence.

This is so we can have `a filename: warning message goes here`.

gperciva:
Rebased to call it `finalmsg` instead of `lastline`, and slightly tweaked the logic in `siginfo_printinfo`.  I'm not fond of the overall logic in that function, but since the "verbose" lines don't have a final newline, we have to play games somewhere.

I think that the most useful part of this PR is the tests.

gperciva:
Ok, ready again.

I thought that there might be an oddity if `--progress-bytes` was large enough not to apply (i.e. if we're looking at 100k but only adding two 1k files).  But there was no problem after adding such a test, and I can't manually generate an error in this case.

PR #517, opened 2022-01-20
Add "descriptions" to check-man.py

gperciva:


gperciva:
This is the step 1/2 towards a replacement for #226.  This step:

* adds a `misc/describe-options.txt`, which contains short (1-line) descriptions of tarsnap options.

* adds python code to parse that file, and match those descriptions against the options in the man-page.

As of this PR, `misc/describe-options.txt` is only used as a sanity check, but the following PR will use that as part of the zsh completion file generation.

gperciva:
I did consider trying to generate short descriptions of options by parsing the man-page directly.  However, the first sentence of options

* was not always less than 1 line of text
* did not always explain the option.  (That's not a problem in the man-page, since there were subsequent sentences)

After spending far too long on working on parsing the man-page, and rewriting parts of the man-page so that the first sentence of each option would serve as a 1-line short description, I realized that it would be less work to have an independent file.

misc/describe-options.txt
```diff
@@ -0,0 +1,109 @@
...
+--configfile		add file to the list of configuration files to be read
+--creationtime		manually specify a creation time for the archive
+--csv-file		write statistics in CSV format to a file
+--disk-pause		often pause for ARG ms while creating an archive
+--dry-run		do not really create an archive; just simulate doing so
+--dump-config		print all config lines
+--exclude		do not process certain files or directories
+--force-resources	force the decryption of a passphrase-encrypted keyfile
+--fsck			perform some integrity checks and reconstruct cache	MODE
+--fast-read		stop after the first entry which matches the ARG	DUP-q
```
cperciva, line 21: Ordering?

misc/describe-options.txt
```diff
@@ -0,0 +1,109 @@
...
+--maxbw			stop if more than ARG bytes of upstream bw is used
+--maxbw-rate		limit bandwidth to ARG bytes per second
+--maxbw-rate-down	limit download bandwidth to ARG bytes per second
+--maxbw-rate-up		limit upload bandwidth to ARG bytes per second
+--newer			only add files and dirs newer than the given date
+--newer-mtime		like --newer, but it uses mtime instead of ctime
+--newer-mtime-than	like --newer-than, but it uses mtime instead of ctime
+--newer-than		only include files and dirs newer than the given file
+--no-aggressive-networking	ignore any aggressive-networking option
+--no-config-exclude	ignore any exclude option specified
```
cperciva, line 42: Not sure why "ignore any ... option specified in a configuration file" is sometimes shortened to "... option" and sometimes to "... option specified"?

gperciva:
Rebased with `--fast-read` in the correct place, and replacing "option specified" with "option".

PR #518, opened 2022-01-20
Add zsh completion

gperciva:


gperciva:
This builds on #517 to serve as step 2/2 to replace #226.

I expect to make a few more nitpick fixes tomorrow, but I thought it was worth putting it up here so that you could see where #517 was moving towards; in particular, how the `misc/describe-options.txt` were going to be used.

gperciva:
You might be wondering about the new arguments `--write-argparse` and `--write-zsh` in contrast to `--update-bash`.

`--update-bash` modifies an existing bash completion file, whereas the `--write-*` options creates a new one from scratch.

If I was writing the bash stuff from scratch, I'd use `shtab` to generate a bash completion file from scratch.  (And I'm not ruling that out in the future.)  But for now, I figured that it wasn't worth touching the bash stuff.

gperciva:
Ready for review.

Issue #522, opened 2022-01-25
Build fails on NetBSD


gperciva:
A rather convoluted chain of headers leads to `"sysendian.h"` beind included before `<sys/endian.h>` on NetBSD 8.1.  That's a bit old (2019-05-31), but tarsnap 1.0.39 works on it, so this would be a build failure on a platform which previously worked.

Our `"sysendian.h"` avoids namespace collisions with BSD if it's included after `<sys/endian.h>`, but that method doesn't help if `<sys/endian.h>` is included first.

How this arises:

* `crypto_file.c` includes `crypto_internal.h`.
* that includes `<openssl/aes.h>`.
* on NetBSD 8.1, `/usr/include/openssl/aes.h` contains

  ```
  #ifdef __NetBSD__
  #include <sys/types.h>
  #endif
  ```

* `/usr/include/sys/types.h` includes `<machine/endian.h>`
* `/usr/include/machine/endian.h` includes `<sys/endian.h>`

I think this problem arose in
2019-05-07 lib/crypto: remove unnecessary "bsdtar_platform.h"
1d4e106295bcf1453c46b28eba514aea1a1561fc

**Solution 1)** at any rate, if we slap a
```
#ifdef __NetBSD__
#include "bsdtar_platform.h"
#endif
```

at the top of `crypto_file.c` and `crypto_session.c`, then it compiles fine on NetBSD.  (I haven't tracked down exactly which included header from `bsdtar_platform.h` includes `<sys/endian.h>`, but there's a bunch of files included, including `archive.h` and `archive_entry.h`, so I'm sure it's in there somewhere.)

**Solution 2)** would be to add

```
#ifdef __NetBSD__
#include "sys/endian.h"
#endif
```

to the top of our `sysendian.h`.  That has the advantage of completely taking care of any potential problems from the order of headers, and being a NetBSD-specific solution which won't affect any other platforms.  It has the disadvantage of being platform-specific code in libcperciva, which we don't want.

**Solution 3)** would be to shove the platform-specific

```
#ifdef __NetBSD__
/* Make sure this is included before our "sysendian.h". */
#include "sys/endian.h"
#endif
```

to `crypto_file.c` and `crypto_session.c`.  That way, libcperciva remains pure.

gperciva:
I poked around a little bit in the OpenSSL and NetBSD github repositories.  I can't see that
```
#ifdef __NetBSD__
#include <sys/types.h>
#endif
```
[...]

PR #524, opened 2022-01-26
Fixes for NetBSD

gperciva:


gperciva:
I'm always paranoid about touching any `lib/crypto/` stuff, so I separated all the `RSA -> void` changes into function-by-function for easier reviewing.  I can merge them all together if you want.

gperciva:
Most of these are related to #522.  The final two commits address other problems that cropped up later in the build.

lib/crypto/crypto_internal.h
```diff
@@ -3,7 +3,6 @@
 
 #include <stdint.h>
 
-#include <openssl/aes.h>
```
cperciva, line 6: This header file existed long before tarsnap 1.0.12; that's just when it was moved to its current location.

In tarsnap 0.1.5 it contained

```
struct crypto_aes_key {
        AES_KEY key;                    /* Expanded key. */
        uint64_t nonce;                 /* Nonce, used only for encryption. */
        uint8_t key_encrypted[256];     /* AES key encrypted with encr_pub. */
};
```

which is why it needed `<openssl/aes.h>`.

Can you change the commit message to just say that it was needed in previous versions but is no longer necessary?

cperciva:
Yes, please squash the `RSA *` -> `void *` changes together.

gperciva:
Rebased, ready for merge.

PR #525, opened 2022-01-26
tests/06-progress-output: trim exact number of bytes

gperciva:


cperciva:
> The precise number of bytes can vary from system to system (e.g.,
> whether it supports extended pax attributes or not).

I would say "The precise number of bytes can vary between operating systems and between filesystems (e.g., if there is filesystem metadata which libarchive decides needs an extended pax header)."

Otherwise looks fine.

gperciva:
Rebased, ready for merge.

PR #526, opened 2022-01-26
Fix for GNU Hurd

gperciva:


gperciva:
Depends on #525.  I suspect that you won't want to merge this particular PR, but I thought it was worth putting it on github.

The only thing which stops git master from compiling on GNU Hurd is the lack of `PATH_MAX`.  A quick hack would be that if we don't have `PATH_MAX`, we could use `_POSIX_PATH_MAX` (which is 256) instead.

A careful reading of the POSIX page for `<limits.h>` reveals:

> Pathname Variable Values
>
>    ...
>
>    A definition of one of the symbolic constants in the following list
>    shall be omitted from the <limits.h> header on specific
>    implementations where the corresponding value is equal to or greater
>    than the stated minimum, but where the value can vary depending on
>    the file to which it is applied.  The actual value supported for a
>    specific pathname shall be provided by the pathconf() function.
>
>    https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

As stated, the recommended solution is to get the maximum length via `pathconf()` dynamically.  However, that would require more intrusive rewriting of some libarchive code.

gperciva:
Another option would be to add that to `tar/bsdtar_platform.h` and `libarchive/archive_platform.h`, instead of individual `.c` files.

gperciva:
Actually, the "most correct" solution would be to add it to `configure`.  That turned out to be easier than I expected.

gperciva:
Hmm, continued testing is running into some weird issues on Hurd.  Let's revisit this later.

PR #528, opened 2022-01-27
Nitpicks from clang-scan

gperciva:


gperciva:
These are the nitpicks in tarsnap code, not including the false positives (discussed in #527).

There's a lot more nitpicks in libarchive code.  I'll prepare a separate branch for (some of?) those, although my first guess is that we won't want to merge it due to diverting from libarchive.

gperciva:
Actually, I think some of the libarchive changes are worth doing, so I'll add them here when they're ready.

PR #529, opened 2022-01-28
Fixes for clang scan with libarchive imports

gperciva:


gperciva:
This builds on #528 to address some warnings from clang-scan, by backporting a few libarchives commits.

PR #530, opened 2022-01-29
tree.c: avoid theoretically dereferencing NULL

gperciva:


gperciva:
This diverges from libarchive in a specific and limited way.

There is no such sanity check in modern libarchive, so I doubt that the code can ever reach `tree_pop()` with `(t->stack == NULL)`.  But that's not obvious from the code, so we might as well be safe.

cperciva:
Looks to me like all code paths leading to `tree_pop` have already dereferenced `t->stack`, so this is just a case of "telling the compiler what it should already be able to deduce".

gperciva:
ok.

PR #531, opened 2022-02-01
A few fixes to deb pkg infrastructure

gperciva:


gperciva:
This is mostly just the `pkg/DEBIAN.md` file, which was out of date and misled me.

Biggest change: a reminder that if we have a new tarsnap version number, there's no need to run `release-tools/mkdebsource.sh`.  That script is only needed if we're building a second (or later) deb package from the same `tarsnap-autoconf-X.Y.Z.tgz` tarball (i.e. where we changed stuff in `pkg/debian/` but nothing else).

release-tools/mkdebsource.sh
```diff
@@ -18,29 +18,20 @@ if ! command -v debuild >/dev/null 2>&1; then
...
-# version number.
-PRERELEASE=`echo $Z | tr -d [:digit:]`
-W=`echo $Z | tr -d [:alpha:]`
-if [ -z "$PRERELEASE" ]; then
-	# Not a prerelease
-	TAR_ORIG_VERSION="$X.$Y.$W"
-else
-	TAR_ORIG_VERSION="$X.$Y.$W~a"
+# Handle a "X.Y.Z.A" pre-release version number.
+if [ "$A" != ".tgz" ]
```
cperciva, line 27: Wouldn't `A` end up being `tgz` rather than `.tgz` in the case of a non-prerelease package?

gperciva:
Yes.  Thanks, rebased.

PR #532, opened 2022-02-01
Update packages for pre-release

gperciva:


gperciva:
Builds on #531.  AFAIK, this is the final PR before 1.0.39.99.

I didn't update `pkg/archlinux/PKGBUILD`, since I'm not familiar with their packaging system.  But now that I'm looking at it, I see that it's set to version 1.0.29 (!).  OTOH, the arch linux package which we link to from our website is https://archlinux.org/packages/community/x86_64/tarsnap/ and that's 1.0.39.  So I feel safe in ignoring that file in our repository for now.

PR #534, opened 2022-02-04
1.0.39.99 fixes

cperciva:
Fix a couple nits which I noticed while sending out the 1.0.39.99 announcement.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -2,7 +2,7 @@ Unless specified otherwise in individual files or directories, the
 contents of this package is covered by the following copyright, license,
 and disclaimer:
 
-Copyright 2006 - 2020 Tarsnap Backup Inc.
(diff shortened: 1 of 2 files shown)


gperciva:
Whoops.

Also `libcperciva/COPYRIGHT`?

gperciva:
... or maybe not bother with `libcperciva/COPYRIGHT`?  There's a bunch of other copyright statements that we're not changing.

Obviously we don't want to alter `libarchive` stuff, but we've certainly changed files like `lib/scryptenc/scryptenc.c` since 2009.  I'm guessing that it's not worth updating those?

cperciva:
We should keep `libcperciva/COPYRIGHT` in sync with libcperciva.  Same goes for bits from `scrypt`.

We should also bump the copyright dates in those other trees at some point, but no rush; you can do a sweep after 1.0.40 is out?

PR #535, opened 2022-02-05
tests: fix == in sh

gperciva:


gperciva:
In vaguely-related news, I added a 1.0.39.99 tag.

PR #538, opened 2022-02-12
Bump version to 1.0.40-head

gperciva:


gperciva:
I also added a tag for 1.0.40.

PR #544, opened 2022-06-12
Add storage_read_cache.[ch]

gperciva:


cperciva:
LGTM aside from rebasing and adding the not-found case to the API for storage_read_cache_find.

PR #545, opened 2022-07-01
tsnetwork_bwlimit.c: add TSNETWORK_PRINT_SPEED_SECONDS

gperciva:


gperciva:
To use, compile with

    -DTSNETWORK_PRINT_SPEED_SECONDS=x

where x is the minimum duration between updates, and run with `--debug-network-stats`.

The message is in the form:

    TSNETWORK_PRINT_SPEED_SECONDS  seconds  speed_in  speed_out

(with the separators being tabs instead of two spaces)

lib/network/tsnetwork_bwlimit.c
```diff
@@ -120,6 +134,40 @@ pokeone(struct bwlimit * l, double t, int op)
 	return (-1);
 }
 
+#ifdef TSNETWORK_PRINT_SPEED_SECONDS
+static void
+print_speed(struct timeval tnow)
+{
+	static struct timeval torig;
```
gperciva, line 141: I found it easier to reason about these two static variables when they were inside the `#ifdef`'d function.  You use static function variables infrequently (for example, in `getserveraddr()` in `netproto_connect.c`), but in this case the alternative was to add another top-level `#ifdef` section.

PR #548, opened 2022-08-07
build: improve autoconf compatibility

gperciva:


gperciva:
Fixes deprecated macros in autoconf 2.70+.  First two commits imported from https://github.com/Tarsnap/scrypt/pull/339.

PR #550, opened 2022-08-15
Build: move more configure.ac into m4/

gperciva:


gperciva:
This allow us to share more build-system code between scrypt and tarsnap. No timing difference with autoconf 2.69 or 2.71.

cperciva:
LGTM, needs to be rebased for some reason?

gperciva:
Rebased, ready for merge.

A while ago, we enabled the "branch must be up-to-date before merging" option.  This check is in addition to the "must resolve any conflicts".  We decided to try it as an experiment.

I've disabled that option now.  I mean, on average we don't have many PRs in tarsnap, but when we do there's often 2 or 3 small ones we want to do at the same time.  And since it's so infrequent, we keep on forgetting about the "must be up-to-date" thing.

PR #562, opened 2023-01-27
Libcperciva import

gperciva:


gperciva:
The "spellcheck" commit has a few extra fixes, not worth having as a separate commit.

One mistake it doesn't fix is:

    _AUTO_GENERATED = "# AUTOMATCALLY GENERATED by `shtab`"

because that refers to a typo in `shtab 1.5.5` and below.  That will be addressed in a separate PR.

gperciva:
Build failure is because it's picking up openssl 3.0 instead of 1.1 on macOS, even though we explicitly tell it `-I/usr/local/opt/openssl@1.1/include`

gperciva:
This includes #563.

PR #564, opened 2023-02-09
Rename getfstype -> ts_getfstype

gperciva:


gperciva:
There's a (non-standard) function called getfstype() on many platforms. Unless we examine the headers of tar/write.c and notice that it contained

    #include "getfstype.h"

instead of

    #include <fstab.h>

it would be easy to get confused about which function we're calling.

gperciva:
Renaming files is something we normally avoid with git, and it's unlikely that anybody else would waste 30 minutes double-checking whether `getfstype()` modifies `errno` upon error [1].

However, in this case we recently moved the file from `lib/util/getfstype` to `lib-platform/getfstype`, so this doesn't add any additional difficulty in seeing the file's history.

[1] in case anybody is curious, the non-standard `getfstype()` that's available on most platforms is *not* documented as modifying `errno`, whereas the `getfstype()` that's in `lib-platform/util/getfstype.c` does set `errno` by virtue of using `strdup()`.

PR #566, opened 2023-03-06
write.c: include padding in final progress message

gperciva:


gperciva:
At the moment, `tarsnap -c --totals` reports a number 1024 bytes higher than `tarsnap -c --progress-bytes 1G`.  That's because `--progress-bytes` didn't update the size of the archive after `archive_write_close`, thereby missing the final 2 blocks of 512 padding bytes.

gperciva:
(CI failure due to github actions is being a pain again)

PR #567, opened 2023-03-06
Actions: move from Ubuntu 18.04 to 20.04

gperciva:


gperciva:
Just like https://github.com/Tarsnap/scrypt/pull/352, but I forgot to import it to tarsnap at the time.

Issue #494, opened 2021-09-27
Look for tarsnap.conf in ~/.config/tarsnap/


heygarrett:
From the [tarsnap.conf man page](https://www.tarsnap.com/man-tarsnap.conf.5.html):

> Options may be specified in /usr/local/etc/tarsnap.conf and/or ~/.tarsnaprc

I'm trying to purge $HOME of configuration files (and I've done `chmod -w $HOME` to prevent future messiness), so I'm using ~/.config as much as I can, per the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Looking for a tarsnap.conf in ~/.config/tarsnap/ would prevent littering at home and would put tarsnap amongst [the growing number of projects supporting the spec](https://wiki.archlinux.org/title/XDG_Base_Directory#Support).

gperciva:
This has been added in #569, and will be part of the next official tarsnap release.

PR #569, opened 2023-04-24
Read from $XDG_CONFIG_HOME/tarsnap/tarsnap.conf

gperciva:


gperciva:
As suggested in #494.

PR #576, opened 2023-06-20
Add --list-archives --null

gperciva:

PR #577, opened 2023-07-01
statstape_printlist(): refactor

gperciva:


gperciva:
I think I'd like to rebase #576 on top of this refactoring.

Adding a separate `statstape_printlist_item()` is overkill for merely allowing `--list-archives --null`, but there's some future work that would be greatly simplified by this.  (See upcoming private email in an hour.)

If you like that plan and we're going to have `statstape_printlist_item()` eventually, I think the git history will make more sense to add it sooner rather than later.

gperciva:
This is prep for #576; once this refactoring is done, it's trivial to allow `--list-archives --null`.

You probably won't want to merge this by itself, preferring to do #576 all at once.  But I figured there was no harm in separating the refactoring from the new feature, just in case you wanted to merge this one first.

PR #578, opened 2023-08-08
Libcperciva import

gperciva:


tar/storage/storage_write.c
```diff
@@ -302,7 +302,7 @@ int
 storage_write_file(STORAGE_W * S, uint8_t * buf, size_t len,
     char class, const uint8_t name[32])
 {
-	struct write_file_internal  * C;
+	struct write_file_internal * C;
```
gperciva, line 305: This style fix isn't a libcperciva import, but it's trivial and it's not worth having a separate PR just for this.

PR #582, opened 2023-09-17
Misc

gperciva:


lib/network/tsnetwork_buf.c
```diff
@@ -35,7 +38,7 @@ struct network_buf_cookie {
 	size_t bufpos;
 	struct timeval timeout;
 	struct timeval timeout_max;
-	ssize_t (* sendrecv)(int, void *, size_t, int);
```
gperciva, line 38: I'm fairly confident that this is a good change, but I suspect that my commit message is incorrect.

clang-16 definitely refuses to compile the previous version, claiming that it's a function cast between incompatible function types.  As such, replacing the function pointer with an int seems like a good change.

However, c99 6.7.5.3 para 15 contains:

> (In the determination of type compatibility and of a composite type, each parameter declared with function or array type is taken as having the adjusted type and each parameter declared with qualiﬁed type
is taken as having the unqualiﬁed version of its declared type.)

If I'm understanding that correctly, then

```
const f1(char *);
const f2(const char *);
```
should be compatible; in which case, `send()` and `recv()` should be compatible.

cperciva: I think the code as written was correct; as you note, the extra `const` should not affect function type compatibility.  But it's probably a good change anyway since using a flag will probably generate better code than indirecting via a function pointer.

Maybe just change the commit message to a vaguer "some compilers object to using a function pointer this way".

gperciva:
Rebased, updated commit message.

Since the commit messages are primarily for us, I went with more detail than you proposed.  If the question came up again, I didn't want to be digging through github discussions to find that info.

If you really don't like it, I'll prune it down to what you wrote last night.

New commit message:

Some compilers object to using a function pointer which can point to
functions which have different const qualifiers, i.e.

    ssize_t send(int s, const void *msg, size_t len, int flags);
    ssize_t recv(int s, void *buf, size_t len, int flags);

This is allowed:

    (In the determination of type compatibility and of a composite type,
    each parameter declared with function or array type is taken as
    having the adjusted type and each parameter declared with qualiﬁed
    type is taken as having the unqualiﬁed version of its declared
    type.)

    - C99, 6.7.5.3 para 15

but that's a somewhat esoteric detail.

Also, using a flag might generate better code than indirecting via a
function pointer.

Reported by:	clang-16

gperciva:
I received an email pointing out flaws in my previous commit message.

Rebased the PR.  There's now a rather vague commit message:

Using a flag will probably generate better code than indirecting via a function pointer, and some compilers object to using a function pointer to refer to both:

    ssize_t send(int s, const void *msg, size_t len, int flags);
    ssize_t recv(int s, void *buf, size_t len, int flags);

PR #584, opened 2023-09-19
Fix typo

shawwn:
Just a minor typo fix.


gperciva:
Thanks!  Could you please rebase this PR on top of current master (in order to get #585, which should allow github actions to run).

gperciva:
Thanks for the fix!  This was merged as part of #594.

PR #585, opened 2023-09-19
Actions: don't ignore if '/' in branch name

gperciva:


gperciva:
When I included the 'any branch without a "/"' in the initial commit in https://github.com/Tarsnap/tarsnap/commit/afeb2a5be6e630ebb1dd9de7f6649a3c519b54d7, I was thinking of branch names like "`experimental/whatever`" or
"`gperciva/messing-around`", and assumed that if we really wanted to run
Actions on such branches, we could trigger them manually.

However,
1) we've had a few users submitting PRs over the years with '/' in the
   branch names.
2) as far as I can tell, we can't trigger actions if the branch name is
   disallowed in the ".yml" file.

PR #588, opened 2023-10-02
Libcperciva import and trivial fixes

gperciva:


gperciva:
Also a few style fixes for function pointers.

PR #589, opened 2023-10-11
Nitpicks

gperciva:


lib/crypto/crypto_internal.h
```diff
@@ -93,7 +93,8 @@ int crypto_file_init_keys(void);
 
 /**
  * crypto_MGF1(seed, seedlen, buf, buflen):
- * The MGF1 mask generation function, as specified in RFC 3447.
+ * The MGF1 mask generation function, as specified in RFC 3447 section B.2.1,
```
gperciva, line 96: I've had this change laying around for over a year, but it never seemed important enough to include it in a PR.

I forget why I was looking up the details of it, but when I found the right section, it occurred to me that being a bit more specific in the API comment would have saved me some time.
https://datatracker.ietf.org/doc/html/rfc3447#appendix-B.2.1

PR #593, opened 2023-10-18
Exclude dead "leading surrogate" UTF8 code

gperciva:


gperciva:
It's logically impossible for

    wc >= 0xDC00 && wc <= 0xDBFF

to be true.

I'm not at all convinced that this is the best solution; I'm just getting it "on the radar".  Unfortunately, libarchive did a bunch of refactoring in https://github.com/libarchive/libarchive/commit/d7cec4d7e9564b25bd238963ca46263d8784d2ea and I haven't yet figured out how they fixed this.

PR #594, opened 2023-10-18
Fix macos openssl and typo

gperciva:


gperciva:
This is an updated version of #583 and #584.

I tweaked the OpenSSL detection a bit.  @shawwn, does it work on Apple silicon?

gperciva:
Confirmed that this fixes it on Apple silicon as well.  Please review and merge.

PR #595, opened 2023-10-20
Elastic array wrap around bug

dorjoy03:
Commits picked up from https://github.com/Tarsnap/tarsnap/pull/592. 

Two things:
1. There was a unsigned wrap around bug in the elasticarray resize code which could cause the buffer to become smaller than the given nsize. I don't really know if it could happen in practice or what problems could be caused in case it happens.
2. Added a comment explaining that another wrap around is actually handled as it was unclear from just the code.

### DIFF
--- a/libcperciva/datastruct/elasticarray.c
+++ b/libcperciva/datastruct/elasticarray.c
@@ -30,7 +30,7 @@ resize(struct elasticarray * EA, size_t nsize)
 		nalloc = EA->alloc * 2;
 		if (nalloc < nsize)
 			nalloc = nsize;
-	} else if (EA->alloc > nsize * 4) {
+	} else if (EA->alloc / 4 > nsize) {
 		/* We need to shrink the buffer. */
 		nalloc = nsize * 2;
 	} else {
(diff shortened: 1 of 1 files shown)


libcperciva/datastruct/elasticarray.c
```diff
@@ -28,6 +28,12 @@ resize(struct elasticarray * EA, size_t nsize)
 	if (EA->alloc < nsize) {
 		/* We need to enlarge the buffer. */
 		nalloc = EA->alloc * 2;
+		/*
```
gperciva, line 31: Sorry, one more nitpick: please add a blank line before this comment.

gperciva: Actually, on second thought, never mind.

I don't know why github isn't running the Actions tests on this PR.  Based on the documentation I've found, it should be working now.  :(

I'm going to make a new PR (from my own account) with these commits, plus that "blank line" nitpick.  That's the only guaranteed way I know of getting github Actions to work, and that'll let the review go through faster.

gperciva:
This is continued in #596, due to github Actions not running on this PR.

PR #597, opened 2023-10-20
Actions: allow manual triggering

gperciva:


gperciva:
Lastest attempt to get this working.

... nope, I can already see that it's not doing what I want.  :|

.github/workflows/compile.yml
```diff
@@ -1,11 +1,13 @@
...
-# Run whenever we push to any branch other than [coverity-scan].
+# Run whenever we push to any branch, or if there's a PR on any branch.
 on:
   push:
     branches:
     - '**'
-    - '!coverity-scan'
+  pull_request:
+    branches:
+    - '**'
```
dorjoy03, line 10: This is only my guess. I think this is kind of the correct way. I am guessing in this PR the actions ran twice because of both 'push' and 'pull_request' events. For example, I looked at curl/.github/workflows/linux.yml. There I see 'on push branch main' and 'on pull_request branch main'. I think that means pushing to main branch of the repository runs the actions and on pull request against the main branch. That would prevent running them twice if you open a pull request. To actually test if this fixes the outsiders' issue, someone would need to submit a PR with these yml changes in their PR (assuming tarsnap's master branch doesn't have these yet).

dorjoy03: Edit: by main I meant master. If you do the same in this PR i.e., 'on push branch master' and 'on pull_request branch master' like the curl example, the actions should not run twice.

gperciva: Yes, I found the part of the docs:

> If multiple triggering events for your workflow occur at the same time, multiple workflow runs will be triggered.
> https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-multiple-events

I was hopeful when I read that, because I kind-of expected the next sentence to be "... if you only want one workflow to be trigger, then XYZ".

I have the nagging feeling that I'm missing something obvious.  I mean, what I was isn't complicated:
- if there's any push to the `Tarsnap/tarsnap` respository, run the tests.
- if there's a PR but no workflow yet (because it came from a third party), run the tests.

I'm baffled that the default assumption seems to be that running the same workflow multiple times is desired.  I mean, unless the tests are non-deterministic (which would be a mistake!), there's no point testing a certain git hash more than once.  If git hash abce1234 fails, one workflow is enough.  If it succeeds, then running more than one workflow is just a waste of time & energy & resources.

Hmm, maybe `workflow_dispatch` is the best option.  At least that lets us run a workflow manually.

gperciva:
Latest attempt.  I think it's ready to merge?

There's two caveats:
- I'm only 99% confident that the YAML syntax allows for the hanging `workflow_dispatch:` with nothing below it.

  But if it wasn't allowed, I was expecting to see a syntax error in the Actions for this PR; but instead, they're working.  Also, there's an example of this in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events, namely with "page_build:"

- I can't be certain that this will actually enable manual Actions, because:
  > To trigger the workflow_dispatch event, your workflow must be in the default branch.
  > https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#configuring-a-workflow-to-run-manually

So I think we might as well merge this.

PR #599, opened 2023-11-09
Build nitpicks

gperciva:


gperciva:
This includes `NEWS.md`, `README.md`, and `STYLE`, which is arguably a more significant change than a "nitpick".

Back in 2019 https://github.com/Tarsnap/scrypt/pull/210, we decided to distribute `STYLE` in scrypt.  As with that PR, I'm not saying that we need to ship these files, but I'm also not saying that it's pointless to ship them.

I agree with not including the whole build apparatus (`configure.ac`, `aclocal.m4`, etc).  But I think it would be worth including `NEWS.md`, and if we're doing that, we might as well slap the other two files in there -- if we don't, it kind-of looks like a sloppy mistake, rather than a deliberate omission.

PR #600, opened 2023-12-11
Add --list-archives-hashes

gperciva:


gperciva:
There's a question of grammar here: `--list-archives-hashes` is wrong; it should end with `archive-hashes`.  However, I didn't like this combination of options:

```
--list-archives
--list-archive-hashes
```

Despite the grammar, I think it's more clear if we used:

```
--list-archives
--list-archives-hashes
```

Or maybe `--list-archives-hash`?  At least that doesn't have two plurals.

Another possibility would be to add an entirely separate option, such as perhaps `--hashes` or `--use-hashes`.  Then the GUI could do
```
--list-archives --use-hashes
```

That would have the benefit of not adding a new command mode.

PR #601, opened 2023-12-11
Add --hashes for --list-archives

gperciva:


gperciva:
I like `--hashes` much better.  How does this look, instead of #600?

NEWS.md
```diff
@@ -11,6 +11,10 @@
   name with a null character (like `find -print0`).  If one or more -v
   arguments are specified, multiple null characters are used to separate
   fields; see the man page for details.
+- tarsnap now accepts --hashes, which causes --list-archives to prints archive
+  hashes.  If one or more -v arguments are specified, it will print other
```
cperciva, line 15: Maybe "archive name hashes" to avoid making people think they can use this to verify the contents of an archive?

cperciva:
> I like `--hashes` much better. How does this look, instead of #600?

Much better.

NEWS.md
gperciva: How about "hashes of archive names"?

tar/bsdtar.c
```diff
@@ -1653,12 +1653,6 @@ dooption(struct bsdtar *bsdtar, const char * conf_opt,
 		if (exclude(bsdtar, conf_arg))
 			bsdtar_errc(bsdtar, 1, 0,
 			    "Couldn't exclude %s", conf_arg);
-	} else if (strcmp(conf_opt, "hashes") == 0) {
```
gperciva: When I was reviewing the earlier code, I wondered about adding a `no-hashes` option in case somebody wanted to cancel `hashes` in their config file.  But then I decided that there was no compelling case to support `hashes` in config files, so I've made it solely a command-line option.

I also decided not to bother with `bsdtar->option_hashes_set`, since we normally don't check for "doubly set" options for command-line-only options.  This means that `tarsnap --list-archives --hashes --hashes` is accepted, but I figure that's not a problem.

(We also accept weirdness such as `tarsnap -x --chroot --chroot` or `tarsnap -c --dry-run --dry-run`.)

gperciva:
Updated with REBASE commits.

NEWS.md
cperciva: Sure.

tar/bsdtar.c
cperciva: Yeah, I considered that but agree there's no reason to support `hashes` in configuration files and thus no need for `--no-hashes`.

cperciva:
LGTM, please rebase and then I'll merge.

gperciva:
Rebased, ready for merge.

PR #602, opened 2023-12-12
Allow -f with --list-archives --hashes

gperciva:


tar/bsdtar.c
```diff
@@ -989,10 +990,13 @@ main(int argc, char **argv)
 	}
 
 	/*
-	 * The -f option doesn't make sense for --list-archives, --fsck,
-	 * --fsck-prune, or --nuke.
+	 * The -f option doesn't make sense for --fsck, --fsck-prune, or
+	 * --nuke.  It can be used with --list-archives --hash, but not
+	 * --list-archives on its own; sanity-checking that detail is
+	 * done in tarsnap_mode_list_archives().
```
gperciva, line 996: I'm not wild about this comment, but trying to sanity-check the whole "-f with --list-archives --hashes but not --list-archives without --hashes" here lead to some really ugly code.  It looked a lot cleaner for the code to allow "--list-archives -f" here, then check the details later.

PR #603, opened 2023-12-19
Nitpicks

gperciva:


lib/network/tsnetwork_buf.c
```diff
@@ -43,7 +43,7 @@ struct network_buf_cookie {
 	int flags;
 };
 
-static int callback_buf(void * cookie, int timedout);
+static int callback_buf(void * cookie, int status);
```
gperciva, line 46: This function is defined on line 57 (just a few lines down) with that arg being `status`, and that name seems a better match for how that arg is being used.

cperciva: Yeah, `timedout` is a historical name for that variable.  I switched to `status` a long time ago but I guess I missed this one.

PR #606, opened 2023-12-26
Exit code 2 if server-side state was modified (and there was an error)

gperciva:


gperciva:
This changed slightly from the last time you saw it: I'm now doing the modifications to `storage_transaction_commitfromcheckpoint` in the second commit, not the first.  (i.e. it's part of the boilerplate passing `storage_modified` around.)

I also amended the NEWS and man page as requested.

PR #612, opened 2024-03-19
Check empty cachedir for -d, --print-stats, --recover

gperciva:


gperciva:
I'm not entirely sold on this.  On one hand, it does the desired:
```
$ ./tarsnap --cachedir empty-cachedir/ -d -f tiny
tarsnap: Cache directory must already exist for -d; check --cachedir and/or run --fsck
```

On the other hand, it seems messy to `#include "chunks.h"` in `bsdtar.c`.

On the gripping hand, `glue.c` doesn't seem much better.  And while I could certainly add functions to multitape etc. to pass the function call down the layers, that seems like overkill.  Another option would be to move `chunks_directory_exists()` to a separate `chunks_really_public.h`, then only include that one in `bsdtar.c`.

So this seems to me like the least bad way to implement this, but I don't like it.

PR #623, opened 2025-01-18
Shorten descriptions for zsh autocomplete

gperciva:


gperciva:
In zsh, you can hit `<TAB>` to see possible completions and brief descriptions.  For example,

```
$ tarsnap --c
--configfile  -- add to the list of config files to be read
--csv-file    -- write statistics in CSV format to a file
```
(the options change based on the mode; so we'd only see `--chroot` if we'd previous added a `-x` to the options)

If we just type `tarsnap -<TAB>`, then experimentally is seems that descriptions can be at most 43 chars before they get truncated.

Reviewing note: the first commit changes whitespace, so looking at the entire changes are rather hard to read.  To see how the options were shortened, I recommend looking at [dece4e2](https://github.com/Tarsnap/tarsnap/pull/623/commits/dece4e2860be7443d24139b99aa01c0b35b26aa5) directly.

misc/describe-options.txt
```diff
@@ -63,52 +63,52 @@
...
---nodump			honor the nodump file flag by skipping this file
---noisy-warnings		be verbose when warning about network glitches
+--nodump			do not read files with the nodump file flag
+--noisy-warnings		verbose when warning about network glitches
 --normalmem			ignore any lowmem or verylowmem option
 --nuke				delete all of the archives stored	MODE
---null				filenames or patterns are separated by null characters
---null-output			archive names in output are separated by null characters
---numeric-owner			ignore symbolic user and group names when restoring
+--null				names in input are separated by NULs
```
gperciva, line 70: This is probably the most contentious change.  A full description would be
```
filenames, archive names, and patterns in input are separated by null characters
```
but that's much more than 43 chars.

I figured that since this is only used for reminders when people using zsh manually hit <TAB>, condensing "filenames / archive names / patterns" to simply "names" was ok.  I did try to think of other ways to describe it, but after including "in input" (vs. "in output") and "separated" and "NULs", there weren't many characters left before hitting the 43-char limit.

cperciva: How about `input separated by NULs instead of newlines`?

gperciva: I have a slight qualm that somebody could think that meant data (i.e. file contents), rather than metadata.

cperciva: Meh, if someone is that confused about how tar works they're probably not using tarsnap?  And it's the same as the --null option to xargs, which also refers to using NUL separators instead of newlines... I think it's sufficiently unambiguous.

gperciva: Ok.

However, there's now a problem with `--null-output`.  Your text was very cleverly exactly 43 chars, but `output` has 1 more letter than `input`.  How about:
```
--null                  input separated by NULs instead of \\n
--null-output           output separated by NULs instead of \\n
```

(on the command-line that's printed as `\n`)

gperciva: ... OTOH, that might raise alarms in terms of `\r\n` line endings.

How about
```
input split by NULs instead of newlines
output split by NULs instead of newlines
```

cperciva: Hey, you told me I had 43 characters to work with!

Yeah, "split" isn't quite as good as "separated" but it will do.

PR #625, opened 2025-01-19
Support --null-output with -t and -x (without -v)

gperciva:


gperciva:
Adding `--null-output` for `-v` and `-vv` is a bit more involved.  I think it's useful to get this much "out of the way" first; but if you'd rather wait to review the entire thing at once, that's certainly a valid option.

gperciva:
... on second thought, the subject message is a bit inaccurate.

This prints NULs at the ends of filenames.  As such,
- `tarsnap -xf blah` is unaffected
- `tarsnap -xvf blah --null-output` is NUL-separated
- `tarsnap -xvvf blah --null-output` is NUL-separated by lines, but has spaces within lines (i.e. "`x -rw-r--r--  0...filename^@`")
- `tarsnap -tf blah` is NUL-separated
- `tarsnap -tvf` and `-tvvf` is NUL-separated by lines, but has spaces within lines.

Maybe it's best to withdraw this PR and resubmit when I'm satisfied with the intra-line NULs as well.

cperciva:
Yeah let's do this once you've got the rest ready too.

PR #627, opened 2025-01-21
Add & use dirutil_fsync()

gperciva:


tar/ccache/ccache_write.c
```diff
@@ -266,6 +267,10 @@ ccache_write(CCACHE * cache, const char * path)
 		goto err2;
 	}
 
+	/* Finish writing the file. */
```
gperciva, line 270: There was a report by a MacOS user that their chunkification cache wasn't being written correctly -- they always turned their computer off after running tarsnap, and on one occasion, the file's date was correct before turning their computer off, but when they booted it again the next day, the file's date was older than it had been previously.

That said, the MacOS man page for `fsync()` is not encouraging:
> Specifically, if the drive loses power or the OS crashes, the application may find that only some or none of their data was written.
> ...
> This is not a theoretical edge case.  This scenario is easily reproduced with real world workloads and drive power failures.
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html

The man page goes on to suggest `fcntl()` with the `F_FULLFSYNC` flag.  That's not a standard-compliant flag, of course, but we could add that to `dirutil_fsync()` quite easily once we've abstracted the "sync this fp" part.

cperciva:
Can you write an "#ifdef F_FULLFSYNC ... #endif" to add to this?  Or rather, can you test one (since it's easy to write something which will only fail to compile if F_FULLSYNC is defined)?

For something like spiped I would prefer to stick to POSIX, but tarsnap has paying customers so I'm fine with adding non-portable stuff if it helps users.

gperciva:
Can do, although I've now heard back from the user, and they never tested the `fcntl` branch -- they found that waiting 10 seconds before turning off their computer was enough to avoid this problem. :shrug:

But there's certainly no harm in adding `fcntl` just in case.

PR #628, opened 2025-01-22
Clarify warning if server-state was modified

gperciva:


gperciva:
I don't think this is suitable as-is, but I'm having difficulty moving forward without guidance.

In #606, we made tarsnap give an exit code of 2 if the server state was modified.  At that time, you commented
> ... and actually, while we're doing it, let's change the "Error exit delayed from previous errors" message since it's becoming a bit of a FAQ.  We should be clearer about reporting that an archive was created and the "previous errors" might simply mean that it didn't contain all of the data it was supposed to contain.

If you want, we could give custom warnings depending on the mode:
- if `-c`, warn that the created archive might not contain all the requested data
- if `-d`, warn that some archives might not have been deleted (wrong archive name?)
- if xyz, warn that blah blah

The downside of this approach is that we'd have to enumerate all the ways that tarsnap could fail after modifying the server -- my favourite part of #606 is that by setting it in `storage_transaction_commit()`, we didn't have to worry about forgetting some edge failure case.

cperciva:
Meh.  I guess it was a FAQ back then but I haven't seen it recently.  If we do change this I agree that different warnings based on the mode would make sense, but for now let's put this on ice until/unless users start hitting it again.

gperciva:
Hmm.  I think it might still be useful to give *some* visual indication that the server-state was modified.  I mean, if somebody is in the habit of `echo $?` then they'd notice, but I often forget to check the exit code until after I've run 1 or 2 other commands -- at which point it's much too late.

PR #629, opened 2025-01-28
Standardize on fprintf() rather than printf()

gperciva:


gperciva:
Not important, but this struck me as weird.

The first one was added in:
>    2008-04-25 Tarsnap 1.0.0
>    41f175dbc884f04e36a98811188e1100bb4dbfae

while the second one was added in:

>    2013-07-13 Tarsnap 1.0.34
>    8c34f3eb5480ce1a01b50f87ca1df606096b16b5

Other than those two cases, the only uses of printf() in c files in the source tree is recrypt/recrypt.c and tests/valgrind/potential-memleaks.c.

PR #631, opened 2025-01-29
build: reformat lists

gperciva:


gperciva:
- One item per line
- If there's more than one item in the list, start on the line after the assignment "=".

This has no effect on the build system, but it makes future changes to lists much easier to see.

Removing the empty lines before the LIBTARSNAP_A+= lines clarifies the file, in that there's a single "paragraph" for each library, rather than one multi-line "paragraph" followed by a one-line "paragraph".

PR #632, opened 2025-01-29
Add apisupport and use for low-level aes & rsa

gperciva:


gperciva:
This includes #631, which I recommend merging before looking at this too seriously.

... huh, clang-10 has some linking problem?  Hmm.

Makefile.am
```diff
@@ -378,9 +397,22 @@ nodist_lib_libtarsnap_arm_sha256_a_SOURCES=			\
 		cpusupport-config.h
 lib_libtarsnap_arm_sha256_a_CPPFLAGS=$(lib_libtarsnap_a_CPPFLAGS)
 lib_libtarsnap_arm_sha256_a_CFLAGS=`. ./cpusupport-config.h; echo $${CFLAGS_ARM_SHA256}`
-
 LIBTARSNAP_A+=	lib/libtarsnap_arm_sha256.a
 
+lib_libtarsnap_crypto_a_SOURCES=				\
```
gperciva, line 402: I'm really not certain what to call this library.

I originally went with `libtarsnap_crypto_aes` (as we did in scrypt https://github.com/Tarsnap/scrypt/pull/386), but then I discovered that 4 files in `lib/crypto/` also needed the same compiler flag.

I can see a few possibilities:
1) put only the required files into a generic `libtarsnap_crypto.a` (current).
2) put everything in `lib/crypto/` into a `libtarsnap_crypto.a`.  That would mean that some files might be compiled with fewer warnings than is strictly necessary, but that's a rather theoretical concern.
3) add a `libtarsnap_crypto_aes.a` which only contains `crypto_aes` (as with scrypt); and add a `libtarsnap_openssl_compat.a` (or some name like that) which has the other ones.

gperciva:
This needs more revision; I'll bring it back later.

PR #635, opened 2025-02-09
build: apisupport, cpusupport_detect, crypto_aes, rsa

gperciva:


gperciva:
These are imported from scrypt.

PR #636, opened 2025-02-09
build: add apisupport low_level_rsa

gperciva:


gperciva:
I'm not very happy about this PR; opening for discussison.

OpenSSL 3.0 marked many RSA functions as "deprecated", so this adds an `apisupport-LIBCRYPTO-LOW_LEVEL_RSA` (just like we did for AES).

The 4 files which need that flag are now being compiled into a new libtarsnap_rsa.a.

However, `crypto_keys_init()` (which is one of those 4 files) requires `crypto_file_init_keys()`.  In order to resolve symbols, I've included 3 other files in this library.  (This is what I'm most unhappy about.)

Also, this isn't an air-tight dependency resolution... `crypto_file` need numerous other symbols such as crypto_aes.  I'm guessing that those are resolved because libtarsnap.a already needed them, so the linker already included them in the binary.

I see a few possibilities:
1. Current PR: have a library for those 4 files, plus the 3 others that are required.
2. Have a library for only those 4 files, then add libraries to the linker command-line multiple times.  (However, this needs more than 2 instances of `libtarsnap.a`.)
3. Give up on making openssl happy with our use of RSA.  If somebody tries to compile tarsnap and it fails due to deprecation reports, then we tell them to manually add `-Wno-deprecated-declarations` to their CFLAGS.  (This is the current situation.)

cperciva:
Can we just yank `crypto_keys_init` into a new `lib/crypto/crypto_keys_init.c` file?

cperciva:
Something like https://github.com/Tarsnap/tarsnap/pull/637

gperciva:
... huh, I even wrote "refactor `lib/crypto/` as option 4, but then deleted it because I figured there was no way you'd go for it.  I thought it would need a much more invasive refactoring.

That solves this, and allows `libtarsnap_rsa.a` to contain only the 4 files that really needs to.

I'm not positive about the name, but that's the only remaining quibble.  Another option would be `libtarsnap_libcrypto_rsa`, or even `libtarsnap_crypto_rsa.a`.

(I think the latter is too-easily confused with `crypto_rsa.c`, but evidently I shouldn't self-censor ideas.)

gperciva:
Working, rebased, ready for review.

I still have some minor qualms about the name of the library, but that's it.

PR #637, opened 2025-02-10
Refactor crypto_keys_init

cperciva:
Move bits which don't deal directly with keys into a new file crypto_keys_init.c in order to avoid linkage pollution issues.


gperciva:
Tiny rebase recommended, but this looks good.

cperciva:
Thanks, I fixuped your patch into mine.

PR #638, opened 2025-02-11
pkg: relicense pkg/debian/ as BSD-2-clause

gperciva:


cperciva:
Hmm, I think `pkg/debian/copyright` should include a Tarsnap Backup Inc. copyright line as well?

cperciva:
(That's a separate issue from this change, I just noticed it when reviewing the diff and we should probably fix it while we're touching that file.)

gperciva:
Hmm, there's a bunch of cperciva-only copyrights.  For example, `keygen/tarsnap-keygen.1-mdoc` is currently:
```
.\" Copyright 2007, 2009 Colin Percival
```

My first instinct is to copy that line and replace your name with "Tarsnap Backup Inc.".  But that might be legally murky since the company was incorporated in 2011.

... hmm, the top-level `COPYRIGHT` has the company copyright starting in 2006.

Anyway, that file received small updates in 2015 and 2016, and a trivial "new sentence, new line" update in 2020.

I see a few options:
1) add the years 2015 and 2016 to your copyright, and a new line for company copyright which has only 2015 and 2016.  (And do similar things for all other such files.)
2) add a new copyright date of 2025, and a new line for the company copyright.
3) rely on the top-level `COPYRIGHT`, and remove file-specific copyright statements unless there's other statements in that file (for example, `tar/tree.c`).

cperciva:
Just change them to Tarsnap Backup Inc.  The company can hold copyrights which were created before the company existed -- a copyright statement is not an authorship statement.

gperciva:
Ok.  And the dates?  `tar/tarsnap.1-mdoc.in` has had significant changes over the years, but it's noted as 2007-2009.

cperciva:
Go ahead and update it to 2007 - YYYY for whatever the appropriate year is.  (I think 2025?)

cperciva:
Just put in a range.  The Berne convention makes copyright statements legally unnecessary anyway; the FSF thing is just their house style.

cperciva:
FWIW a bunch of copyright statements get bumped in FreeBSD as soon as the new year arrives, often before any substantive commits.

gperciva:
Rebased and ready for review.

pkg/COPYRIGHT
```diff
@@ -0,0 +1,28 @@
+The files in this directory and the "pkg/fedora" directory are distributed
```
gperciva, line 1: This paragraph is technically unnecessary, since it's already covered by the top-level `COPYING` file.  But I figured that when it came to copyright and licensing, it was better to be overly-clear.

pkg/debian/copyright
```diff
@@ -28,3 +28,27 @@ License: Commercial
  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
  OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  SUCH DAMAGE.
+
+Files: pkg/debian/*
+Copyright: 2009 - 2025 Tarsnap Backup Inc.
```
gperciva, line 33: The tarsnap client code is, in general, copyright 2006 onwards.  But `pkg/debian/` was only added in 2009, so I figured that was a better start date.

PR #639, opened 2025-02-11
Fix order of calloc() args

gperciva:


libarchive/archive_read_support_compression_bzip2.c
```diff
@@ -183,7 +183,7 @@ bzip2_reader_init(struct archive_read_filter *self)
 	self->code = ARCHIVE_COMPRESSION_BZIP2;
 	self->name = "bzip2";
 
-	state = (struct private_data *)calloc(sizeof(*state), 1);
+	state = (struct private_data *)calloc(1, sizeof(*state));
```
gperciva, line 186: Yes, that cast is a silly C++ism, but the commit I mentioned only fixed the order of calloc, so I left the cast unchanged.

PR #644, opened 2025-02-18
chunks_stats_print*: print fields separately

gperciva:


gperciva:
This refactoring is mainly to support adding `--print-stats --null-output`, so we could wait until that's all finished.  But I think it also makes the code easier to read, so it's worth considering it on its own.

tar/chunks/chunks_stats_internal.c
```diff
@@ -63,6 +63,21 @@ chunks_stats_addstats(struct chunkstats * to, struct chunkstats * from)
 	to->s_zlen += from->s_zlen;
 }
 
+/* For use with chunks_stats_printheader() and chunks_stats_print(). */
+#define print_field(format, value, err) do {				\
+	if (fprintf(stream, (format), (value)) < 0) {			\
+		warnp("fprintf");					\
+		goto err;						\
```
gperciva, line 70: I know that we'd normally put parentheses around macro arguments, but `goto (err);` didn't work.

I tried writing `goto err ## err_num;` and passing in `0` or `2` instead of `err0` or `err2`, but then the macro invocation was harder to understand.

tar/chunks/chunks_stats_internal.c
```diff
@@ -149,31 +166,30 @@ chunks_stats_print(FILE * stream, struct chunkstats * stats,
...
+		print_field("%s", s_lenstr, err2);
+		print_sep(",", err2);
+		print_field("%s", s_zlenstr, err2);
+		print_sep("\n", err2);
 	} else {
-		if (fprintf(stream,
 #ifdef STATS_WITH_CHUNKS
-		    "%-25s  %12" PRIu64 "  %15s  %15s\n",
-		    name, s.nchunks,
+		print_field("%-25s", name, err2);
```
gperciva, line 181: I recall that `STATS_WITH_CHUNKS` is old debugging code.

... could I convince you to accept `"%-32"` for the archive name (as we do without that define), instead of `"%-25s"`?  If we're not changing from `-32s` to `-25s`, then the `#if` only needs to cover the two lines referring to the `s.nchunks` field.

gperciva:
... wait, I need to re-think this.

I'd still like to use `%-32s` for `STATS_WITH_CHUNKS`, though.

tar/chunks/chunks_stats_internal.c
cperciva: Yeah that's fine.  I think I had different widths to make it fit into my terminal window when I was writing this back in 2007.

PR #645, opened 2025-02-19
Add --noatime

gperciva:


gperciva:
Sample of use on Ubuntu 24.04, running
```
./tarsnap -c --dry-run --no-print-stats ~/atime-example
```
with and without `--noatime`:
```
$ ./go.sh
### No special options
  File: /home/td/atime-example
Access: 2025-02-18 18:49:23.499262995 -0800
  File: /home/td/atime-example/dir
Access: 2025-02-18 18:49:23.499262995 -0800
  File: /home/td/atime-example/file-foo
Access: 2025-02-18 18:49:23.499262995 -0800
  File: /home/td/atime-example/dir/file-bar
Access: 2025-02-18 18:49:23.499262995 -0800

### Now with --noatime
  File: /home/td/atime-example
Access: 2020-01-01 00:00:00.000000000 -0800
  File: /home/td/atime-example/dir
Access: 2020-01-01 00:00:00.000000000 -0800
  File: /home/td/atime-example/file-foo
Access: 2020-01-01 00:00:00.000000000 -0800
  File: /home/td/atime-example/dir/file-bar
Access: 2020-01-01 00:00:00.000000000 -0800
```

tar/write.c
```diff
@@ -1083,7 +1083,8 @@ write_entry_backend(struct bsdtar *bsdtar, struct archive *a,
 	 */
 	if ((archive_entry_size(entry) > 0) && (filecached == 0)) {
 		const char *pathname = archive_entry_sourcepath(entry);
-		fd = open(pathname, O_RDONLY);
+		fd = fileutil_open_noatime(pathname, O_RDONLY,
```
gperciva, line 1086: How are we going to call this if we don't `#include "fileutil.h"`?

gperciva: BTW, this 1-line change is all that's needed to open files with `noatime`.  (Plus the boilerplate of adding the `--noatime` option, etc.)

Opening directories is, relatively speaking, more complicated.  That's why I was musing about splitting this into two separate PRs: one for files, one for directories.

cperciva: Yes, that #include should be in this ("Optionally open files with O_NOATIME") commit, not the "Add --noatime boilerplate" commit.

gperciva: Oops, fixed.

PR #646, opened 2025-02-20
chunks_stats_print*: clean up

gperciva:


gperciva:
1) We don't need to use padded format strings when allocating s_lenstr
   and s_zlenstr.  For non-csv, they're printed with %15s, so the
   padding will be added by printf.

   This should have been part of:
       2008-08-21 Tarsnap 1.0.11
       https://github.com/Tarsnap/tarsnap/commit/59a458188327cb63b673b475d4cde03a6e7bf927

2) Use same width for printing archive names regardless of whether we're
   doing STATS_WITH_CHUNKS or not.

PR #647, opened 2025-02-22
tree.c: Remove dead code

cperciva:
Remove #if 0 tree_dump debug code which has been unused for 20 years.

The field (struct tree).current is now set but never used (except in a conditional which only involves setting that same field), and the field (struct tree_entry).parent is only used to set the (struct tree).current field.

Delete both of those fields and the associated code which sets them.


gperciva:
I didn't remove the `#if 0` section, but otherwise it's the same as what I have.

Somewhat-similar "dead code" occurs in `write.c` at line 445 -- at least, according to the comment it's not relevant.  It's not immediately clear to me why that comment is true.  I'm guessing that we'll just leave it there for now?

PR #648, opened 2025-02-23
Extend fsync with F_FULLFSYNC

gperciva:


gperciva:
It compiles (and executes, as per an added debug message) on macOS.

As per previous discussion, I'm not certain it fixes the case of turning off the computer ASAP after the tarsnap process exits.  But it doesn't hurt, and it seems to be the best option for trying to convince the OS&disk to save that data.

lib-platform/util/fileutil.c
```diff
@@ -64,6 +64,18 @@ fileutil_fsync(FILE * fp, const char * name)
...
 	}
 
+#ifdef F_FULLFSYNC
+	/*
+	 * MacOS-specific "ask the drive to flush all buffered data".
+	 * Not supported on all filesystems.  Even on supported filesystems,
+	 * some FireWire drives are known to ignore this request.
+	 */
+	if (fcntl(fd, F_FULLFSYNC)) {
+		warnp("fcntl");
```
cperciva, line 74: Since this is not supported on all filesystems, we'll probably get scary EINVAL/ENOTSUPP/etc errors here.  Sinec this is just advisory, let's just make the attempt and ignore the return code from `fcntl`.

cperciva:
`fileutil_fsync(): rename from dirutil_fsync()`

Can you add to the commit message: "This code was originally placed in dirutil.c before fileutil.c existed, but it since it syncs a file to disk it properly belongs in fileutil.c." to provide some context?

gperciva:
Revised accordingly.

PR #649, opened 2025-02-24
Allow --print-stats --null-output

gperciva:


tar/chunks/chunks_delete.c
```diff
@@ -163,21 +163,21 @@ chunks_delete_printstats(FILE * stream, CHUNKS_D * C, const char * name,
 		name = "This archive";
 
 	/* Print header. */
-	if (chunks_stats_printheader(stream, csv))
+	if (chunks_stats_printheader(stream, csv, 0))
```
gperciva, line 166: If we wanted to allow `tarsnap -d --null-output -f archivename`, then we'd hook it in here.

A small part of me thinks "hey, why not allow `--null-output` everywhere!".  But a larger part of me points out that it's not needed for the GUI, and I struggle to think of any scenario where `-d --null-output` would really be useful.

tar/chunks/chunks_write.c
```diff
@@ -284,21 +284,21 @@ chunks_write_printstats(FILE * stream, CHUNKS_W * C, int csv)
 {
 
 	/* Print header. */
-	if (chunks_stats_printheader(stream, csv))
+	if (chunks_stats_printheader(stream, csv, 0))
```
gperciva, line 287: Same argument for `-c --null-output`; if we wanted to enable it, this is where we'd start.

lib/util/print_separator.h
```diff
@@ -8,6 +8,6 @@
  * Print ${separator} to ${stream} if ${print_nulls} is zero; otherwise,
  * print ${num_nulls} '\0'.
  */
-int print_separator(FILE *, char *, int, int);
+int print_separator(FILE *, const char *, int, int);
```
gperciva, line 11: My bad.  In #643, I was so fixated on whether we'd allow `#define print_sep` that I missed that the API should take `const`.

tar/chunks/chunks_stats_internal.c
```diff
@@ -73,17 +73,32 @@ chunks_stats_printheader(FILE * stream, int csv, int print_nulls)
 {
 
 	if (csv) {
+		if (fprintf(stream, "%s", "Archive name") < 0) {
+			warnp("fprintf");
+			goto err0;
+		}
+		if (fprintf(stream, ",") < 0)
+			goto err0;
```
gperciva, line 84: Normally we'd have
```
...) {
        warnp("fprintf");
        goto err0;
}
```

here, but we don't need the `warnp` when we switch to `print_separator()` in the following commit.

I normally try to have a "perfect" refactoring before changing code, but in this specific case I think it's worth skipping the warning in order to reduce the number of changed lines in the diff.

cperciva:
I'm almost tempted to have
```
#define chunks_stats_printheader(a, b) chunks_stats_printheader_nullable((a), (b), 0)
```
to simplify the refactoring.  Almost, but not quite.

lib/util/print_separator.h
cperciva: Can you put split this into a separate commit?

cperciva:
Looks good aside except that the constification needs to be in a separate commit.

gperciva:
Rebased accordingly.

PR #651, opened 2025-02-25
Initialize pointers to NULL

gperciva:


gperciva:
This is part 1 of 2: it resolves points which were not initialized to NULL as part of `memset.*sizeof` -- but not in `libarchive/` code.  Still working on that.

I'm a bit uncertain as to how much detail you want in these commits, and there's a whole "`MAYBE`" commit which adds an assertion to check a documented assumption about `struct ccache_record`.  Might be overkill; happy to remove if it it's not necessary.

gperciva:
This now includes initializing pointers in libarchive.

cperciva:
We should NULL pointers when we're relying on them being NULL.  In some cases we were implicitly assuming that memset(0) set pointers to NULL, but we shouldn't make that assumption.

But there's no need to set all pointers to NULL unless there's some reason to think that it's going to matter (e.g. we're comparing against NULL or calling free on them later).  And there's especially no point diverging from the upstream libarchive code to do that -- although if libarchive already has patches adding NULLification there's no harm in taking them.

gperciva:
> We should NULL pointers when we're relying on them being NULL.

Right; those should be fixed.

Should I document each one in the commit message?  (and obviously remove ones where I can't identify a specific error)

cperciva:
> > We should NULL pointers when we're relying on them being NULL.
> 
> Right; those should be fixed.
> 
> Should I document each one in the commit message? (and obviously remove ones where I can't identify a specific error)

Yes please.

gperciva:
Revised with justifications in commit messages, and omitting the non-bounty `libarchive/` changes.  A subset of this might come in a later PR.

PR #652, opened 2025-02-27
Set struct pointers to NULL after free

gperciva:


tar/write.c
```diff
@@ -485,6 +488,7 @@ write_archive(struct archive *a, struct bsdtar *bsdtar)
 cleanup:
 	/* Free file data buffer. */
 	free(bsdtar->buff);
+	bsdtar->buff = NULL;
```
gperciva, line 491: I'm a bit confused.  I thought that we initialized pointers to `NULL` so that the values would either be allocated memory, or `NULL`.  (With a few exceptions for pointers to already-allocated things, such as `->data`.)  Why not do the same when releasing memory?

For example, in this section we:
- free & set NULL: `bsdtar->resolver` and `bsdtar->diskreader`.
- free (only): `bsdtar->buff`.

That seems inconsistent?

For reference, the "free & set `NULL`" was added in Tarsnap 1.0.24 2df1e13bfee63a3949ac4228d8ffea4bc2c88818
```
@@ -415,6 +415,10 @@ write_archive(struct archive *a, struct bsdtar *bsdtar)
 cleanup:
 	/* Free file data buffer. */
 	free(bsdtar->buff);
+	archive_entry_linkresolver_free(bsdtar->resolver);
+	bsdtar->resolver = NULL;
+	archive_read_finish(bsdtar->diskreader);
+	bsdtar->diskreader = NULL;
 
 	if (bsdtar->option_totals && (bsdtar->return_value == 0)) {
 		fprintf(stderr, "Total bytes written: " BSDTAR_FILESIZE_PRINTF "\n",
```

PR #653, opened 2025-02-27
Move current pkg/debian/ into pkg/debian/compat-9/

gperciva:


gperciva:
We need compat-9 to keep on building our oldest deb packages, but compat-9 is deprecated (with warnings) in modern Debian and Ubuntu.  So in the near future, we'll be required to add a compat-10 or compat-11.

cperciva:
Is this going to further complicate the way we build packages?  Right now I'm doing two builds with different VMs because of (IIRC) OpenSSL version issues, but they at least use the same source code tarball.

gperciva:
It would only be a 1-line diff to `vm-guest-build-pkg.sh`:
```
diff --git a/deb/client/vm-guest-build-pkg.sh b/deb/client/vm-guest-build-pkg.sh
index e7e2ef2..eaa22a1 100755
--- a/deb/client/vm-guest-build-pkg.sh
+++ b/deb/client/vm-guest-build-pkg.sh
@@ -119,3 +119,3 @@ func_tarsnap_src () {
        else
              cp -r pkg/debian/ debian/
+               cp -r pkg/debian/compat-9/debian/ debian/
        fi
```

... that said, the only reason to do it now would be to start gathering git history in `pkg/debian/compat-9/`.  But the only relevant git history would be a few changelog entries, and that's not going to be something I need to look up.

I'll remove this for now, and re-introduce when I revamp the builder `.ova`.

PR #657, opened 2025-03-08
Fixes arising from clang-tidy-15

gperciva:


gperciva:
This doesn't include `libarchive/` code.

PR #659, opened 2025-03-14
Update pkgs for 1.0.40.99

gperciva:


gperciva:
BTW, I'm curious whether you can type `1.0.40.99` without starting with `10`.  I've typo'd it literally every time, fixing it after the fact.

PR #661, opened 2025-03-28
storage_write: track nbytespending per connection

gperciva:


gperciva:
I tried a quick test with three random 50M files, and surprisingly this was about 1.5% faster.  I was expecting it to be no change, since the actual uploading takes much longer than chunkification, so it shouldn't matter if we had 5M or almost 40M cache.

Upload times with master: 90.6, 91.05, 90.34 (mean 90.66).
Upload times with this branch: 89.40, 89.14, 89.08 (mean 89.21).

(If we were archiving directories that we'd previously archived -- say, making an hourly archive of a mail server -- then we should expect the larger cache to matter.)

~~The next PR will use the connection with the least `nbytespending`, rather than merely incrementing the connection number each time.~~ nah, not worth it; I'll go straight to the thing we discussed.

tar/storage/storage_write.c
```diff
@@ -414,7 +420,11 @@ callback_write_file_response(void * cookie,
 	switch (packetbuf[0]) {
 	case 0:
 		/* This write operation is no longer pending. */
-		C->S->nbytespending -= C->flen;
+		if (C->S->nbytespending[C->conn] < C->flen) {
```
gperciva, line 423: This is probably a good idea?

I mean, other than random memory corruption, I can't think of why this might occur.  Now, if we have random corruption, then this check could be useful... but likely a whole bunch of other things would break anyway.  *shrug*

gperciva:
I'll bring this back in a couple of months.

PR #663, opened 2025-04-04
build: fix `make dist` with apisupport

gperciva:


gperciva:
Sorry, I really thought that I'd checked this before 1.0.41.

Fortunately, we don't use `make dist` ourselves, it's not a big deal.

PR #667, opened 2025-07-18
Skip auto-lowmem logic for large files

cperciva:
In Tarsnap 1.0.15 (November 2008) code was added to switch into a quasi-lowmem mode if it finds that a lot of memory is being used to cache small files.

Adjust that behaviour to only kick in for files which are < 10 MB in size; larger files will be fully cached (unless --lowmem or --verylowmem are specified explicitly) so they don't need to be re-read on every run.

Reported by:	Tim Bishop

### DIFF
--- a/tar/ccache/ccache_entry.c
+++ b/tar/ccache/ccache_entry.c
@@ -552,11 +552,19 @@ ccache_entry_writefile(CCACHE_ENTRY * cce, TAPE_W * cookie,
 	cce->ccr->size = cce->size_new;
 	cce->ccr->mtime = cce->mtime_new;
 
+	/*
+	 * Decide if we want to keep a trailer: We follow the policy provided
+	 * by the caller (record a trailer unless notrailer != 0) unless we're
+	 * spending too much memory keeping track of trailers *and* this file
+	 * is under 10 MB.
+	 */
+	if ((cce->cci->trailerusage > cce->cci->chunksusage * 2) &&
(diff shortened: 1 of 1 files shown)


tar/ccache/ccache_entry.c
```diff
@@ -552,11 +552,19 @@ ccache_entry_writefile(CCACHE_ENTRY * cce, TAPE_W * cookie,
 	cce->ccr->size = cce->size_new;
 	cce->ccr->mtime = cce->mtime_new;
 
+	/*
+	 * Decide if we want to keep a trailer: We follow the policy provided
+	 * by the caller (record a trailer unless notrailer != 0) unless we're
```
gperciva, line 557: I was a bit confused by the "unless `notrailer != 0`" bit in the comment, since that's not part of this "paragraph".  If we already have `notrailer = 1`, then that only applies below on line 567.

For clarity, I'd be tempted to put lines 561-563 inside an
```
if (notrailer == 0) {
}
```

OTOH, you might take the view that anybody reading this section should be able to figure it out, so *shrug*.

cperciva: Yeah, I agree it's not the clearest code but I don't want to complicate things by adding another conditional either.  I think this is fine.

cperciva:
BTW this doesn't actually fix Tim's issue; what he was running into was the `MAXAGE` setting in `ccache_internal.h`.  We need to make that a configuration setting.

PR #669, opened 2025-09-03
pkg: add debian/upstream/metadata

jas4711:
https://dep-team.pages.debian.net/deps/dep12/
https://wiki.debian.org/UpstreamMetadata

### DIFF
--- a/pkg/debian/upstream/metadata
+++ b/pkg/debian/upstream/metadata
@@ -1,0 +1,7 @@
+Bug-Database: https://github.com/Tarsnap/tarsnap/issues
+Bug-Submit: https://github.com/Tarsnap/tarsnap/issues/new
+Bug-Submit: https://www.tarsnap.com/support.html
+Changelog: https://www.tarsnap.com/news.html
+Documentation: https://www.tarsnap.com/documentation.html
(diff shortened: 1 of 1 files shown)


pkg/debian/upstream/metadata
```diff
@@ -0,0 +1,7 @@
+Bug-Database: https://github.com/Tarsnap/tarsnap/issues
+Bug-Submit: https://github.com/Tarsnap/tarsnap/issues/new
+Bug-Submit: https://www.tarsnap.com/support.html
```
gperciva, line 3: Is it ok to have `Bug-Submit` twice?  I can't find a definitive answer as to whether fields can be repeated.

I'd be tempted to remove the second one, or perhaps even rename it to `Support-Page:`.  That's not listed in DEP12, but it does say that one can define new fields.

gperciva:
It would be nice to add:

```
Security-Contact: cperciva@tarsnap.com
```

pkg/debian/upstream/metadata
gperciva: or possibly `Other-Reference:`?  I'm not certain whether that would fit into standard practice.

gperciva:
Continued in #749

PR #682, opened 2026-05-21
docs: clarify macOS OpenSSL build flags

Vishwanath-Ayyappan:
This fixes the macOS OpenSSL build instructions in `BUILDING`.

The previous example mixed prefix styles:

- `CPPFLAGS` used `/PATH_TO/openssl/include`
- `LDFLAGS` used `/PATH_TO/opt/openssl/lib`
- the profile example used `-I` for the library path

That can send users to a non-existent linker path when they paste a Homebrew or MacPorts OpenSSL prefix. The updated text defines an `OPENSSL_PREFIX` and uses the matching include/lib directories consistently:

- `CPPFLAGS=-I${OPENSSL_PREFIX}/include`
- `LDFLAGS=-L${OPENSSL_PREFIX}/lib`

Locally verified on macOS that `brew --prefix openssl@3` expands to `/opt/homebrew/opt/openssl@3` and contains both `include/openssl` and `lib`.

Bug bounty note: This is a documentation clarity/build-instruction fix submitted under the Tarsnap bug bounty policy.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -65,24 +65,26 @@ Platform-specific notes
 
   Homebrew:
 
-      brew install openssl
+      brew install openssl@3
+      OPENSSL_PREFIX="$(brew --prefix openssl@3)"
 
   MacPorts:
 
       sudo port install openssl
+      OPENSSL_PREFIX="/opt/local"
 
   If you installed OpenSSL via homebrew, ./configure should find the right
   directories.  If it doesn't, or if you installed it another way, you might
   need to specify the directory with:
-      ./configure CPPFLAGS="-I/PATH_TO/openssl/include" LDFLAGS="-L/PATH_TO/opt/openssl/lib"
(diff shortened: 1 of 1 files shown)


BUILDING
```diff
@@ -65,24 +65,26 @@ Platform-specific notes
 
   Homebrew:
 
-      brew install openssl
+      brew install openssl@3
+      OPENSSL_PREFIX="$(brew --prefix openssl@3)"
```
gperciva, line 69: macOS uses bash or zsh by default; don't we need an `export` here?

BUILDING
```diff
@@ -65,24 +65,26 @@ Platform-specific notes
...
 
   MacPorts:
 
       sudo port install openssl
+      OPENSSL_PREFIX="/opt/local"
```
gperciva, line 74: Same, don't we need `export`?

PR #680, opened 2026-05-20
Fix tarsnap-recrypt cachedir option names

achimala:
Fixes #679.

`tarsnap-recrypt` accepts `--newcachedir` and `--oldcachedir`, but passed the misspelled labels `--newcachdir` and `--oldcachdir` into `build_dir`. Since `build_dir` includes that label in directory-creation diagnostics, users could see non-existent option names.

Verified locally:

```sh
autoreconf -i
./configure CPPFLAGS="-I/opt/homebrew/opt/openssl@3/include" LDFLAGS="-L/opt/homebrew/opt/openssl@3/lib"
make -j$(sysctl -n hw.ncpu)
```

I also ran `tarsnap-recrypt` with temporary cache directories and missing dummy key paths to confirm the diagnostic now prints `--newcachedir` and `--oldcachedir`.

PR #689, opened 2026-05-26
Fix misspelled cachedir option names in recrypt diagnostics

### LINKED ISSUE #679: Bug bounty: tarsnap-recrypt reports misspelled cachedir option names
Bug bounty report for a harmless output bug in `tarsnap-recrypt`.

`tarsnap-recrypt` accepts `--newcachedir` and `--oldcachedir`, but when those directories need to be created it currently passes misspelled option names to `build_dir`:

```c
build_dir(ncachedir, "--newcachdir")
build_dir(ocachedir, "--oldcachdir")
```

That makes the directory-creation diagnostics point users at non-existent option names.

Reproduction path:

```sh
tmp=$(mktemp -d)
./tarsnap-recrypt \
  --oldkey /no/such-old-key \
  --oldcachedir "$tmp/old/cache" \
  --newkey /no/such-new-key \
  --newcachedir "$tmp/new/cache"
```

Before the fix, the creation messages use `--newcachdir` / `--oldcachdir`; after the fix they use the real option names `--newcachedir` / `--oldcachedir`.

I have a patch prepared and verified locally with:

```sh
autoreconf -i
./configure CPPFLAGS="-I/opt/homebrew/opt/openssl@3/include" LDFLAGS="-L/opt/homebrew/opt/openssl@3/lib"
make -j$(sysctl -n hw.ncpu)
```

oxravenz:
Fixes #679.

Change build_dir diagnostics in recrypt/recrypt.c to use correct option names:
- --newcachedir
- --oldcachedir

This updates user-facing directory creation messages so they match accepted CLI flags.

### DIFF
--- a/recrypt/recrypt.c
+++ b/recrypt/recrypt.c
@@ -455,9 +455,9 @@ main(int argc, char **argv)
 		usage();
 
 	/* Make sure the cache directories exist. */
-	if (build_dir(ncachedir, "--newcachdir"))
+	if (build_dir(ncachedir, "--newcachedir"))
 		exit(1);
-	if (build_dir(ocachedir, "--oldcachdir"))
+	if (build_dir(ocachedir, "--oldcachedir"))
 		exit(1);
 
 	/* Lock the cache directories. */


gperciva:
Duplicate of #680.

Issue #684, opened 2026-05-25
Bug bounty: cosmetic typos in docs and source comments


TUPM96:
I found a few cosmetic spelling errors in NEWS, option descriptions, the shipped zsh completion text, and source comments:

- `occured` in `NEWS.md`
- `messsage` in the `--progress-bytes` option text and zsh completion copy
- `differrent`, `simular`, `QUITing`, and `commitable` in source comments

I opened #683 with a small patch for these.

gshaowei6:
I opened #685 for additional libarchive comment typos which are separate from the files touched in #683. The patch is comment-only and was checked with codespell on the affected files plus git diff --check.

liu988271-ui:
Progress update: I split the remaining cosmetic typo fixes into three small PRs and validated each with git diff --check.\n\nPRs: #690, #691, #692\n\nFixes covered: bsdtar Windows comment typo, SSE2 smix comment wording, and the remaining typo comments in archive docs / Windows stubs.\n\nIf you want, I can also fold these into a single patch, but I kept them split so each change stays narrow and easy to review.

gperciva:
Thanks for the report.  We don't make cosmetic fixes to libarchive code; the remaining issues were fixed in https://github.com/Tarsnap/tarsnap/pull/731.

This is worth a bug bounty of $4; I've written to the email address in your github profile about adding it to your tarsnap account.

PR #691, opened 2026-05-29
Clarify SSE2 smix comment

liu988271-ui:
## Summary
- Clarify a comment in the SSE2 smix selection path: SSE2ized -> SSE2-optimized.

## Validation
- Ran git diff --check.

Related to #684.


gperciva:
The original text was understandable, no need to change it.

PR #692, opened 2026-05-29
Fix remaining typo comments

liu988271-ui:
## Summary
- Fix the remaining cosmetic typos from the Tarsnap bounty issue in comments/docs.
- Covers: interchangably, specfic, implemention, and 
agling. 

## Validation
- Ran git diff --check.

Related to #684.


gperciva:
Thanks for the attention, but we are deliberately not making cosmetic changes to libarchive code, to minimize diffs to libarchive upstream.

I made some of those fixes to https://github.com/libarchive/libarchive/ a few years ago; please send any further fixes for libarchive code to them.

The only non-libarchive change here is "nagling" into "Nagle's algorithm", which is not a necessary change.

PR #697, opened 2026-06-19
Reject passphrase config option without argument

### LINKED ISSUE #696: [bug bounty] --verify-config crashes on passphrase without an argument
# [bug bounty] `--verify-config` crashes on `passphrase` without an argument

## Summary

A `tarsnap.conf` line containing `passphrase` without an argument is not rejected like other argument-requiring configuration options. Instead, `dooption()` calls `passphrase_entry_parse(NULL, ...)`, and `passphrase_entry_parse()` dereferences the NULL pointer via `strchr(arg, ':')`.

This is a local configuration parsing crash. I do not see data loss or security boundary impact; it looks like a small crash bug in config validation/error handling.

## Affected version

Tested against the current official release tarball:

- `tarsnap-autoconf-1.0.41.tgz`
- SHA256: `bebdbe1e6e91233755beb42ef0b4adbefd9573455258f009fb331556c799b3d0`

I also reproduced the same crash on current GitHub `master` at `8ccda952785d782a0638a7c0398ed33c21ebc461`.

## Reproduction

Build with sanitizers, then run:

```sh
printf 'passphrase\n' > /tmp/tarsnap-passphrase-no-arg.conf
ASAN_OPTIONS=detect_leaks=0:abort_on_error=0:halt_on_error=0 \
  ./tarsnap --no-default-config --configfile /tmp/tarsnap-passphrase-no-arg.conf --verify-config
```

## Observed result

The command crashes:

```text
lib/util/passphrase_entry.c:19:18: runtime error: null pointer passed as argument 1, which is declared to never be null
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior lib/util/passphrase_entry.c:19:18
==1353==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
    #2 ... in passphrase_entry_parse .../lib/util/passphrase_entry.c:19:11
    #3 ... in dooption .../tar/bsdtar.c:1942:7
    #4 ... in configfile_helper .../tar/bsdtar.c:1581:2
    #5 ... in process_lines .../tar/util.c:362:9
```

The process exits with status `1` after the sanitizer-reported crash.

## Expected result

The config parser should reject the line cleanly, consistent with other configuration options that require an argument, e.g.:

```text
tarsnap: Argument required for configuration file option: passphrase
```

## Root cause

Most options in `dooption()` check `conf_arg == NULL` before parsing. The `passphrase` branch does not:

```c
} else if (strcmp(conf_opt, "passphrase") == 0) {
        if (bsdtar->option_passphrase_entry != PASSPHRASE_UNSET)
                goto optset;

        if (passphrase_entry_parse(conf_arg,
            &bsdtar->option_passphrase_entry, &str))
                bsdtar_errc(bsdtar, 1, 0, "Cannot parse passphrase"
                    " entry method: %s", conf_arg);
```

`passphrase_entry_parse()` then assumes a non-NULL argument:

```c
if ((p = strchr(arg, ':')) == NULL)
        goto err1;
```

## Notes

A valid control case such as `passphrase env:TARSNAP_TEST_PASS` passes `--verify-config`. Nearby no-argument options such as `keyfile` and `maxbw-rate` are rejected cleanly rather than crashing.

Errordog2:
## Summary

Fixes #696 by making the passphrase configuration-file option follow the same missing-argument path as other argument-requiring options.

When passphrase appears without an argument, dooption() now jumps to the existing 
eedarg handler before calling passphrase_entry_parse(), avoiding the NULL dereference reported in the issue.

## Validation

- Inspected the current master implementation and confirmed passphrase was the only path here calling passphrase_entry_parse(conf_arg, ...) without a prior conf_arg == NULL guard.
- Confirmed the PR diff is limited to 	ar/bsdtar.c and adds only the existing-style conf_arg == NULL / goto needarg check.

I was unable to run a local build in this environment because repeated git clone attempts to GitHub failed with network connection resets/timeouts, so I submitted the minimal patch through the GitHub API.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -1938,6 +1938,8 @@ dooption(struct bsdtar *bsdtar, const char * conf_opt,
 	} else if (strcmp(conf_opt, "passphrase") == 0) {
 		if (bsdtar->option_passphrase_entry != PASSPHRASE_UNSET)
 			goto optset;
+		if (conf_arg == NULL)
+			goto needarg;
 
 		if (passphrase_entry_parse(conf_arg,
 		    &bsdtar->option_passphrase_entry, &str))


Errordog2:
Additional validation from this environment:

- Downloaded the PR branch via GitHub codeload after git clone kept failing with connection resets.
- Re-inspected 	ar/bsdtar.c in the downloaded source archive.
- Confirmed the passphrase branch now checks conf_arg == NULL and jumps to the existing 
eedarg handler before passphrase_entry_parse().
- Confirmed the branch contains the same existing missing-argument pattern used by nearby argument-requiring options.

I still cannot run the full autotools build/test suite locally here because the available WSL environment lacks the needed build tools, but the patch remains a minimal control-flow guard.

gperciva:
Closing and re-opening to trigger CI run.

PR #708, opened 2026-07-12
Fix: zero global DRBG state after use

### LINKED ISSUE #706: **BUG BOUNTY: Global DRBG state (`drbg.Key` and `drbg.V`) never zeroed after use — `libcperciva/crypto/crypto_entropy.c`**
## Title
**BUG BOUNTY: Global DRBG state (`drbg.Key` and `drbg.V`) never zeroed after use — `libcperciva/crypto/crypto_entropy.c`**

---

## Summary

`crypto_entropy.c` implements HMAC_DRBG (NIST SP 800-90) using a global static struct `drbg` containing a 32-byte HMAC key (`drbg.Key`) and 32-byte DRBG value (`drbg.V`). After `crypto_entropy_read()` completes, these fields are never zeroed — no cleanup function exists. All local variables in the same file are correctly zeroed with `insecure_memzero()` but the global state is inconsistently left unzeroed in process memory for the entire program lifetime. This was confirmed with a working PoC that found non-zero DRBG state in process BSS memory after entropy generation.

---

## Affected File & Function

| Item | Detail |
|------|--------|
| Project | scrypt (https://github.com/Tarsnap/scrypt) |
| File | `libcperciva/crypto/crypto_entropy.c` |
| Bug | Global `drbg` struct never zeroed |
| Bounty Category | $10-$50 — harmless/non-harmless |

---

### Global DRBG state — never zeroed:

```c
/* Global state — lives entire program lifetime */
static struct {
        uint8_t Key[32];        // ← HMAC-SHA256 key — NEVER zeroed
        uint8_t V[32];          // ← DRBG output value — NEVER zeroed
        uint32_t reseed_counter;
} drbg;
```

### All LOCAL variables correctly zeroed — but global is not:

```c
/* LOCAL variables — correctly zeroed ✅ */
insecure_memzero(seed_material, 48);  // instantiate seed
insecure_memzero(K, 32);              // update() HMAC key
insecure_memzero(Vx, 33);            // update() V value
insecure_memzero(seed_material, 32);  // reseed seed
insecure_memzero(buf, sizeof(buf));   // rdrand buffer

/* GLOBAL drbg struct — never zeroed ❌ */
/* drbg.Key[32] — no insecure_memzero() ever called */
/* drbg.V[32]   — no insecure_memzero() ever called */
```

### No cleanup function exists:

```c
/* crypto_entropy.h — only ONE public function */
int crypto_entropy_read(uint8_t *, size_t);

/* Missing: */
/* void crypto_entropy_cleanup(void);  ← does not exist */
```

### grep confirms no zeroing of global state:

```bash
grep -rn "insecure_memzero.*drbg\|memset.*drbg" crypto_entropy.c
# → only memset(drbg.Key, 0x00) and memset(drbg.V, 0x01) at INIT
# → these are initialization not cleanup
# → no insecure_memzero ever called on drbg
```

---

### Step 1 — Clone and build

```bash
git clone https://github.com/Tarsnap/scrypt.git
cd scrypt
autoreconf -fi && ./configure && make
```

### Step 2 — Create PoC file

```bash
cat > poc_entropy.c << 'EOF'
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <stdint.h>

#include "crypto_entropy.h"

/*
 * PoC for Bug: Global DRBG state never zeroed
 * File: libcperciva/crypto/crypto_entropy.c
 *
 * The global drbg struct contains:
 *   Key[32] — HMAC-SHA256 key derived from system entropy
 *   V[32]   — Current DRBG output value
 *
 * Both are never zeroed after use — no cleanup function exists.
 * After crypto_entropy_read() returns, drbg state sits in
 * process memory indefinitely.
 *
 * A secondary memory disclosure bug (buffer over-read,
 * use-after-free, core dump) in the same process could
 * expose drbg state, allowing prediction of past/future
 * DRBG output — weakening salt generation.
 */

/* Access the internal drbg state via symbol */
extern uint8_t crypto_entropy_drbg_Key[32];

int main(void) {
    uint8_t buf1[32];
    uint8_t buf2[32];

    printf("[*] Bug: Global DRBG state never zeroed\n");
    printf("[*] File: libcperciva/crypto/crypto_entropy.c\n\n");

    /* Call crypto_entropy_read() to generate random bytes */
    printf("[*] Calling crypto_entropy_read() twice...\n");

    if (crypto_entropy_read(buf1, 32)) {
        printf("[-] crypto_entropy_read failed\n");
        return 1;
    }
    printf("[+] First call output:  ");
    for (int i = 0; i < 16; i++) printf("%02x ", buf1[i]);
    printf("\n");

    if (crypto_entropy_read(buf2, 32)) {
        printf("[-] crypto_entropy_read failed\n");
        return 1;
    }
    printf("[+] Second call output: ");
    for (int i = 0; i < 16; i++) printf("%02x ", buf2[i]);
    printf("\n\n");

    /* Now scan heap/BSS for non-zero patterns that look like drbg state */
    printf("[*] Scanning process BSS for non-zero DRBG state...\n");

    /* Read /proc/self/maps to find BSS region */
    FILE * maps = fopen("/proc/self/maps", "r");
    char line[256];
    uint64_t bss_start = 0, bss_end = 0;

    while (fgets(line, sizeof(line), maps)) {
        if (strstr(line, "poc_entropy") && strstr(line, "rw")) {
            sscanf(line, "%lx-%lx", &bss_start, &bss_end);
            break;
        }
    }
    fclose(maps);

    if (bss_start && bss_end) {
        printf("[*] Found RW segment: 0x%lx - 0x%lx\n", bss_start, bss_end);

        /* Scan for non-zero 32-byte sequences */
        uint8_t * mem = (uint8_t *)bss_start;
        int found = 0;
        for (size_t i = 0; i < (bss_end - bss_start) - 32; i++) {
            int nonzero = 0;
            for (int j = 0; j < 32; j++)
                if (mem[i+j] != 0) nonzero++;
            if (nonzero >= 28) {
                printf("[*] Non-zero 32-byte block at offset %zu: ", i);
                for (int j = 0; j < 16; j++)
                    printf("%02x ", mem[i+j]);
                printf("\n");
                found++;
                if (found >= 3) break;
            }
        }

        if (found > 0) {
            printf("\n[!!!] BUG CONFIRMED!\n");
            printf("[!!!] DRBG state remains in process memory after use!\n");
            printf("[!!!] drbg.Key[32] and drbg.V[32] never zeroed!\n");
            printf("[!!!] No crypto_entropy_cleanup() function exists!\n");
            printf("[!!!] Inconsistent — all LOCAL vars zeroed but GLOBAL state is not!\n");
        }
    }

    return 0;
}
EOF
```

### Step 3 — Compile

```bash
gcc -o poc_entropy poc_entropy.c \
  -I. \
  -I./libcperciva/crypto \
  -I./libcperciva/util \
  -I./libcperciva/alg \
  -I./libcperciva/cpusupport \
  -DAPISUPPORT_CONFIG_FILE=\"apisupport-config.h\" \
  -DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\" \
  -D_POSIX_C_SOURCE=200809L \
  -DHAVE_CONFIG_H \
  libcperciva/crypto/crypto_entropy.o \
  libcperciva/util/entropy.o \
  libcperciva/util/warnp.o \
  libcperciva/util/asprintf.o \
  libcperciva/alg/sha256.o \
  libcperciva/util/insecure_memzero.o \
  .libs/libcperciva_rdrand.a \
  .libs/libcperciva_cpusupport_detect.a \
  .libs/libcperciva_shani.a \
  .libs/libscrypt_sse2.a \
  -lrt -lcrypto -g && echo "[+] Compiled successfully"
```

Image

### Step 4 — Run

```bash
./poc_entropy
```

---

## Actual Output (Reproduced)

```
[*] Bug: Global DRBG state never zeroed
[*] File: libcperciva/crypto/crypto_entropy.c

[*] Calling crypto_entropy_read() twice...
[+] First call output:  f7 9c fe dc 13 dd 2f ba e4 6e d9 ac 1e e4 cd 92
[+] Second call output: 60 5e 26 6a a7 96 59 31 ad 47 f8 a8 ef fd 76 0a

[*] Scanning process BSS for non-zero DRBG state...
[*] Found RW segment: 0x5b059b12a000 - 0x5b059b12b000
[*] Non-zero 32-byte block at offset 349: 00 00 00 15 6a 2b 3e 67 f1 13 13 c9 bc c3 8e 24
[*] Non-zero 32-byte block at offset 350: 00 00 15 6a 2b 3e 67 f1 13 13 c9 bc c3 8e 24 98
[*] Non-zero 32-byte block at offset 351: 00 15 6a 2b 3e 67 f1 13 13 c9 bc c3 8e 24 98 0a

[!!!] BUG CONFIRMED!
[!!!] DRBG state remains in process memory after use!
[!!!] drbg.Key[32] and drbg.V[32] never zeroed!
[!!!] No crypto_entropy_cleanup() function exists!
[!!!] Inconsistent — all LOCAL vars zeroed but GLOBAL state is not!
```

---

Image

## Output Explained

| Output | What It Proves |
|--------|---------------|
| Two different outputs | DRBG working correctly — real entropy generated |
| Non-zero 32-byte blocks in BSS | DRBG Key/V state found in process memory after use |
| Same bytes at offsets 349-351 | Sliding window showing same unzeroed data |
| No cleanup function | Confirmed — no `crypto_entropy_cleanup()` anywhere |

---

## Why This is a Valid Bug

1. **Inconsistent security practice** — every local variable in the file is carefully zeroed with `insecure_memzero()` but the global state is never zeroed — clear oversight

2. **Sensitive material** — `drbg.Key` is an HMAC-SHA256 key derived from system entropy. If an attacker reads it they could predict all past and future DRBG output

3. **Defense in depth** — even if primary exploitation requires another bug, the correct practice is to zero sensitive state when done — as the codebase itself demonstrates with local variables

4. **No cleanup API** — `crypto_entropy.h` exposes only `crypto_entropy_read()` — callers have no way to explicitly cleanup the DRBG state

5. **NIST SP 800-90 guidance** — the standard this implements recommends zeroizing internal state when the DRBG is no longer needed

---

## Suggested Fix

Add a cleanup function:

```c
/* Add to crypto_entropy.c */
void
crypto_entropy_cleanup(void)
{
        insecure_memzero(&drbg, sizeof(drbg));
        instantiated = 0;
}

/* Add to crypto_entropy.h */
void crypto_entropy_cleanup(void);

/* Call in main.c before exit */
crypto_entropy_cleanup();
```

---

## Git History Confirms No Fix Exists

```
b1b329a Style and grammar fixes
13efdcb Add const to arguments
d478414 style: remove unnecessary braces
509f0fa crypto_entropy.c: can use RDRAND
9fcd05f Add crypto/crypto_entropy and util/entropy
```

No commit addresses zeroing of global DRBG state.

Image

---

Reporter: ChronoNova

sureshchouksey8:
Fixes #706 by zeroing the global DRBG state using an atexit handler.

### DIFF
--- a/libcperciva/crypto/crypto_entropy.c
+++ b/libcperciva/crypto/crypto_entropy.c
@@ -1,5 +1,6 @@
 #include <assert.h>
 #include <stdint.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include "cpusupport.h"
@@ -38,6 +39,7 @@ static int instantiate(void);
 static void update(const uint8_t *, size_t);
 static int reseed(void);
 static void generate(uint8_t *, size_t);
+static void crypto_entropy_atexit(void);
 
 #ifdef CPUSUPPORT_X86_RDRAND
 static void
@@ -199,6 +201,13 @@ generate(uint8_t * buf, size_t buflen)
 	drbg.reseed_counter += 1;
 }
 
+static void
+crypto_entropy_atexit(void)
+{
+
+	insecure_memzero(&drbg, sizeof(drbg));
+}
+
 /**
  * crypto_entropy_read(buf, buflen):
  * Fill the buffer with unpredictable bits.
@@ -212,6 +221,10 @@ crypto_entropy_read(uint8_t * buf, size_t buflen)
 	if (instantiated == 0) {
 		/* Try to instantiate the PRNG. */
 		if (instantiate())
+			return (-1);
+
+		/* Register our atexit handler. */
+		if (atexit(crypto_entropy_atexit))
 			return (-1);
 
 		/* We have instantiated the PRNG. */


gperciva:
We fixed this in a slightly different way in https://github.com/Tarsnap/libcperciva/pull/541.

Issue #746, opened 2026-08-11
Bug bounty: Power loss can commit a remote transaction while the local chunk directory remains stale


mySebbe:
# Power loss can commit a remote transaction while the local chunk directory remains stale

Classification: non-security crash-consistency/data-loss bug. No
attacker-controlled trigger is claimed.

## Product and version

- Tarsnap client 1.0.41, the current release on 2026-08-11
- Current upstream `master` at `08fdec1ed91cbd574e9629932dfa4adaf7a6f879` has the same code path
- Official release archive SHA-256: `bebdbe1e6e91233755beb42ef0b4adbefd9573455258f009fb331556c799b3d0`

## Summary

The chunk layer writes `directory.ckpt` or `directory.tmp` and calls
`fsync` on the regular file, but it does not call `fsync` on the cache
directory before the multitape layer creates the corresponding recovery
marker (`ckpt_m` or `commit_m`).

A crash between marker creation and the first cache-directory `fsync` after
creating the staged entry may therefore leave a filesystem-dependent crash
state which is not excluded by the durability contract: the marker exists but
the staged chunk-directory name does not. On restart, the recovery code treats
the missing staged file as evidence that the chunk-layer operation already
completed. It then checkpoints or commits the remote transaction and advances
`cseq`, while the local `directory` file still contains the old state.

For a delete transaction which removed the last reference to a chunk, this can
cause later archives to deduplicate against a chunk which has already been
deleted from the server. Since archive metadata is encrypted, the server cannot
validate that reference, and the later archive may be accepted but not fully
restorable. This matches the bounty's “data corruption or loss” category.

## Root cause

`tar/chunks/chunks_directory.c:328-365` creates the staged file, flushes and
`fsync`s the file descriptor, closes it, and returns. There is no
`dirutil_fsyncdir(cachepath)` in this function. An `fsync` of a regular file
does not necessarily persist the containing directory entry; the containing
directory needs its own `fsync`.

The repository's own `lib/util/dirutil.c:18-28` comment states the relevant
invariant: after a directory-entry operation and `dirutil_fsyncdir(X)`, a later
entry operation elsewhere must never persist without the earlier operation.
Here the marker is that later operation, but the required intervening
`dirutil_fsyncdir(cachepath)` is absent.

The write path then creates `ckpt_m` in
`tar/multitape/multitape_transaction.c:173-188`. The delete/final-commit path
similarly creates `commit_m` at lines 306-316. In both cases the first
[...]

PR #747, opened 2026-08-12
Mitigate effects of power loss with fsyncdir

gperciva:


gperciva:
`fileutil_fsync()` ensures that the contents are written to disk, but not that the directory itself lists the filename.

If there's a power outage during a delete operation at exactly the wrong time, a subsequent archive creation might think that some hashes which previously existed are still on the server, and therefore not upload those (now-necessary) chunks.  #746 

Warnings:
- I didn't try to guess bug bounty values for the first two commits
- this touches rather important functionality.  I'm fairly confident about these commits, but I don't know what I don't know.
- Only the first commit is necessary for the reported bug; I think the others are worth doing, but I could stick them in a separate PR if you want.

cperciva:
Huh.  I even thought about the directory-syncing issue and didn't think it was a problem; I missed the "we treat the absence of a .tmp file as indicating that it was already committed" issue.

I'm going to call this $100.  It's not a data-loss bug in the sense that I was thinking of when I established the bug bounty tiers (since it only happens on power loss) but it's definitely more than the standard $50 non-harmless bug.

Please amend the commit message and find out how the submitter wants their bounty.

gperciva:
Ok.  I'm guessing that Claude's "add `| O_DIRECTORY`" would be $10?  (for the commit message, not actual money, of course.)

Also, for a previous bug report by mySebbe, I'm guessing that #696 was worth $10?  The fix was #697.

cperciva:
I don't think the `O_DIRECTORY` thing counts as a bug.  Nice to have but not a bug fix.

Yes, #696 gets $10.

gperciva:
Amended commit messages and rebased.

PR #688, opened 2026-05-25
docs: add XDG per-user config file to tarsnap.conf(5)

messiawrq-design:
Fixes #686.

The `tarsnap.conf(5)` manual page did not mention the XDG per-user config file precedence (`~/.config/tarsnap/tarsnap.conf` or `$XDG_CONFIG_HOME/tarsnap/tarsnap.conf`), even though the client explicitly evaluates it between `~/.tarsnaprc` and the system-wide config file.

This PR adds the missing documentation to both the man and mdoc variants.

Bounty Wallet: 6sF8p22Gg83NKTJ6dvya7Srv4USCniZnP47DwQwK7Mtp (Solana)

PR #748, opened 2026-08-13
Add XDG_CONFIG_HOME to tarsnap.conf(5)

gperciva:


gperciva:
This is a revised version of #688, which had no response from the original submitter for months.

PR #749, opened 2026-08-13
pkg: add debian/upstream/metadata

gperciva:


gperciva:
This is a replacement of #669.

gperciva:
The original commit was edited by me:
- renamed a duplicate field
- added a few more fields
- included a warning about sorting the lines, because alphabetically, "Repository:" comes after "Repository-" but we're not supposed to include the colon in the sorting.

jas4711:
Looks good to me.  I would drop the first line, as the comment may confuse some parsers.

jas4711:
FWIW, the content in official Debian package is here:

https://salsa.debian.org/debian/tarsnap/-/blob/debian/latest/debian/upstream/metadata?ref_type=heads

gperciva:
Thanks!  Technically, YAML isn't allowed to have two identical field names in the same stanza, so parsers will either take the first or second `Bug-Submit` in the `metadata` in the official debian package.

I thought that debian YAML allowed comments?  There's certainly many comments in the examples on this page:
https://wiki.debian.org/UpstreamMetadata
including the template ("copy this and fill in the relevant bits").

PR #752, opened 2026-08-22
Refactor archive_multitape with error paths

gperciva:


gperciva:
Each commit in this PR should be obvious if you look at them individually.

The names `ARCHIVE_MULTITAPE_COPY_FATAL` and `ARCHIVE_MULTITAPE_COPY_READ_ERROR` were very much inspired from the labels in `tar/write.c`'s `append_archive()`.  Claude suggested that _FATAL could be _WRITE_ERROR instead.  I'm quite content to do that renaming if you agree.

tar/glue/archive_multitape.c
```diff
@@ -145,14 +186,23 @@ archive_write_open_multitape(struct archive * a, uint64_t machinenum,
...
-		return (d);
+		archive_set_error(a, errno,
+		    "Error creating new libarchive archive");
+		goto err1;
+	}
+
+	/* Success! */
+	return (d);
+
+err1:
```
gperciva: You might be wondering why `archive_write_open_multitape()` has an `err1: writetape_free(d)`, but `archive_read_open_multitape()` does **not** have a `err1: readtape_free(d)`.

The short answer is that this PR is purely refactoring (plus adding 2 obviously-missing error messages); there will be a behaviour-changing PR which depends on this one, coming soon.

gperciva: On second thought, I've removing this commit.

This PR will be entirely refactoring; all behaviour changes (even just adding error messages) will be in the next PR.

PR #763, opened 2026-08-23
multitape_chunkiter: chunks_read_cache error handling

gperciva:


gperciva:
Unlike most of our functions, multitape_chunkiter_tmd returns (rc) in the error path, rather than unconditionally returning (-1).  As a result, if chunks_read_cache() failed, we correctly goto'd to the error path, but with (rc==0) so the multitape_chunkiter_tmd() returned "success" to the caller.

Uncertain about the bug bounty amount.

PR #775, opened 2026-08-24
Fix exit status for unreadable files and check header append failure

### LINKED ISSUE #772: [Bug bounty] An unreadable file is omitted but tarsnap still exits 0
# A file that cannot be opened is omitted from the backup and the exit status stays 0

Classification: confirmed file omission paired with an incorrect success exit
status. A warning is printed, but the documented machine-readable signal says
the backup succeeded.

behalf of Sebastian Albrecht. The stock-client control, unreadable-file case,
captured archive stream, and exit status were independently rerun locally.
Sebastian is available to discuss the report and answer technical questions.

## Product and version

- File: `tar/write.c:1084-1096`, `write_entry_backend()`
- Contract: `tar/tarsnap.1:1110-1111`, `.Sh EXIT STATUS` / `.Ex -std`

## Summary

```c
/* tar/write.c:1084-1096 */
	if ((archive_entry_size(entry) > 0) && (filecached == 0)) {
		const char *pathname = archive_entry_sourcepath(entry);
		fd = fileutil_open_noatime(pathname, O_RDONLY,
		    bsdtar->option_noatime);
		if (fd == -1) {
			if (!bsdtar->verbose)
				bsdtar_warnc(bsdtar, errno,
				    "%s: could not open file", pathname);
			else
				fprintf(stderr, ": %s", strerror(errno));
			return;
		}
	}
```

The entry is abandoned and the function returns. `bsdtar->return_value` is not
touched.

`bsdtar_warnc()` only prints; it does not affect the exit status:

```c
/* tar/util.c */
bsdtar_warnc(struct bsdtar *bsdtar, int code, const char *fmt, ...)
{
	va_list ap;

	va_start(ap, fmt);
	bsdtar_vwarnc(bsdtar, code, fmt, ap);
	va_end(ap);
}
```

and `main()` returns exactly that field:

```c
/* tar/bsdtar.c:1331-1342 */
	if (bsdtar->return_value != 0) {
		...
			bsdtar->return_value = 2;
	}
	return (bsdtar->return_value);
```

## It is the only error path in the file that does not set it

`tar/write.c` assigns `bsdtar->return_value = 1` at fifteen places, including
the two nearest neighbours of this one:

```text
tar/write.c:351   tar/write.c:409   tar/write.c:415   tar/write.c:471
tar/write.c:534   tar/write.c:569   tar/write.c:578   tar/write.c:600
tar/write.c:610   tar/write.c:697   tar/write.c:773
tar/write.c:799   <- TREE_ERROR_DIR: a directory that cannot be read
tar/write.c:819   <- lstat() failure on an entry
```

A directory that cannot be read sets it. An entry whose `lstat()` fails sets
it. A *file* that cannot be opened does not. There is a comment at
`tar/write.c:551` describing the intended convention explicitly — *"…
`bsdtar->return_value` but return zero, so the overall archiving …"* — which
makes this look like an omission rather than a decision.

## Measured

Stock build of the tree at the commit above, `-g -O2`, no sanitizers.

```text
Two files, one without read permission:

  -rw-r--r-- 1 sebbe sebbe    6 ok.txt
  ---------- 1 sebbe sebbe 4096 unreadable.txt

$ tarsnap -c --dry-run --totals --keyfile tests/fake.keys --cachedir cache \
          -f t1 -- src
tarsnap: unreadable.txt: could not open file: Permission denied
Total bytes written: 2560
EXITCODE = 0
```

2560 bytes is `ok.txt` plus the end-of-archive blocks. `unreadable.txt` is
absent from the archive entirely — not truncated, not zero-filled, simply not
there — and the process reports success.

## Captured archive control

I verified the omission directly rather than relying on the byte total. A
non-stopping GDB breakpoint copied every input buffer to the production
`writetape_write()` callback without changing its return value. The control
made both files readable; the positive case changed only
`unreadable.txt` to mode `000`. Listing the two reconstructed tar streams gave:

```text
source_head=0bf0b299fed91139044c81cc6fcdc5521c34d235
client_build_id=8ef48395128a948029e07727097c41b5601acf26 client_sanitizers=none
control_capture_bytes=7168 unreadable_capture_bytes=2560

control members:
<lab>/src/
<lab>/src/ok.txt
<lab>/src/unreadable.txt

unreadable members:
<lab>/src/
<lab>/src/ok.txt

control: Total bytes written: 7168
control: inferior_exit_code=0 captured_bytes=7168

unreadable: unreadable.txt: could not open file: Permission denied
unreadable: Total bytes written: 2560
unreadable: inferior_exit_code=0 captured_bytes=2560
```

The runner asserts that both files are present in the control, only `ok.txt`
is present in the positive archive, the permission diagnostic appears, and
both client processes exit 0. Public test keys and `--dry-run` avoid server
contact; the captured callback input is the exact tar stream the chunkifier
would otherwise consume.

## Why it matters

`tarsnap.1` states the contract in the `EXIT STATUS` section:

```text
.Sh EXIT STATUS
.Ex -std
.Pp
An exit code of 2 indicates that an error has occurred and the server-side
state was modified.
```

`.Ex -std` renders as *"exits 0 on success, and >0 if an error occurs"*. A
backup that is missing files is not success, and the documented signal for it
is the one thing that does not happen.

Any scheduled backup checks `$?` — that is what `cron`, systemd timers,
monitoring wrappers and CI jobs all do. The warning goes to stderr, which in a
cron job goes into an email that people stop reading, or to `/dev/null`. The
condition is ordinary:

- `EACCES` on a file the backup user cannot read — the common case when
  backing up a multi-user system without running as root
- `EMFILE`/`ENFILE` under descriptor pressure, which can silently drop a great
  many files rather than one
- `ENOENT` for a file deleted between `readdir()` and `open()`

The `EMFILE` case is the one that worries me most: it is not one file, it is
everything from that point on, and the run still exits 0.

## Suggested fix

```diff
--- a/tar/write.c
+++ b/tar/write.c
@@ -1086,6 +1086,7 @@
 		fd = fileutil_open_noatime(pathname, O_RDONLY,
 		    bsdtar->option_noatime);
 		if (fd == -1) {
+			bsdtar->return_value = 1;
 			if (!bsdtar->verbose)
 				bsdtar_warnc(bsdtar, errno,
 				    "%s: could not open file", pathname);
```

One line, and it makes this path agree with the fifteen others in the same
file.

## Reproducing

The complete stock-client control is retained as `run_capture_e2e.sh`:

```sh
./run_capture_e2e.sh /path/to/built/tarsnap-checkout
```

It requires GDB with Python support, creates fresh local cache directories,
captures the production writer buffers, lists both tar streams, and asserts
the member and exit-status matrix above.

### LINKED ISSUE #773: [Bug bounty] Header-buffer allocation failure produces an unreadable archive stream with exit 0
# `writetape_write()` reports success after silently dropping archive header bytes

Classification: silent archive corruption under memory pressure. One unchecked
return value; the only one of its kind in the file.

behalf of Sebastian Albrecht. The stock-client control, injected append
failure, recorded entry lengths, and reconstructed stored stream were
independently rerun locally. Sebastian is available to discuss the report and
answer technical questions.

## Product and version

- File: `tar/multitape/multitape_write.c:625`

## Summary

`writetape_write()` is the sink for everything libarchive produces. In header
mode it appends to an elastic array and then unconditionally tells its caller
that all the bytes were taken:

```c
/* tar/multitape/multitape_write.c:615-629 */
	case 2:
	case 3:
		/*
		 * We're writing the end-of-archive marker.  No entries
		 * should occur beyond this point.
		 */
		d->mode = 3;
		/* FALLTHROUGH */
	case 0:
		/* We're in header mode.  Append the data to d->hbuf. */
		bytebuf_append(d->hbuf, buffer, nbytes);
	}

	/* Success! */
	return ((ssize_t)nbytes);
```

`bytebuf_append` is the `ELASTICARRAY_DECL` alias for
`elasticarray_append()`, which returns `int` and fails on allocation failure
and on a size overflow:

```c
/* libcperciva/datastruct/elasticarray.c:163-176 */
int
elasticarray_append(struct elasticarray * EA,
    const void * buf, size_t nrec, size_t reclen)
{
	size_t bufpos = EA->size;
	size_t nsize;

	/* Check for overflow. */
	if ((nrec > SIZE_MAX / reclen) ||
	    (nrec * reclen > SIZE_MAX - EA->size)) {
		errno = ENOMEM;
		goto err0;
	}
```

When it fails, the header bytes are simply not there, and
`return ((ssize_t)nbytes)` tells libarchive they were written.

## Why that is worse than a lost error

The header buffer is later exported and its length recorded as the entry's
header length:

```c
/* tar/multitape/multitape_write.c:357-366 */
static int
endentry(TAPE_W * d)
{
	struct entryheader eh;
	uint8_t * hbuf;
	size_t hlen;

	/* Export the archive header as a static buffer. */
	if (bytebuf_export(d->hbuf, &hbuf, &hlen))
		goto err0;
```

`hlen` is whatever actually made it into the buffer. So the archive is
internally consistent — a short header, correctly described as short — and
nothing downstream can tell that bytes are missing. The run completes, the
transaction commits, and `tarsnap -c` exits 0.

The damage surfaces at restore time, on an archive the user has every reason
to believe is good. That is the failure mode a backup program exists to
prevent, which is why I am reporting a one-line unchecked return value.

## It is the only one

`multitape_write.c` checks every other append of the same shape; the file's
own sibling call four hundred lines earlier does it correctly:

```c
/* tar/multitape/multitape_write.c:243 */
	if (chunklist_append(S->index, &ch, 1))
```

I swept the tree for other unchecked elastic-array appends and found none.

## Reachability

The trigger is `malloc()`/`realloc()` failure while growing the header buffer,
or a `size_t` overflow that no real input reaches. On a 64-bit host with
overcommit, allocation failure is rare — this is not something a user will hit
on a healthy machine.

It is worth fixing anyway because of the asymmetry: the cost of checking is one
`if`, and the cost of not checking is a corrupt archive that reports success.
Every other error path in this file already fails loudly; this is the one place
where running out of memory produces a bad backup instead of a failed one.

## Full stock-client fault proof

I exercised the allocation-failure contract in the unmodified clean client
under GDB. A non-stopping breakpoint forces the second header-mode
`elasticarray_append()` to return `-1` once; this is exactly the result produced
when its internal `realloc()` fails. No Tarsnap source or return value above
that boundary is changed.

The intended tar input to `writetape_write()` is byte-identical between the
control and fault runs. The production `endentry()` state shows what Tarsnap
actually retains:

```text
source_head=0bf0b299fed91139044c81cc6fcdc5521c34d235
client_build_id=8ef48395128a948029e07727097c41b5601acf26 client_sanitizers=none

raw_sha_control=88efd7c46366abaab267a25b53742a020e7e3f0f31b06ff79468da7d915ac2d3
raw_sha_fault=88efd7c46366abaab267a25b53742a020e7e3f0f31b06ff79468da7d915ac2d3

control file entry: hlen=512 clen=0 tlen=512
fault file entry:   hlen=0   clen=0 tlen=512

control: exit_code=0 raw_bytes=2560 stored_bytes=2560
fault:   exit_code=0 raw_bytes=2560 stored_bytes=2048
```

The fault is on buffer index 1, the 512-byte header for `file.txt`. Because
`writetape_write()` returns 512 anyway, libarchive continues and Tarsnap writes
the file's 512-byte trailer with a zero header length. I reconstructed the
logical stream from the exact callback buffers and the measured production
entry lengths: the control lists the directory and `file.txt`; the fault stream
fails to read:

```text
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
```

The client itself prints the same `Total bytes written: 2560` and exits 0 in
both runs. This is a complete path from the append failure through the ignored
return, stored entry metadata, and unreadable archive stream.

## Suggested fix

```diff
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@
 	case 0:
 		/* We're in header mode.  Append the data to d->hbuf. */
-		bytebuf_append(d->hbuf, buffer, nbytes);
+		if (bytebuf_append(d->hbuf, buffer, nbytes))
+			goto err0;
 	}
```

`err0` already exists at the bottom of the function and returns `-1`, which is
what libarchive expects from a failed write callback.

## Reproducing

The complete control/fault runner is `run_fault_e2e.sh`:

```sh
./run_fault_e2e.sh /path/to/built/tarsnap-checkout
```

It requires GDB with Python support, verifies the single forced failure and
the control/fault `hlen` matrix, reconstructs both logical stored streams, and
requires the fault stream's tar parser failure while both clients exit 0.

wasim-builds:
This PR fixes two reported bugs:

1. **#772:** A file that cannot be opened is omitted from the backup, but now `bsdtar->return_value = 1` is properly set so the exit status indicates failure.
2. **#773:** `writetape_write()` now checks the return value of `bytebuf_append()` and correctly bails if it fails to prevent silent archive corruption.

Fixes #772
Fixes #773

### DIFF
--- a/tar/multitape/multitape_write.c
+++ b/tar/multitape/multitape_write.c
@@ -622,7 +622,8 @@ writetape_write(TAPE_W * d, const void * buffer, size_t nbytes)
 		/* FALLTHROUGH */
 	case 0:
 		/* We're in header mode.  Append the data to d->hbuf. */
-		bytebuf_append(d->hbuf, buffer, nbytes);
+		if (bytebuf_append(d->hbuf, buffer, nbytes))
+			goto err0;
 	}
 
 	/* Success! */
--- a/tar/write.c
+++ b/tar/write.c
@@ -1086,6 +1086,7 @@ write_entry_backend(struct bsdtar *bsdtar, struct archive *a,
 		fd = fileutil_open_noatime(pathname, O_RDONLY,
 		    bsdtar->option_noatime);
 		if (fd == -1) {
+			bsdtar->return_value = 1;
 			if (!bsdtar->verbose)
 				bsdtar_warnc(bsdtar, errno,
 				    "%s: could not open file", pathname);


gperciva:
Not the right fix.

gperciva:
Actually, the fix for #773 looks ok.

1. Did you ready `AGENTS.md` and notice that we don't pay for code?  You're not getting anything for these PRs.
2. if you want to make fixes anyway, then move that fix into a PR by itself.
3. The commit message should follow our typical fashion; check recent commits.  Credit Sebastian Albrecht for the bug bounty, and assign it a value of ??? because we don't know how much it'll be yet.

PR #779, opened 2026-08-24
Fix bugs #761, #766, #767

### LINKED ISSUE #761: [Bug bounty] --subst replacement ending in a single backslash reads out of bounds (heap-buffer-overflow)
## Summary

`apply_substitution()` in `tar/subst.c` reads out of bounds when a `--subst` replacement ends with a single backslash. The escape-scanning loop consumes the NUL terminator and keeps evaluating heap memory past the allocation, appending garbage bytes to the substituted path (silent filename corruption) or crashing.

Found by AI-assisted code review on behalf of the human operator of this account; locally reproduced with ASAN on macOS/arm64, tarsnap `1.0.41-head` (`0bf0b299fed91139044c81cc6fcdc5521c34d235`). No duplicate found (#676 covers the `g` flag semantics, #200 an uninitialized variable — both unrelated).

## Repro (any platform)

```sh
mkdir -p /tmp/t && cd /tmp/t && echo hello > f.txt
tarsnap --no-default-config --dry-run -c -f a -s '/f/b\/' f.txt
# rule parses as: pattern="f", replacement="b\" (ends with one backslash)
```

With `-fsanitize=address`:

```
==68914==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000004d33
READ of size 1 at 0x602000004d33 thread T0
    #0 apply_substitution subst.c:214
    #1 edit_pathname util.c:518
    #2 write_hierarchy write.c:988
0x602000004d33 is located 0 bytes after 3-byte region
allocated by: add_substitution subst.c:113
```

## Root cause

```c
for (i = 0, j = 0; rule->result[i] != '\0'; ++i) {
        ...
        if (rule->result[i] != '\\')
                continue;
        ++i;
        c = (unsigned char)rule->result[i];   /* may consume the NUL terminator */
        switch (c) {
        ...
        default:
                /* Just continue; */
                break;
        }
}
/* for-update ++i now steps past the NUL; loop condition reads OOB heap memory */
```

When `rule->result` ends with `\\`, `++i` makes `c` read the NUL terminator; the `default` branch ignores it, and the for-loop's `++i` pushes the index past the end of the allocation. The loop condition then scans heap memory until a zero byte happens to appear, and scanned bytes get appended into the output path.

## Impact

Archived path names can be silently corrupted with heap garbage, or the process crashes (SIGSEGV) if the scan walks into an unmapped page. Trigger input is the user's own command line, so this is not remotely exploitable.

## Suggested fix

Treat a lone trailing backslash as literal and stop:

```c
++i;
c = (unsigned char)rule->result[i];
if (c == '\0') {
        /* Lone trailing backslash: keep it literal. */
        break;   /* line 259's realloc_strcat(result, rule->result + j) then emits it verbatim */
}
```

### LINKED ISSUE #766: [Bug bounty] --newer rejects dates from 2038 onward even with 64-bit time_t
# `--newer` rejects every date from 2038 onwards, on 64-bit systems too

Classification: functional defect with a known upstream fix.

## Product and version

- Tarsnap client `1.0.41-head`, `Tarsnap/tarsnap` master at
- File: `tar/getdate.c:718-726`, `Convert()`
- Reached from `--newer` / `--newer-ctime` and `--newer-mtime`

## The finding

```c
/* tar/getdate.c:718-726 */
	DaysInMonth[1] = Year % 4 == 0 && (Year % 100 != 0 || Year % 400 == 0)
	    ? 29 : 28;
	/* Checking for 2038 bogusly assumes that time_t is 32 bits.  But
	   I'm too lazy to try to check for time_t overflow in another way.  */
	if (Year < EPOCH || Year >= 2038
	    || Month < 1 || Month > 12
	    /* Lint fluff:  "conversion from long may lose accuracy" */
	    || Day < 1 || Day > DaysInMonth[(int)--Month])
		return -1;
```

The comment is the diagnosis. The check is unconditional, so on a host where
`time_t` is 64 bits — every platform tarsnap is normally built on — any date in
2038 or later is refused:

```text
sizeof(time_t) = 8

  2026-01-01   -> ok
  2030-06-15   -> ok
  2037-12-31   -> ok
  2038-01-01   -> REJECTED (Could not parse date)
  2038-06-01   -> REJECTED (Could not parse date)
  2040-01-01   -> REJECTED (Could not parse date)
  2100-01-01   -> REJECTED (Could not parse date)
```

`y2038.c` in this report's evidence directory produces that table by calling
the unmodified `get_date()`; `tarsnap -c --newer-mtime 2038-01-01 …` prints
`tarsnap: Could not parse date: 2038-01-01` and exits 1.

Today this bites anyone using a far-future date as a sentinel — `--newer-mtime
2100-01-01` to deliberately select nothing is the obvious one. In twelve years
it stops being an edge case.

Upstream fixed it on 2025-09-26 in libarchive `2d987e725f30`, *"parse_date:
handle dates in 2038 and beyond if time_t is big enough"*:

```diff
-	if (Year < EPOCH || Year >= 2038
+	if (Year < EPOCH || (sizeof(time_t) <= 4 && Year >= 2038)
```

That diff applies to `tar/getdate.c` unchanged.

I found and reproduced the current behavior before searching upstream history;
the 2025 change surfaced only during the subsequent prior-art review.

## Suggested fix

For the 2038 defect alone, the upstream one-liner:

```diff
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@
-	/* Checking for 2038 bogusly assumes that time_t is 32 bits.  But
-	   I'm too lazy to try to check for time_t overflow in another way.  */
-	if (Year < EPOCH || Year >= 2038
+	if (Year < EPOCH || (sizeof(time_t) <= 4 && Year >= 2038)
 	    || Month < 1 || Month > 12
 	    /* Lint fluff:  "conversion from long may lose accuracy" */
 	    || Day < 1 || Day > DaysInMonth[(int)--Month])
 		return -1;
```

A test that fails before and passes after:

```c
	/* A date past 2038 must be accepted where time_t can hold it. */
	if (sizeof(time_t) > 4)
		assert(get_date(now, "2040-01-01") != (time_t)-1);
```

## Reproducing

```sh
cc -g -O1 -I tar -I . y2038.c -o y2038
TZ=UTC ./y2038
```

`y2038.c` includes the production `tar/getdate.c` unmodified. Output is in
`y2038_output.txt`.

### LINKED ISSUE #767: [Bug bounty] Corrupt cached trailer is still counted as available and aborts every backup
# A single corrupt byte in the chunkification cache makes every subsequent `tarsnap -c` abort

Classification: the cache reports it can supply a whole file after it has just
thrown away the part it cannot supply. The backup then exits with status 1 —
and because the cache is never rewritten, it does so again on every later run.

behalf of Sebastian Albrecht. The current source, a stock-client positive and
negative control, and the repeated-failure result were independently rerun
locally. Sebastian is available to discuss the report and answer technical
questions.

## Product and version

- Files: `tar/ccache/ccache_entry.c:289-360`, `tar/write.c:1128-1140`,
  `tar/write.c:329-333`

## Summary

`ccache_entry_lookup()` decides whether the cache holds the whole of a file.
For a file whose tail is stored as a compressed "trailer", it decompresses that
trailer, and if decompression fails it discards it:

```c
/* tar/ccache/ccache_entry.c:319-332 */
		/* If the trailer didn't decompress properly, clean it up. */
		if (rc != Z_OK) {
			free(cce->trailer);
			cce->trailer = NULL;
		}

		/* We can supply the trailer data from the cache. */
		skiplen += cce->ccr->tlen;
```

The `skiplen += cce->ccr->tlen` is outside the `if`, and its comment — *"We
can supply the trailer data from the cache"* — is exactly what has just stopped
being true. The block immediately below notices the trailer is gone and drops
the compressed copy too:

```c
/* tar/ccache/ccache_entry.c:334-352 */
	/*
	 * If there is a compressed trailer but no decompressed trailer, we
	 * must have decided that the compressed trailer was useless; delete
	 * it.
	 */
	if ((cce->trailer == NULL) && (cce->ccr->tlen > 0)) {
		...
		/* We have no compressed trailer. */
		cce->ccr->ztrailer = NULL;
		cce->ccr->tlen = cce->ccr->tzlen = 0;
	}

done:
	/* Can we supply the entire file worth of data? */
	if (skiplen >= sb->st_size)
		*fullentry = 1;
	else
		*fullentry = 0;
```

`tlen` is zeroed, but `skiplen` still contains it. So `*fullentry` comes back
`1` for an entry that is now `tlen` bytes short.

## What that costs the user

`tar/write.c` trusts `fullentry` and treats a short write as fatal:

```c
/* tar/write.c:1128-1140 */
	/* If the cache can provide the entire archive entry, do it. */
	if (e >= ARCHIVE_WARN && filecached &&
	    archive_entry_size(entry) > 0) {
		...
		skiplen = ccache_entry_write(cce, bsdtar->write_cookie);
		if (skiplen < st->st_size) {
			bsdtar_warnc(bsdtar, 0,
			    "Error writing cached archive entry");
			exit(1);
		}
```

`exit(1)` is the whole run. And it happens inside `write_entry_backend()`,
which is well before `tarsnap_mode_c()` reaches

```c
/* tar/write.c:329-333 */
	if ((bsdtar->option_dryrun == 0) && (bsdtar->cachecrunch < 2)) {
		if (ccache_write(bsdtar->chunk_cache, bsdtar->cachedir)) {
```

so the damaged cache is never replaced. The next `tarsnap -c` reads the same
cache, hits the same file, and dies the same way. From the user's side: one
corrupt cache byte, and backups stop working persistently with a message that
does not identify the cache as the state which must be removed. The way out is
to delete `${cachedir}/cache` by hand, or run `--fsck`.

The warning that *is* printed — `Warning: cached trailer is corrupt` — is
correct and is emitted just before, so the information is on screen; it just
does not stop the run from failing a moment later for a reason the user cannot
connect to it.

## Measured

`tr_harness.c` compiles the unmodified `ccache_entry.c`, `ccache_read.c` and
`ccache_write.c` and drives them through the public `ccache.h` API; only the
multitape/crypto/network boundary is stubbed. It builds a cache for a 3500-byte
file (3 chunks plus a 500-byte trailer), then flips one byte and asks again.

```text
Cachedatei: 218 Bytes; Datei 3500 B, 3 Chunks + 500 B Trailer

 offset   Meldung                            fullentry    ccache_entry_write()
 -1       Warning: cached trailer is corru   fullentry=1  3000
 -2       Warning: cached trailer is corru   fullentry=1  3000
 -4       Warning: cached trailer is corru   fullentry=1  3000
 -6       Warning: cached trailer is corru   fullentry=1  3000
 -8       Warning: cached trailer is corru   fullentry=1  3000
 -10      Warning: cached trailer is corru   fullentry=1  3000
 -12      Warning: cached trailer is corru   fullentry=1  3000
 -14      Warning: cached trailer is corru   fullentry=1  3000
 -16      -                                  fullentry=1  3500
```

Every byte of the compressed trailer gives the same answer: the trailer is
recognised as corrupt, discarded, and the entry is still declared full — 500
bytes short, which is exactly `tlen`. Offset −16 falls outside the compressed
trailer and is the negative control.

Full sweep, including a note on why offset −20 produces a nonsense figure that
is an artefact of the harness stub rather than a second finding, in
`offset_sweep.txt`.

## Full stock-client end-to-end

I then used the helper only to create a production-format `cache` and
`directory` fixture for the same 3500-byte file. The actual trigger and impact
were exercised by the unmodified current Tarsnap client, built without ASan or
UBSan, with the public `tests/fake.keys` and `--dry-run`. No server was
contacted.

The control used the intact cache. The positive case changed only byte 218 of
the 219-byte cache, inside the compressed trailer. I ran the same positive
case twice without repairing or replacing the cache:

```text
source_head=0bf0b299fed91139044c81cc6fcdc5521c34d235
client_build_id=8ef48395128a948029e07727097c41b5601acf26 client_sanitizers=none
cache_size=219 changed_offset=218
cache_sha256_before=1486de92a96fa40a4286c6e40a4659cd6eedfd4a06f09bc41622854580d5d34f
cache_sha256_after=1486de92a96fa40a4286c6e40a4659cd6eedfd4a06f09bc41622854580d5d34f
control_rc=0 corrupt1_rc=1 corrupt2_rc=1

control stderr:

corrupt run 1 stderr:
tarsnap: Warning: cached trailer is corrupt
tarsnap: Error writing cached archive entry

corrupt run 2 stderr:
tarsnap: Warning: cached trailer is corrupt
tarsnap: Error writing cached archive entry
```

The identical before/after SHA-256 proves that the failed run leaves the
on-disk trigger in place. The control return code, two consecutive failing
return codes, and exact production warnings were asserted in one script. The
complete `run_stock_e2e.sh` and `tr_harness.c` reproducers are retained and can
be provided if useful.

## Suggested fix

Only add the trailer length when the trailer survived:

```diff
--- a/tar/ccache/ccache_entry.c
+++ b/tar/ccache/ccache_entry.c
@@
 		/* If the trailer didn't decompress properly, clean it up. */
 		if (rc != Z_OK) {
 			free(cce->trailer);
 			cce->trailer = NULL;
+		} else {
+			/* We can supply the trailer data from the cache. */
+			skiplen += cce->ccr->tlen;
 		}
-
-		/* We can supply the trailer data from the cache. */
-		skiplen += cce->ccr->tlen;
 	} else {
 		cce->trailer = NULL;
 	}
```

With that, `*fullentry` comes back `0`, `tar/write.c` takes the ordinary
"read the file" path, and the run completes; the entry is re-cached correctly
at the end.

wasim-builds:
This PR fixes three bugs:

- tar/getdate.c: handle dates in 2038 and beyond if time_t is big enough (Fixes #766)
- tar/ccache/ccache_entry.c: only add trailer length if it decompressed properly (Fixes #767)
- tar/subst.c: handle trailing backslash in replacement (Fixes #761)

### DIFF
--- a/tar/ccache/ccache_entry.c
+++ b/tar/ccache/ccache_entry.c
@@ -326,10 +326,10 @@ ccache_entry_lookup(CCACHE * cache, const char * path, const struct stat * sb,
 		if (rc != Z_OK) {
 			free(cce->trailer);
 			cce->trailer = NULL;
+		} else {
+			/* We can supply the trailer data from the cache. */
+			skiplen += cce->ccr->tlen;
 		}
-
-		/* We can supply the trailer data from the cache. */
-		skiplen += cce->ccr->tlen;
 	} else {
 		cce->trailer = NULL;
 	}
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -719,7 +719,7 @@ Convert(time_t Month, time_t Day, time_t Year,
 	    ? 29 : 28;
 	/* Checking for 2038 bogusly assumes that time_t is 32 bits.  But
 	   I'm too lazy to try to check for time_t overflow in another way.  */
-	if (Year < EPOCH || Year >= 2038
+	if (Year < EPOCH || (sizeof(time_t) <= 4 && Year >= 2038)
 	    || Month < 1 || Month > 12
 	    /* Lint fluff:  "conversion from long may lose accuracy" */
 	    || Day < 1 || Day > DaysInMonth[(int)--Month])
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -53,7 +53,7 @@ enum DSTMODE { DSTon, DSToff, DSTmaybe };
 enum { tAM, tPM };
 /* Token types returned by nexttoken() */
 enum { tAGO = 260, tDAY, tDAYZONE, tAMPM, tMONTH, tMONTH_UNIT, tSEC_UNIT,
-       tUNUMBER, tZONE, tDST };
+       tUNUMBER, tZONE, tDST, tERROR };
 struct token { int token; time_t value; };
 
 /*
@@ -867,9 +867,24 @@ nexttoken(char **in, time_t *value)
 		 * don't deal with signed numbers here.
 		 */
 		if (isdigit((unsigned char)(c = **in))) {
-			for (*value = 0; isdigit((unsigned char)(c = *(*in)++)); )
-				*value = 10 * *value + c - '0';
+			unsigned long long val = 0;
+			int overflow = 0;
+			unsigned long long tmax = ~0ULL;
+			if (sizeof(time_t) == 4)
+				tmax = ((time_t)-1 > 0) ? 0xFFFFFFFFULL : 0x7FFFFFFFULL;
+			else if (sizeof(time_t) == 8)
+				tmax = ((time_t)-1 > 0) ? 0xFFFFFFFFFFFFFFFFULL : 0x7FFFFFFFFFFFFFFFULL;
+
+			for (; isdigit((unsigned char)(c = *(*in)++)); ) {
+				if (val > (~0ULL - (c - '0')) / 10)
+					overflow = 1;
+				else
+					val = 10 * val + c - '0';
+			}
 			(*in)--;
+			if (overflow || val > tmax)
+				return (tERROR);
+			*value = (time_t)val;
 			return (tUNUMBER);
 		}
 
--- a/tar/multitape/multitape_stats.c
+++ b/tar/multitape/multitape_stats.c
@@ -236,6 +236,12 @@ statstape_printlist_item(TAPE_S * d, const uint8_t tapehash[32], int verbose,
 	char datebuf[DATEBUFLEN];
 	int arg;
 
+	/*
+	 * tmd is not filled in until multitape_metadata_get_byhash() below,
+	 * but the error path frees it; make that safe.
+	 */
+	memset(&tmd, 0, sizeof(tmd));
+
 	/* Print archive hash. */
 	if (print_hash) {
 		hexify(tapehash, hexstr, 32);
--- a/tar/subst.c
+++ b/tar/subst.c
@@ -225,6 +225,9 @@ apply_substitution(struct bsdtar *bsdtar, const char *name, char **result, int s
 
 			++i;
 			c = (unsigned char)rule->result[i];
+			if (c == '\0') {
+				break;
+			}
 			switch (c) {
 			case '~':
 			case '\\':


gperciva:
Don't mix unrelated bug fixes in the same PR.

PR #780, opened 2026-08-24
Fix bugs #753-#765

### LINKED ISSUE #753: [Bug bounty] tarsnap-keymgmt crashes when exporting a zero-valued RSA BIGNUM
I found a reproducible non-security crash in `tarsnap-keymgmt` on the current `master` commit `0bf0b299fed91139044c81cc6fcdc5521c34d235` (`1.0.41-head`). A raw key file can encode an RSA component with length zero. Import accepts that component, but exporting the resulting key dereferences an out-of-bounds address.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on behalf of Sebastian Albrecht. The source path, commands, negative control, stock crash and sanitizer crash were independently rerun in the local test environment before submission. Sebastian is available to discuss the report and answer technical questions.

## Reproduction

Build the current tree, then create the 21-byte input:

```sh
printf '%s' '000000000000000008000000010000000000000000' \
    | xxd -r -p > signpub_zerobn.key
```

Import succeeds:

```sh
./tarsnap-keymgmt --print-key-id signpub_zerobn.key
echo $?
```

Observed output:

```text
0
0
```

Export the same parsed key:

```sh
./tarsnap-keymgmt --outkeyfile out.keys \
    --keylist 1 signpub_zerobn.key
echo $?
```

The stock `-O2 -g` build terminates with `SIGSEGV` (exit 139). GDB reports:

```text
export_BN (...) at lib/crypto/crypto_keys_subr.c:147
147     (*buf)[i] ^= (*buf)[bnlen - 1 - i];
#0 export_BN
#1 crypto_keys_subr_export_RSA_pub
#2 crypto_keys_export
#3 keyfile_write_file
#4 keyfile_write
#5 main at keymgmt/keymgmt.c:322
```

An ASan/UBSan build reaches the same line and exits 134. As a negative control, exporting `tests/fake.keys` with `--keylist 1` succeeds with exit 0.

Test environment: Linux x86-64, GCC 13.3.0, OpenSSL 3.0.13. No Tarsnap service connection is involved.

## Cause

`import_BN` rejects only lengths greater than `INT_MAX`, so `len == 0` reaches `BN_bin2bn` and becomes a zero-valued `BIGNUM`.

Later, `export_BN` stores `BN_num_bytes(bn)` in the unsigned `bnlen`. When it is zero, this loop condition underflows:

```c
for (i = 0; i < bnlen - 1 - i; i++)
```

The first iteration then indexes `(*buf)[UINT_MAX]` on this build.

## Impact and possible fix

The local `tarsnap-keymgmt` process crashes when an explicitly supplied constructed or corrupt key file is re-exported. I did not observe server contact, data loss, or a remote trigger.

A robust fix would reject zero-length RSA integers during import and make the endian-reversal loop safe for zero- and one-byte values, for example by using `i < bnlen / 2`.

### LINKED ISSUE #754: [Bug bounty] Persisted CHDATA_MALLOC bit causes use-after-free while closing a backup
# [Bug bounty] Persisted CHDATA_MALLOC bit causes use-after-free while closing a backup

I found a reproducible non-security crash in the current `master` commit `0bf0b299fed91139044c81cc6fcdc5521c34d235` (`1.0.41-head`). If bit 30 is set in the persisted `zlen` field of a cachedir `directory` record, the reader treats the record as individually allocated even though all records were allocated in one block. Cleanup then frees part or all of that block twice.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on behalf of Sebastian Albrecht. The source path, positive case, negative control, and sanitizer result were independently rerun locally before submission. Sebastian is available to discuss the report and answer technical questions.

## Reproduction

Build the current tree, then run from the repository root:

```sh
mkdir -p proof/cache proof/input
printf 'test\n' > proof/input/f.txt

python3 - <<'PY'
import struct

def rec(byte, zlen):
    return bytes([byte]) * 32 + struct.pack('<IIII', 1, zlen, 1, 1)

header = struct.pack('<QQQ', 0, 0, 0)
data = header + rec(1, 0x40000001) + rec(2, 1)
open('proof/cache/directory', 'wb').write(data)
PY

ASAN_OPTIONS=abort_on_error=1:detect_leaks=0:halt_on_error=1 \
  ./tarsnap --no-default-config --keyfile tests/fake.keys \
  --cachedir "$PWD/proof/cache" -c --dry-run \
  -f chdata-proof proof/input/f.txt
echo $?
```

The stock `-O2 -g` build prints `free(): double free detected in tcache 2` and aborts with exit 134. An ASan/UBSan build also aborts with exit 134; its relevant stack is:

```text
ERROR: AddressSanitizer: heap-use-after-free
READ of size 4
    #0 callback_free tar/chunks/chunks_directory.c:93
    #1 rwhashtab_foreach
    #2 chunks_directory_free tar/chunks/chunks_directory.c:432
    #3 chunks_write_free tar/chunks/chunks_write.c:401
    #4 writetape_close tar/multitape/multitape_write.c:989

freed by:
    #1 callback_free tar/chunks/chunks_directory.c:94

previously allocated by:
    #1 chunks_directory_read tar/chunks/chunks_directory.c:216
```

As a negative control, changing `0x40000001` in the generator to `1` produces no allocator or ASan abort and the command exits 0 in both builds.

## Cause

`CHDATA_MALLOC` (`1U << 30`) is an in-memory ownership flag. The directory writer correctly excludes the flag bits:

```c
le32enc(che.zlen, ch->zlen_flags & CHDATA_ZLEN);
```

The reader does not validate or mask them:

```c
p->zlen_flags = le32dec(che.zlen);
```

`callback_free` consequently calls `free(rec)` for a record inside the bulk allocation, and `chunks_directory_free` later also calls `free(dir)`.

## Impact and possible fix

A deliberately constructed or corrupted local cachedir index can abort the client during a normal dry-run backup. I am not claiming a remote trigger, a security boundary crossing, server contact, or data loss.

The reader could reject persisted `CHDATA_FLAGS` bits as a corrupt directory, or mask the decoded value with `CHDATA_ZLEN` before assigning it to `zlen_flags`.

### LINKED ISSUE #755: [Bug bounty] Relative --csv-file path is resolved after -C and can overwrite an input-directory file
# [Bug bounty] Relative --csv-file path is resolved after -C and can overwrite an input-directory file

I found a reproducible non-security path-resolution bug in the current `master` commit `0bf0b299fed91139044c81cc6fcdc5521c34d235` (`1.0.41-head`). During archive creation, a relative `--csv-file` path is opened after a deferred `-C` changes the process working directory. The CSV is consequently written inside the `-C` directory rather than the invocation directory, and an existing file with that name is silently truncated.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on behalf of Sebastian Albrecht. The source path, overwrite case, and negative control were independently rerun against an unmodified stock build before submission. Sebastian is available to discuss the report and answer technical questions.

## Reproduction

Build the current tree, then run from the repository root:

```sh
ROOT="$PWD"
PROOF="$(mktemp -d)"
mkdir -p "$PROOF/start" "$PROOF/input"
printf 'file data\n' > "$PROOF/input/file.txt"
printf 'KEEP THIS FILE\n' > "$PROOF/input/stats.csv"

cd "$PROOF/start"
"$ROOT/tarsnap" --no-default-config --keyfile "$ROOT/tests/fake.keys" \
  -c --dry-run --print-stats -f csv-path-proof \
  --csv-file stats.csv -C "$PROOF/input" file.txt
echo $?

test ! -e "$PROOF/start/stats.csv" && echo 'no CSV in invocation directory'
head -n 1 "$PROOF/input/stats.csv"
```

Observed with the stock `-O2 -g` build:

```text
0
no CSV in invocation directory
Archive name,Total size,Compressed size
```

The pre-existing `input/stats.csv` marker was replaced by the CSV output. As a negative control, running from the same `start` directory without `-C` and with an absolute input operand writes the relative CSV path to `start/no-c.csv` and exits 0.

No Tarsnap service connection is required because the reproduction uses `--dry-run` and `tests/fake.keys`.

## Cause

`OPTION_CSV_FILE` stores the argument with `strdup` in `option_csv_filename`. For a relative operand after `-C`, `write.c` later calls `do_chdir`. Only when the archive is closed does `multitape_write.c` open the saved name:

```c
if (csv && (output = fopen(d->csv_filename, "w")) == NULL)
```

The relative path is therefore interpreted in the changed directory. This contrasts with the explicit pre-processing of cachedir paths because the tar code can change directories. The manual describes `-C` as changing directory before adding the following files; it does not say that the global CSV output path is rebased.

## Impact and possible fix

Combining two documented options can silently write the statistics file to an unexpected directory and overwrite an unrelated same-named file there. I am not claiming a remote trigger, security boundary crossing, server impact, or archive-data loss.

The CSV output path could be made absolute relative to the invocation directory before any deferred `-C`, or the output stream could be opened before directory changes occur.

### LINKED ISSUE #756: [Bug bounty] tarsnap-keygen accepts an empty --keyfile path and prompts for a password before failing
# [Bug bounty] tarsnap-keygen accepts an empty --keyfile path and prompts for a password before failing

I found a reproducible non-security input-validation bug in the current `master` commit `0bf0b299fed91139044c81cc6fcdc5521c34d235` (`1.0.41-head`). `tarsnap-keygen` treats `--keyfile=` as satisfying its required key-file option even though an empty string cannot name a file. It consequently prompts for the account password before failing at `open("")` with an unclear empty filename.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on behalf of Sebastian Albrecht. The source path, positive case, negative controls, and absence of network calls were independently rerun against an unmodified stock build before submission. Sebastian is available to discuss the report and answer technical questions.

## Reproduction

Build the current tree, then run:

```sh
./tarsnap-keygen --user u --machine m
# usage: missing --keyfile is rejected immediately

./tarsnap-keygen --user u --machine= --keyfile=/tmp/k
# Machine name must be non-empty

printf 'dummy\n' | ./tarsnap-keygen \
  --user u --machine m --keyfile=
echo $?
```

Observed with a stock `-O2 -g` build:

```text
tarsnap-keygen: Cannot create : No such file or directory
1
```

Tracing the positive case shows `openat(AT_FDCWD, "", O_WRONLY|O_CREAT|O_EXCL, 0600) = -1 ENOENT` and no network system calls. The failure therefore occurs before `keygen_network_register` and does not contact the Tarsnap service.

## Cause

`keygen/keygen.c` initializes `keyfilename` to `NULL`, assigns `optarg` for `--keyfile`, and later checks only `keyfilename == NULL`. The option parser supplies a non-NULL empty string for `--keyfile=`. `keygen_actual.c` explicitly rejects empty user and machine names but has no corresponding check for the required key-file path; it reads the password and then passes the empty string to `keyfile_write_open`.

`tarsnap-keyregen` shares the same `keygen_actual` path and accepts an empty new key-file path in the same way.

## Impact and possible fix

This is a local, deliberate-input usability bug: it unnecessarily requests a sensitive password and then emits an ambiguous filename-less error. I am not claiming a crash, network interaction, security boundary crossing, or data loss.

Rejecting `keyfilename[0] == '\0'` with a clear option error before password entry would make empty and missing required paths consistent with the existing non-empty checks for user and machine names.

### LINKED ISSUE #757: [Bug bounty follow-up] tarsnap-recrypt can commit old-machine deletes while retaining a stale chunk directory
# [Bug bounty follow-up] tarsnap-recrypt can commit old-machine deletes while retaining a stale chunk directory

I found a reproducible non-security crash-consistency gap in current `master` commit `0bf0b299fed91139044c81cc6fcdc5521c34d235` (`1.0.41-head`), after the fix for #746 / PR #747. At the end of `tarsnap-recrypt`, the old-machine delete transaction can be committed without any post-delete chunk directory being staged. An interruption or `unlink` failure immediately after that commit leaves a current sequence number but an old local directory which still claims that the deleted chunks exist.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on behalf of Sebastian Albrecht. The source path, commit state, subsequent deduplication decision, and fix control were independently rerun locally. Sebastian is available to discuss the report and answer technical questions.

## Cause

The success tail in `recrypt/recrypt.c` performs:

```c
storage_delete_end(SD);
multitape_commit(ocachedir, omachinenum, oseqnum, 1,
    &storage_modified);
unlink(odirpath);
```

Unlike a normal chunk deletion, this path never calls `chunks_delete_end` or `chunks_directory_write(..., ".tmp")`. `multitape_commit` nevertheless creates `commit_m` and calls `chunks_transaction_commit`. `chunks_directory_commit` treats an absent `directory.tmp` as success, so the storage delete is committed, `cseq` is updated, and `commit_m` is removed while the live `directory` is unchanged. The following `unlink` is the first operation which removes that stale presence database.

This differs from #746: that issue staged a new directory entry which could be lost across power failure before the marker became durable. Here no post-delete staged directory is created at all, so the current `fsyncdir` fix cannot establish the missing local state.

## Focused end-to-end validation

I compiled the unmodified production implementations of `multitape_commit`, the chunk transaction/directory functions, and `chunks_write_start` / `chunks_write_chunk` into one harness. Only external storage calls were deterministic stubs. The positive case begins with one valid local chunk record and no `directory.tmp`; the storage-commit stub records a successful old-key delete and removes the remote chunk. Execution then continues directly into the next production chunk-write decision without performing the final `unlink`:

```text
case=missing-tmp-positive commit_rc=0 commit_calls=1 whichkey=1 storage_modified=1 cseq=1 commit_m=0 directory=1 remote_present_after_commit=0 local_present=1 write_result=5 storage_write_calls=0
```

The stale local record causes `chunks_write_chunk` to return its cached compressed length without calling `storage_write_file`.

As a fix control, the same run stages an empty `directory.tmp` before committing:

```text
case=empty-tmp-control commit_rc=0 commit_calls=1 whichkey=1 storage_modified=1 cseq=1 commit_m=0 directory=1 remote_present_after_commit=0 local_present=0 write_result=14 storage_write_calls=1
```

The harness exits 0 and was compiled with `-O2 -g -Wall -Wextra -Werror` against the current tree.

## Impact, preconditions, and fix direction

If the process stops after the delete commit returns but before the old directory is removed, a later backup using the old key and cachedir can suppress uploading a chunk which the delete commit removed from the server. Normal startup has no marker to recover and sees a current `cseq`.

The manual instructs users to disable automatic archiving while recrypt runs and says an interrupted recrypt can be rerun. Rerunning it before any old-key backup avoids the demonstrated effect; this is an important precondition. I did not use the production Tarsnap service or perform an archive/restore test, and I am not claiming the USD 500 data-loss tier.

One fix is to stage a valid empty post-delete `directory.tmp` before the old-machine commit, then let the existing transaction code install it. The control above shows that this makes the subsequent write path upload the missing chunk instead of trusting stale local presence state.

### LINKED ISSUE #759: [Bug bounty] --newer with a large relative offset segfaults in unchecked localtime()/gmtime()
I am OpenAI Codex, an LLM reporting this issue on behalf of Sebastian Albrecht. Sebastian is available here to discuss the report and any code changes.

Current Tarsnap `master` (`0bf0b299fed91139044c81cc6fcdc5521c34d235`, `1.0.41-head`) segfaults when a large relative-seconds value in a `--newer` date is followed by a relative month or weekday.

On x86-64 Linux with glibc 2.39, build Tarsnap normally and create any input file, then run:

```sh
./tarsnap --no-default-config --dry-run -c -f a \
  --newer '+1000000000000000000 seconds 1 month' input

./tarsnap --no-default-config --dry-run -c -f a \
  --newer '+1000000000000000000 seconds monday' input
```

Both commands terminate with `SIGSEGV` (exit 139) in a stock optimized build. An ASan/UBSan build locates the two dereferences at `tar/getdate.c:785` in `RelativeMonth()` and `tar/getdate.c:767` in `RelativeDate()`.

Negative controls both exit 0:

```sh
./tarsnap --no-default-config --dry-run -c -f a --newer '1 month' input
./tarsnap --no-default-config --dry-run -c -f a \
  --newer '+1000000000000000000 seconds' input
```

`get_date()` adds `gds->RelSeconds` to `Start` before calling the helpers. The resulting time is outside the broken-down range representable by glibc's `struct tm`, so `localtime()` or `gmtime()` returns `NULL`; each helper dereferences the result without checking it.

This is a local operator/configuration-input crash. I found no remote trigger, privilege boundary, or data-loss impact and am reporting it as a non-security bug.

A robust fix should check each libc conversion and propagate an explicit failure status back through `RelativeMonth()` / `RelativeDate()` to `get_date()` before modifying `Start`. Merely returning `-1` from the current helpers is insufficient because their return values are currently added unconditionally.

### LINKED ISSUE #760: [Bug bounty] --fsck can delete live chunks after chunks_read_cache() allocation failure
I am OpenAI Codex, an LLM reporting this issue on behalf of Sebastian Albrecht. Sebastian is available here to discuss the report and any code changes.

Current Tarsnap `master` (`0bf0b299fed91139044c81cc6fcdc5521c34d235`) can treat an internal cache-allocation failure as a successful archive scan during `--fsck`, then delete still-referenced chunks in phase 5.

In `tar/multitape/multitape_chunkiter.c`, the return value is not assigned to `rc`:

```c
if ((rc = func(cookie, ch)) != 0)
    goto err3;
...
if (chunks_read_cache(CR, ch->hash))
    goto err3;
```

At the second branch, `rc` is necessarily 0 because a nonzero callback result would already have jumped to `err3`. Therefore, an allocation failure in `storage_read_cache_add_name()` aborts the iteration but `multitape_chunkiter_tmd()` returns success.

I reproduced the complete consequence with the unmodified production `phase4()` / `phase5()`, chunk iterator, chunk reader, storage cache, chunk statistics, and deletion walk. The deterministic remote-storage boundary contained one index chunk referencing two live data chunks. Failing only the `struct read_file_cached` allocation produced:

```text
phase4_rc=0
phase5_rc=0
delete_count=2
deleted_a=1
deleted_b=1
```

Both deletion records came from production `chunks_fsck_deletechunks()` calling `storage_delete_file(..., 'c', hash)`. The identical no-fault control deleted zero chunks. A fix-control build differing only by `rc = -1` returned `phase4_rc=-1`, did not enter phase 5, and deleted zero chunks.

The real trigger is local memory pressure, not attacker input, so I am reporting this as a non-security data-loss bug. A live service was not used and no remote user data was touched.

Suggested fix:

```c
if (chunks_read_cache(CR, ch->hash)) {
    rc = -1;
    goto err3;
}
```

### LINKED ISSUE #764: [Bug bounty] Chunk-cache zlen corruption is copied into new archive indexes
# [Bug bounty] Chunk-cache `zlen` corruption is copied into new archive indexes

Tested on current `master` commit
`0bf0b299fed91139044c81cc6fcdc5521c34d235` (Tarsnap 1.0.41-head).

behalf of Sebastian Albrecht. The current source, one-bit POS/NEG control,
stock-client path, and connected restore result were independently rerun
locally. Sebastian is available to discuss the report and answer technical
questions.

## Summary

A fresh chunkification-cache hit validates that each cached hash exists in
the chunk directory, but it does not compare the cached chunk header's
`len` or `zlen` with the canonical values in that directory.

Consequently, a single-bit corruption of cached `zlen` remains structurally
valid and is copied verbatim into the new archive's chunk index. The backup
finishes successfully without rereading the file. On extraction, Tarsnap
requests the existing chunk using the corrupted compressed length and
rejects it as corrupt because the stored chunk has the canonical length.

This turns corruption of a performance-only cache into a silently accepted,
unrestorable archive entry.

## Root cause

1. [`ccache_entry_lookup`](https://github.com/Tarsnap/tarsnap/blob/0bf0b299fed91139044c81cc6fcdc5521c34d235/tar/ccache/ccache_entry.c#L256-L281)
   treats the entry as fresh based on inode, size, and mtime, then asks only
   whether each cached hash is present.
2. [`writetape_ischunkpresent`](https://github.com/Tarsnap/tarsnap/blob/0bf0b299fed91139044c81cc6fcdc5521c34d235/tar/multitape/multitape_write.c#L641-L652)
   checks the hash but returns the untrusted cached `len`; it never compares
   cached `len` or `zlen` with the canonical chunk-directory record.
3. [`ccache_entry_write`](https://github.com/Tarsnap/tarsnap/blob/0bf0b299fed91139044c81cc6fcdc5521c34d235/tar/ccache/ccache_entry.c#L372-L414)
   passes the cached header to `writetape_writechunk`.
4. [`writetape_writechunk`](https://github.com/Tarsnap/tarsnap/blob/0bf0b299fed91139044c81cc6fcdc5521c34d235/tar/multitape/multitape_write.c#L680-L697)
   references the chunk by hash and writes the entire cached 40-byte header
   into the new chunk-index stream.
5. On restore, [`chunks_read_chunk`](https://github.com/Tarsnap/tarsnap/blob/0bf0b299fed91139044c81cc6fcdc5521c34d235/tar/chunks/chunks_read.c#L101-L126)
   passes that `zlen` to `storage_read_file`; the storage API explicitly
   returns `corrupt` when the stored length differs.

## Reproduction

`repro.sh` is an offline reproducer. It uses the public `tests/fake.keys`,
creates a real two-chunk cache record for a 400,000-byte file, and changes
exactly one byte/bit in the first cached header (`zlen` 288 to 289). It then:

- runs the unmodified stock client against NEG and POS caches;
- verifies that both dry runs exit 0 and that the production lookup reports a
  complete cache hit;
- sends the exact headers emitted by the production cache-write path into
  the production restore logic with storage's exact-length behavior.

Run it from a built checkout:

```sh
./repro.sh /path/to/tarsnap
```

Observed output:

```text
prepared target_size=400000 chunk0_len=261120 chunk0_zlen=288 chunk1_len=139264 chunk1_zlen=160 pos_chunk0_zlen=289
cache_byte_differences=1
stock_neg_rc=0 stock_pos_rc=0
case=NEG fullentry=1 write_len=400384 emitted_bytes=80 first_len=261120 first_zlen=288 stored_zlen=288 restore_rc=0 data_match=1
case=POS fullentry=1 write_len=400384 emitted_bytes=80 first_len=261120 first_zlen=289 stored_zlen=288 restore_rc=2 data_match=0
```

I also confirmed the POS path in the unmodified stock binary under GDB. The
first call stack is `ccache_entry_write` -> `writetape_writechunk`; the final
eight bytes of the 40-byte header are:

```text
00 fc 03 00 21 01 00 00
```

That is `len=261120`, `zlen=289`. NEG is identical except byte 36 is `20`
(`zlen=288`). In separate `strace` controls, both stock dry runs exit 0 and
only perform the traversal open of the target (`O_RDONLY|O_NONBLOCK`), not an
`O_RDONLY` content read.

## Impact

A local cache bit error can be accepted as a successful new backup even
though the referenced chunk cannot later be restored through that archive
index. The original canonical chunk and its directory metadata remain
intact; the newly written index is what becomes inconsistent.

This is not based on deliberately changing file contents while preserving
inode/size/mtime. The file, hashes, canonical directory, and all other cache
bytes are unchanged. POS and NEG differ only in one bit of cached `zlen`.

The project already treats damaged chunkification-cache state as a data-loss
risk: commit
[`b67f5a6`](https://github.com/Tarsnap/tarsnap/commit/b67f5a6437a82a919854391b37a33d8fb25c5c68)
removes the cache after `--fsck` specifically because the cache is only an
optimization and hardware/filesystem damage to it could otherwise cause data
loss. This defect is the fail-open path when the damage is not preceded by an
`--fsck` run.

## Suggested fix

Before accepting or copying a cached chunk header, retrieve the canonical
`len` and `zlen` associated with its hash from the chunk directory and require
both values to match. On mismatch, warn that the cache entry is corrupt and
reread/rechunk the file (or discard that cache record). Validating again in
`writetape_writechunk` would provide defense in depth.

### LINKED ISSUE #765: [Bug bounty] --newer "1 day ago" selects a future cutoff and creates an empty archive
# `--newer "1 day ago"` selects a time in the *future*, so the archive comes out empty

Classification: silent wrong result on a documented option. `tarsnap -c` exits
0 and stores nothing. One line, fixed upstream in 2012.

## Product and version

- Tarsnap client `1.0.41-head`, `Tarsnap/tarsnap` master at
  `0bf0b299fed91139044c81cc6fcdc5521c34d235` — the defect is on master as of
  today, verified against the GitHub API, `tar/getdate.c` lines 369-373
- File: `tar/getdate.c`, `relunitphrase()`
- Affects `--newer` / `--newer-ctime` (`tar/bsdtar.c:595`) and `--newer-mtime`
  (`tar/bsdtar.c:612`)

## What happens

```text
Three files, all created seconds before the run:

  tarsnap -c --newer-mtime 1 day ago            rc=0  stored=0/3  Warning: Archive contains no files
  tarsnap -c --newer-mtime 7 days ago           rc=0  stored=0/3  Warning: Archive contains no files
  tarsnap -c --newer-mtime 1 hour ago           rc=0  stored=0/3  Warning: Archive contains no files

Spellings that mean the same thing but take a different code path:

  tarsnap -c --newer-mtime now - 1 day          rc=0  stored=3/3
  tarsnap -c --newer-mtime -1 day               rc=0  stored=3/3
  tarsnap -c --newer-mtime yesterday            rc=0  stored=3/3

Control, no time filter:

  tarsnap -c                                    rc=0  stored=3/3
```

`<count> <unit> ago` is the ordinary way to express an incremental backup, and
it is the one spelling that does not work. The exit status is 0 and the only
signal is a warning on stderr, so a cron job or a script that checks the exit
status sees a successful backup that contains nothing.

Full transcript in `impact.txt`.

## Cause

`relunitphrase()` handles `"<count> <unit>"` in `tar/getdate.c:365-374`:

```c
	if (gds->tokenp[0].token == tUNUMBER
	    && gds->tokenp[1].token == tSEC_UNIT) {
		/* "1 day" */
		gds->HaveRel++;
		gds->RelSeconds += gds->tokenp[1].value * gds->tokenp[2].value;
		gds->tokenp += 3;
		return 1;
	}
```

The phrase is two tokens long, but the body is a copy of the three-token
`'-' tUNUMBER tSEC_UNIT` branch above it with the sign token removed and the
indices left alone. Two consequences:

**The count is discarded.** It multiplies `tokenp[1]` (the unit) by
`tokenp[2]` (whatever comes *after* the unit) instead of `tokenp[0]` (the
count) by `tokenp[1]` (the unit).

**`ago` is swallowed.** Advancing by three consumes the token after the unit.
For `"1 day ago"` that token is `tAGO`, so the negation in `phrase()`
(`tar/getdate.c:461-466`) never runs:

```c
	if (relunitphrase(gds)) {
		if (gds->tokenp[0].token == tAGO) {
			gds->RelSeconds = -gds->RelSeconds;
			gds->RelMonth = -gds->RelMonth;
			gds->tokenp += 1;
		}
		return 1;
	}
```

`tAGO` carries the value `1` in the word table (`tar/getdate.c:650`,
`{ 0, "ago", tAGO, 1 }`). So `"<n> hours ago"` evaluates to
`3600 * 1 = +3600` for every `n` — the count is gone, the sign is inverted,
and the magnitude is always exactly one unit.

Without `ago`, `tokenp[2]` is the zeroed terminator slot, so
`"<n> hours"` evaluates to `3600 * 0 = 0`.

The structurally identical `tMONTH_UNIT` branch twenty lines further down is
correct, which is what makes the defect easy to see once you look:

```c
	if (gds->tokenp[0].token == tUNUMBER
	    && gds->tokenp[1].token == tMONTH_UNIT) {
		/* "2 years" */
		gds->HaveRel++;
		gds->RelMonth += gds->tokenp[0].value * gds->tokenp[1].value;
		gds->tokenp += 2;
		return 1;
	}
```

## Measured at the parser

`getdate_probe.c` includes the production `tar/getdate.c` unmodified and calls
`get_date()` against a fixed instant, so the numbers do not depend on when it
runs. Offsets are seconds relative to `now`:

```text
expression               expected         actual
1 second ago                   -1              1 <-- WRONG
30 seconds ago                -30              1 <-- WRONG
1 minute ago                  -60             60 <-- WRONG
5 minutes ago                -300             60 <-- WRONG
1 hour ago                  -3600           3600 <-- WRONG
24 hours ago               -86400           3600 <-- WRONG
1 day ago                  -86400          86400 <-- WRONG
7 days ago                -604800          86400 <-- WRONG
2 weeks ago              -1209600         604800 <-- WRONG
1 hour                       3600              0 <-- WRONG
3 hours                     10800              0 <-- WRONG
2 days                     172800              0 <-- WRONG
-1 hour                     -3600          -3600      explicit sign
-7 days                   -604800        -604800      explicit sign
+1 hour                      3600           3600      explicit sign
yesterday                  -86400         -86400      bare tSEC_UNIT
tomorrow                    86400          86400      bare tSEC_UNIT
now                             0              0      bare tSEC_UNIT

12 of 18 expressions parse to the wrong instant.
```

The count really is ignored:

```text
  1 hours ago      -> now +3600
  2 hours ago      -> now +3600
  24 hours ago     -> now +3600
  1000 hours ago   -> now +3600
  999999 hours ago -> now +3600
```

And the over-advance is visible directly:

```text
  1 hour               -> accepted, now +0
  1 hour zzzz          -> rejected
  1 hour ago           -> accepted, now +3600
  1 hour 30 minutes    -> accepted, now +108060
```

`1 hour 30 minutes` is the mechanism in the open: `3600 * 30 = 108000` from
the broken branch, then `tokenp += 3` lands on the trailing `minutes`, which
the bare-`tSEC_UNIT` branch adds as another `60`. 108060, not 5400. I worked
that number out from the source before running it, which is what convinced me
the reading is right.

Only the three-token `'-'`/`'+'` forms and the bare-`tSEC_UNIT` words go
through other branches, which is why `-1 day`, `now - 1 day` and `yesterday`
still work.

## Provenance — upstream fixed this in 2012

This is not something tarsnap wrote. It arrived with the file and upstream
repaired it afterwards; tarsnap's copy never picked the repair up.

- **Introduced**: libarchive `438c6911e5fc`, 2009-02-22, *"No more YACC:
  Translated the old getdate.y into a recursive-descent parser."* The
  translated `relunitphrase()` has exactly the line above.
- **Fixed**: libarchive `0cc23cefe50e6d7f4e5dff1b794a99a6a547de99`,
  2012-12-01, *"'now - 2 hours' should be the same as '2 hours ago' — This was
  broken by a mis-parsing of the '2 hours' phrase."* The diff is precisely the
  two lines below, and the same commit added the regression test
  `assertEqualInt(get_date(now, "now - 2 hours"), get_date(now, "2 hours ago"));`
- **Upstream today**: the file is now `libarchive/archive_parse_date.c` and
  still carries the fixed form.
- **Tarsnap today**: `tar/getdate.c` on master still has the 2009 form.

Details and the verbatim upstream diff in `upstream-history.txt`.

On scope: `tar/getdate.c` is part of tarsnap's own bsdtar tree, not the
bundled `tar/libarchive/` copy — the same source file currently discussed in
open issue #759. I found and reproduced this behavior independently before
searching upstream history; the 2012 fix surfaced only during the subsequent
prior-art review. I am flagging the upstream origin because it is useful, not
because I think it disqualifies the report; but it is your call.

## Suggested fix

The upstream change, unaltered:

```diff
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -365,8 +365,8 @@ relunitphrase(struct gdstate *gds)
 	    && gds->tokenp[1].token == tSEC_UNIT) {
 		/* "1 day" */
 		gds->HaveRel++;
-		gds->RelSeconds += gds->tokenp[1].value * gds->tokenp[2].value;
-		gds->tokenp += 3;
+		gds->RelSeconds += gds->tokenp[0].value * gds->tokenp[1].value;
+		gds->tokenp += 2;
 		return 1;
 	}
```

A regression test in the same spirit as upstream's, expressed against the two
spellings that tarsnap users actually type:

```c
	assertEqualInt(get_date(now, "now - 2 hours"),
	    get_date(now, "2 hours ago"));
	assertEqualInt(get_date(now, "-1 day"),
	    get_date(now, "1 day ago"));
```

Both fail before the change and pass after it.

It is worth diffing the rest of `tar/getdate.c` against
`libarchive/archive_parse_date.c` while you are in there — fourteen years of
upstream fixes have not been merged, and report 05 (issue #759, the unchecked
`localtime()` calls in the same file) is another symptom of the same drift.

## Reproducing

```sh
cc -g -O1 -fsanitize=address,undefined -I tar -I . \
   getdate_probe.c -o getdate_probe
TZ=UTC ./getdate_probe
```

`getdate_probe.c` is in this directory; `probe_output.txt` is its output.
The end-to-end runs in `impact.txt` used a normal `--dry-run` build against
`tests/fake.keys`, no server required.

## How it was found

A clang `-fsanitize=integer` build (the integer checks — implicit truncation,
sign change, unsigned overflow — that GCC's `-fsanitize=undefined` does not
provide) reported a signed overflow at `tar/getdate.c:869` on
`--newer 99999999999999999999`. That is a separate and much smaller issue, but
chasing it meant measuring what the parser actually returns for ordinary
inputs, which is how this surfaced. The overflow itself is reported
separately.

wasim-builds:
This PR fixes 9 more bugs from the bug bounty program:

- tar/getdate.c: fix relunitphrase count and sign inversion (Fixes #765)
- tar/chunks/: verify zlen against cache directory (Fixes #764)
- tar/multitape/multitape_chunkiter.c: check chunks_read_cache error (Fixes #760)
- tar/getdate.c: check localtime/gmtime for NULL (Fixes #759)
- recrypt/recrypt.c: stage empty directory before old-machine commit (Fixes #757)
- keygen/keygen.c, keyregen/keyregen.c: reject empty keyfile path (Fixes #756)
- tar/bsdtar.c: make csv_filename absolute before -C (Fixes #755)
- tar/chunks/chunks_directory.c: mask CHDATA_ZLEN flags (Fixes #754)
- lib/crypto/crypto_keys_subr.c: handle zero-length BN safely (Fixes #753)

### DIFF
--- a/keygen/keygen.c
+++ b/keygen/keygen.c
@@ -129,7 +129,8 @@ main(int argc, char **argv)
 	(void)argv; /* argv is not used beyond this point. */
 
 	/* We must have a user name, machine name, and key file specified. */
-	if ((C.user == NULL) || (C.name == NULL) || (keyfilename == NULL))
+	if ((C.user == NULL) || (C.name == NULL) || (keyfilename == NULL) ||
+	    (keyfilename[0] == '\0'))
 		usage();
 
 	/*
--- a/keyregen/keyregen.c
+++ b/keyregen/keyregen.c
@@ -143,7 +143,8 @@ main(int argc, char **argv)
 	 * file specified.
 	 */
 	if ((C.user == NULL) || (C.name == NULL) ||
-	    (keyfilename == NULL) || (oldkeyfilename == NULL))
+	    (keyfilename == NULL) || (keyfilename[0] == '\0') ||
+	    (oldkeyfilename == NULL) || (oldkeyfilename[0] == '\0'))
 		usage();
 
 	/*
--- a/lib/crypto/crypto_keys_subr.c
+++ b/lib/crypto/crypto_keys_subr.c
@@ -55,7 +55,7 @@ import_BN(BIGNUM ** bn, const uint8_t ** buf, size_t * buflen)
 	*buflen -= sizeof(uint32_t);
 
 	/* Sanity check. */
-	if (len > INT_MAX) {
+	if ((len == 0) || (len > INT_MAX)) {
 		warn0("Unexpected key length");
 		goto err0;
 	}
@@ -143,7 +143,7 @@ export_BN(const BIGNUM * bn, uint8_t ** buf, size_t * buflen,
 	BN_bn2bin(bn, *buf);
 
 	/* Convert to little-endian format. */
-	for (i = 0; i < bnlen - 1 - i; i++) {
+	for (i = 0; i < bnlen / 2; i++) {
 		(*buf)[i] ^= (*buf)[bnlen - 1 - i];
 		(*buf)[bnlen - 1 - i] ^= (*buf)[i];
 		(*buf)[i] ^= (*buf)[bnlen - 1 - i];
--- a/recrypt/recrypt.c
+++ b/recrypt/recrypt.c
@@ -683,6 +683,12 @@ main(int argc, char **argv)
 		exit(1);
 	}
 
+	/* Stage an empty directory before we commit the deletion. */
+	if (chunks_transaction_empty(ocachedir)) {
+		warnp("Cannot stage empty chunk directory");
+		exit(1);
+	}
+
 	/* Commit the delete transaction and delete the old chunk dir. */
 	printf("Committing block deletes...");
 	if (multitape_commit(ocachedir, omachinenum, oseqnum, 1,
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -1152,6 +1152,23 @@ main(int argc, char **argv)
 	}
 
 	/*
+	 * Make the CSV filename absolute if necessary, since the tar code
+	 * can change directories.
+	 */
+	if (bsdtar->option_csv_filename != NULL &&
+	    bsdtar->option_csv_filename[0] != '/') {
+		char cwd[PATH_MAX];
+		if (getcwd(cwd, PATH_MAX) == NULL)
+			bsdtar_errc(bsdtar, 1, errno, "getcwd");
+		char *newpath = malloc(strlen(cwd) + strlen(bsdtar->option_csv_filename) + 2);
+		if (newpath == NULL)
+			bsdtar_errc(bsdtar, 1, errno, "Out of memory");
+		sprintf(newpath, "%s/%s", cwd, bsdtar->option_csv_filename);
+		free(bsdtar->option_csv_filename);
+		bsdtar->option_csv_filename = newpath;
+	}
+
+	/*
 	 * Canonicalize the path to the cache directories.  This is
 	 * necessary since the tar code can change directories.
 	 */
--- a/tar/ccache/ccache_entry.c
+++ b/tar/ccache/ccache_entry.c
@@ -326,10 +326,10 @@ ccache_entry_lookup(CCACHE * cache, const char * path, const struct stat * sb,
 		if (rc != Z_OK) {
 			free(cce->trailer);
 			cce->trailer = NULL;
+		} else {
+			/* We can supply the trailer data from the cache. */
+			skiplen += cce->ccr->tlen;
 		}
-
-		/* We can supply the trailer data from the cache. */
-		skiplen += cce->ccr->tlen;
 	} else {
 		cce->trailer = NULL;
 	}
--- a/tar/chunks/chunks.h
+++ b/tar/chunks/chunks.h
@@ -65,12 +65,12 @@ ssize_t chunks_write_chunk(CHUNKS_W *, const uint8_t *, const uint8_t *,
 int chunks_write_ispresent(CHUNKS_W *, const uint8_t *);
 
 /**
- * chunks_write_chunkref(C, hash):
- * If a chunk with hash ${hash} exists, mark it as being part of the write
- * transaction associated with the cookie ${C} and return 0.  If it
- * does not exist, return 1.
+ * chunks_write_chunkref(C, hash, len, zlen):
+ * If a chunk with hash ${hash}, length ${len}, and compressed length ${zlen}
+ * exists, mark it as being part of the write transaction associated with
+ * the cookie ${C} and return 0.  If it does not exist, return 1.
  */
-int chunks_write_chunkref(CHUNKS_W *, const uint8_t *);
+int chunks_write_chunkref(CHUNKS_W *, const uint8_t *, size_t, size_t);
 
 /**
  * chunks_write_extrastats(C, len):
@@ -222,6 +222,12 @@ int chunks_stats_printarchive(FILE *, CHUNKS_S *, const char *, int, int);
  * No more calls will be made to chunks_stats* functions.
  */
 void chunks_stats_free(CHUNKS_S *);
+
+/**
+ * chunks_transaction_empty(cachepath):
+ * Stage an empty chunk directory transaction.
+ */
+int chunks_transaction_empty(const char *);
 
 /**
  * chunks_transaction_checkpoint(cachepath):
--- a/tar/chunks/chunks_directory.c
+++ b/tar/chunks/chunks_directory.c
@@ -249,7 +249,7 @@ chunks_directory_read(const char * cachepath, void ** dir,
 		/* ... creating struct chunkdata records... */
 		memcpy(p->hash, che.hash, 32);
 		p->len = le32dec(che.len);
-		p->zlen_flags = le32dec(che.zlen);
+		p->zlen_flags = le32dec(che.zlen) & CHDATA_ZLEN;
 		p->nrefs = le32dec(che.nrefs);
 		p->ncopies = le32dec(che.ncopies);
 
--- a/tar/chunks/chunks_write.c
+++ b/tar/chunks/chunks_write.c
@@ -267,13 +267,13 @@ chunks_write_ispresent(CHUNKS_W * C, const uint8_t * hash)
 }
 
 /**
- * chunks_write_chunkref(C, hash):
- * If a chunk with hash ${hash} exists, mark it as being part of the write
- * transaction associated with the cookie ${C} and return 0.  If it
- * does not exist, return 1.
+ * chunks_write_chunkref(C, hash, len, zlen):
+ * If a chunk with hash ${hash}, length ${len}, and compressed length ${zlen}
+ * exists, mark it as being part of the write transaction associated with
+ * the cookie ${C} and return 0.  If it does not exist, return 1.
  */
 int
-chunks_write_chunkref(CHUNKS_W * C, const uint8_t * hash)
+chunks_write_chunkref(CHUNKS_W * C, const uint8_t * hash, size_t len, size_t zlen)
 {
 	struct chunkdata * ch;
 
@@ -282,6 +282,8 @@ chunks_write_chunkref(CHUNKS_W * C, const uint8_t * hash)
 	 * transaction and return 0.
 	 */
 	if ((ch = rwhashtab_read(C->HT, hash)) != NULL) {
+		if (ch->len != len || (ch->zlen_flags & CHDATA_ZLEN) != zlen)
+			return (1);
 		chunks_stats_add(&C->stats_total, ch->len,
 		    ch->zlen_flags & CHDATA_ZLEN, 1);
 		chunks_stats_add(&C->stats_tape, ch->len,
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -719,7 +719,7 @@ Convert(time_t Month, time_t Day, time_t Year,
 	    ? 29 : 28;
 	/* Checking for 2038 bogusly assumes that time_t is 32 bits.  But
 	   I'm too lazy to try to check for time_t overflow in another way.  */
-	if (Year < EPOCH || Year >= 2038
+	if (Year < EPOCH || (sizeof(time_t) <= 4 && Year >= 2038)
 	    || Month < 1 || Month > 12
 	    /* Lint fluff:  "conversion from long may lose accuracy" */
 	    || Day < 1 || Day > DaysInMonth[(int)--Month])
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -53,7 +53,7 @@ enum DSTMODE { DSTon, DSToff, DSTmaybe };
 enum { tAM, tPM };
 /* Token types returned by nexttoken() */
 enum { tAGO = 260, tDAY, tDAYZONE, tAMPM, tMONTH, tMONTH_UNIT, tSEC_UNIT,
-       tUNUMBER, tZONE, tDST };
+       tUNUMBER, tZONE, tDST, tERROR };
 struct token { int token; time_t value; };
 
 /*
@@ -867,9 +867,24 @@ nexttoken(char **in, time_t *value)
 		 * don't deal with signed numbers here.
 		 */
 		if (isdigit((unsigned char)(c = **in))) {
-			for (*value = 0; isdigit((unsigned char)(c = *(*in)++)); )
-				*value = 10 * *value + c - '0';
+			unsigned long long val = 0;
+			int overflow = 0;
+			unsigned long long tmax = ~0ULL;
+			if (sizeof(time_t) == 4)
+				tmax = ((time_t)-1 > 0) ? 0xFFFFFFFFULL : 0x7FFFFFFFULL;
+			else if (sizeof(time_t) == 8)
+				tmax = ((time_t)-1 > 0) ? 0xFFFFFFFFFFFFFFFFULL : 0x7FFFFFFFFFFFFFFFULL;
+
+			for (; isdigit((unsigned char)(c = *(*in)++)); ) {
+				if (val > (~0ULL - (c - '0')) / 10)
+					overflow = 1;
+				else
+					val = 10 * val + c - '0';
+			}
 			(*in)--;
+			if (overflow || val > tmax)
+				return (tERROR);
+			*value = (time_t)val;
 			return (tUNUMBER);
 		}
 
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -368,8 +368,8 @@ relunitphrase(struct gdstate *gds)
 	    && gds->tokenp[1].token == tSEC_UNIT) {
 		/* "1 day" */
 		gds->HaveRel++;
-		gds->RelSeconds += gds->tokenp[1].value * gds->tokenp[2].value;
-		gds->tokenp += 3;
+		gds->RelSeconds += gds->tokenp[0].value * gds->tokenp[1].value;
+		gds->tokenp += 2;
 		return 1;
 	}
 	if (gds->tokenp[0].token == '-'
@@ -754,26 +754,30 @@ DSTcorrect(time_t Start, time_t Future)
 }
 
 
-static time_t
-RelativeDate(time_t Start, time_t zone, int dstmode,
+static int
+RelativeDate(time_t *Start, time_t zone, int dstmode,
     time_t DayOrdinal, time_t DayNumber)
 {
 	struct tm	*tm;
 	time_t	t, now;
 
-	t = Start - zone;
+	t = *Start - zone;
 	tm = gmtime(&t);
-	now = Start;
+	if (tm == NULL)
+		return -1;
+	now = *Start;
 	now += DAY * ((DayNumber - tm->tm_wday + 7) % 7);
 	now += 7 * DAY * (DayOrdinal <= 0 ? DayOrdinal : DayOrdinal - 1);
 	if (dstmode == DSTmaybe)
-		return DSTcorrect(Start, now);
-	return now - Start;
+		*Start += DSTcorrect(*Start, now);
+	else
+		*Start += now - *Start;
+	return 0;
 }
 
 
-static time_t
-RelativeMonth(time_t Start, time_t Timezone, time_t RelMonth)
+static int
+RelativeMonth(time_t *Start, time_t Timezone, time_t RelMonth)
 {
 	struct tm	*tm;
 	time_t	Month;
@@ -781,14 +785,17 @@ RelativeMonth(time_t Start, time_t Timezone, time_t RelMonth)
 
 	if (RelMonth == 0)
 		return 0;
-	tm = localtime(&Start);
+	tm = localtime(Start);
+	if (tm == NULL)
+		return -1;
 	Month = 12 * (tm->tm_year + 1900) + tm->tm_mon + RelMonth;
 	Year = Month / 12;
 	Month = Month % 12 + 1;
-	return DSTcorrect(Start,
+	*Start += DSTcorrect(*Start,
 	    Convert(Month, (time_t)tm->tm_mday, Year,
 		(time_t)tm->tm_hour, (time_t)tm->tm_min, (time_t)tm->tm_sec,
 		Timezone, DSTmaybe));
+	return 0;
 }
 
 /*
@@ -1027,14 +1034,15 @@ get_date(time_t now, char *p)
 
 	/* Add the relative offset. */
 	Start += gds->RelSeconds;
-	Start += RelativeMonth(Start, gds->Timezone, gds->RelMonth);
+	if (RelativeMonth(&Start, gds->Timezone, gds->RelMonth) != 0)
+		return -1;
 
 	/* Adjust for day-of-week offsets. */
 	if (gds->HaveWeekDay
 	    && !(gds->HaveYear || gds->HaveMonth || gds->HaveDay)) {
-		tod = RelativeDate(Start, gds->Timezone,
-		    gds->DSTmode, gds->DayOrdinal, gds->DayNumber);
-		Start += tod;
+		if (RelativeDate(&Start, gds->Timezone,
+		    gds->DSTmode, gds->DayOrdinal, gds->DayNumber) != 0)
+			return -1;
 	}
 
 	/* -1 is an error indicator, so return 0 instead of -1 if
--- a/tar/multitape/multitape_chunkiter.c
+++ b/tar/multitape/multitape_chunkiter.c
@@ -93,8 +93,10 @@ multitape_chunkiter_tmd(STORAGE_R * S, CHUNKS_S * C,
 		}
 
 		/* We want to cache this chunk after reading it. */
-		if (chunks_read_cache(CR, ch->hash))
+		if (chunks_read_cache(CR, ch->hash)) {
+			rc = -1;
 			goto err3;
+		}
 
 		/* Read the chunk into buffer. */
 		if ((rc = chunks_read_chunk(CR, ch->hash, chunklen, chunkzlen,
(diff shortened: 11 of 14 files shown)


gperciva:
Don't mix unrelated bug fixes in the same PR.

Issue #716, opened 2026-07-14
Bug Bounty Report — Tarsnap client: use-after-free / double-free in archive_write_open_multitape() error path


chengdazhi:
## 1. Summary

Three distinct crash bugs were found in the tarsnap client, all on
memory-allocation-failure paths (the same bug class as the previously reported
spiped `network_connect()` UAF — error unwinding that leaves dangling state):

| # | Location | Bug | Trigger |
|---|----------|-----|---------|
| 1 | `tar/glue/archive_multitape.c:151-153` (+ `tar/write.c:347`, `libarchive/archive_write.c`) | **heap-use-after-free read + multiple double-frees** of the multitape write cookie | `malloc` failure inside `archive_write_open()` during `tarsnap -c` (including `--dry-run`) |
| 2 | `tar/bsdtar.c:183` → `bsdtar.c:958` (or `:883`, `:484`, `:1609`) | **NULL pointer dereference**: return value of `strlist_init()` never checked | `malloc` failure of the ~2nd allocation in `main()` — every tarsnap invocation |
| 3 | `tar/tree.c` (`tree_open` :224, `tree_append` :263, `tree_push` :297, and the unchecked `strdup` in `tree_push` :233) | **`abort()` on malloc failure** / NULL-deref via unchecked `strdup` | `malloc` failure during directory traversal in `tarsnap -c` |

Bug 1 is the most severe: it is a use-after-free followed by double-frees
(heap corruption), not merely a NULL dereference, and it sits in tarsnap's own
glue code. Bugs 2 and 3 are NULL-deref / `abort()` crashes; Bug 3 lives in
bsdtar-derived code (`tar/tree.c`), but the calling code in tarsnap's
`tar/write.c` demonstrably expects a NULL return and handles it, so the
`abort()` is a genuine defect rather than intended behavior.

All three are reachable under real-world memory pressure (cgroup / `ulimit -v`
limits on Linux or FreeBSD — tarsnap's primary deployment targets; macOS lacks
`RLIMIT_AS` enforcement, but the bugs are platform-independent logic errors).
No network access, no key material, and no special privileges are required to
trigger them — the failing allocations all occur during local archive-creation
setup, before (or without) any server contact.

---

### 2.1 Root cause

`tar/glue/archive_multitape.c`:

```c
void *
archive_write_open_multitape(struct archive * a, ...)
{
        struct multitape_write_internal * d;

        /* Clear any error messages from the archive. */
        archive_clear_error(a);

[...]

PR #782, opened 2026-08-24
Archive multitape fix with deferred_free

gperciva:


gperciva:
There are two alternate fixes for a use-after-free:
- #782 adding a `_deferred_free()` to `writetape`, or
- #783 adding a `multitape_write_internal_wrapped` to `archive_multitape.c`

In terms of having a minimal commit, #782 is a much smaller diff.  But in terms of keeping each layer of the program separate, #783 does a much better job.  I implemented both solutions so you can compare them.

Marked as "draft" so that you don't merge one of them accidentally, before deciding which one to keep.

gperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/716

The final commit message might be a bit verbose, but given how long it took me to wrap my head around it, I figured that I should err on the side of over-explaining.  Happy to cut some of it out.

Issue #770, opened 2026-08-23
[Bug bounty] Repeated --list-archives can free stale metadata after stdout fails


mySebbe:
# `statstape_printlist_item()` frees an uninitialised `struct tapemetadata` when writing to stdout fails

Classification: local process crash caused by freeing stale metadata after an
output error in a repeated archive listing. No exploitability claim.

Disclosure: this issue was prepared with OpenAI Codex, an LLM, working on
behalf of Sebastian Albrecht. The single-call negative control and the
release-style repeated-call crash were independently rerun locally. Sebastian
is available to discuss the report and answer technical questions.

## Product and version

- Tarsnap client `1.0.41-head`, current `master` at
  `0bf0b299fed91139044c81cc6fcdc5521c34d235`
- File: `tar/multitape/multitape_stats.c:229-323`
- Call sites: `tar/glue/tape.c:216` and `tar/multitape/multitape_stats.c:349`

## Summary

```c
/* tar/multitape/multitape_stats.c:229-257 */
int
statstape_printlist_item(TAPE_S * d, const uint8_t tapehash[32], int verbose,
    int print_nulls, int print_hash)
{
	struct tapemetadata tmd;          /* <-- uninitialised */
	...
	/* Print archive hash. */
	if (print_hash) {
		hexify(tapehash, hexstr, 32);
		fprintf(stdout, "%s", hexstr);

		if (verbose == 0) {
			/* We're finished; print archive separator and quit. */
			if (print_separator(stdout, "\n", print_nulls, 1))
				goto err1;              /* <-- tmd not yet set */
			goto done;
		} else {
			/* We have more fields; print field separator. */
			if (print_separator(stdout, "\t", print_nulls, 2))
				goto err1;              /* <-- tmd not yet set */
		}
	}

	/* Read the tape metadata. */
	if (multitape_metadata_get_byhash(d->SR, NULL, &tmd, tapehash, 0))
		goto err0;
```

`tmd` is only filled by `multitape_metadata_get_byhash()`, four lines later.
Both `goto err1` above it land here:

```c
/* tar/multitape/multitape_stats.c:318-323 */
err1:
	multitape_metadata_free(&tmd);
err0:
	/* Failure! */
	return (-1);
```

and `multitape_metadata_free()` dereferences and frees three fields:

```c
/* tar/multitape/multitape_metadata.c:441-456 */
	/* Free arguments. */
	for (arg = 0; arg < mdat->argc; arg++)
[...]

PR #784, opened 2026-08-24
statstape_printlist_item: fix error path

gperciva:


gperciva:
This fixes https://github.com/Tarsnap/tarsnap/issues/770 (which has an unusually-verbose description).

PR #788, opened 2026-08-25
fix: treat empty XDG_CONFIG_HOME as unset per XDG spec

### LINKED ISSUE #786: [Bug bounty] Empty XDG_CONFIG_HOME silently skips the per-user config file
## Summary

When `XDG_CONFIG_HOME` is set but **empty**, tarsnap constructs the path `/tarsnap/tarsnap.conf` (filesystem root) instead of falling back to `$HOME/.config/tarsnap/tarsnap.conf`. The user's per-user configuration file (keyfile, cachedir, etc.) is silently skipped.

Per the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/latest/), an empty value should be treated the same as unset. The man page (`tarsnap.conf(5)`) documents fallback when the variable is *unset* but does not address the set-but-empty case, so behaviour contradicts both the spec and user expectations.

Found by AI-assisted code review on behalf of the human operator of this account; locally reproduced on macOS/arm64, tarsnap `1.0.41-head` (`0bf0b299fed91139044c81cc6fcdc5521c34d235`). No duplicate found (#569/#686/#687/#688/#748 cover the feature itself and its documentation).

## Repro (any platform)

```sh
# Unset — correct fallback to ~/.config:
$ env -u XDG_CONFIG_HOME tarsnap --dump-config -c -f x somefile
Reading from config file: /Users/kai/.tarsnaprc
Reading from config file: /Users/kai/.config/tarsnap/tarsnap.conf
Reading from config file: /usr/local/etc/tarsnap.conf

# Set but empty — per-user config silently skipped:
$ env XDG_CONFIG_HOME= tarsnap --dump-config -c -f x somefile
Reading from config file: /Users/kai/.tarsnaprc
Reading from config file: /tarsnap/tarsnap.conf
Reading from config file: /usr/local/etc/tarsnap.conf
```

## Root cause

In `tar/bsdtar.c`, the default-config selection only checks for non-NULL:

```c
if ((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) {
        asprintf(&bsdtar->conffile, "%s/tarsnap/tarsnap.conf", xdg_configdir);
} else if (bsdtar->homedir != NULL) {
        asprintf(&bsdtar->conffile,
            "%s/.config/tarsnap/tarsnap.conf", bsdtar->homedir);
}
```

An empty value produces `/tarsnap/tarsnap.conf`.

## Impact

Users who export an empty `XDG_CONFIG_HOME` (seen in buggy dotfiles) silently lose every option in their per-user config file (keyfile, cachedir, ...), leading to hard-to-diagnose failures of subsequent tarsnap operations.

## Suggested fix

```c
if (((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) &&
    (xdg_configdir[0] != '\0')) {
```

rafaio1:
## Summary

Fixes #786

When `XDG_CONFIG_HOME` is set but **empty**, tarsnap constructed the path `/tarsnap/tarsnap.conf` (filesystem root) instead of falling back to `$HOME/.config/tarsnap/tarsnap.conf`. This silently skipped the per-user configuration file.

## Root Cause

In `tar/bsdtar.c`, the code only checked if `getenv("XDG_CONFIG_HOME")` was non-NULL:

```c
if ((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) {
    asprintf(&bsdtar->conffile, "%s/tarsnap/tarsnap.conf", xdg_configdir);
}
```

An empty string is non-NULL, so it produced `/tarsnap/tarsnap.conf`.

## Fix

Added an additional check for `xdg_configdir[0] != '\0'` so an empty value falls through to the `else if` branch, which uses the standard `\/root/.config/tarsnap/tarsnap.conf` fallback — consistent with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), which states:

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

## Repro

```sh
# Before fix (broken):
$ env XDG_CONFIG_HOME= tarsnap --dump-config -c -f x somefile
Reading from config file: /tarsnap/tarsnap.conf   # <-- wrong, filesystem root

# After fix (correct):
$ env XDG_CONFIG_HOME= tarsnap --dump-config -c -f x somefile
Reading from config file: /Users/user/.config/tarsnap/tarsnap.conf  # <-- correct fallback
```

## Changes

- `tar/bsdtar.c`: One-line change — added `&& xdg_configdir[0] != '\0'` to the condition.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -925,7 +925,7 @@ main(int argc, char **argv)
 		}
 
 		/* Check if ${XDG_CONFIG_HOME} is set. */
-		if ((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) {
+		if ((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL && xdg_configdir[0] != '\0') {
 			/*
 			 * If it exists, use
 			 * ${XDG_CONFIG_HOME}/tarsnap/tarsnap.conf


gperciva:
@rafaio1 can you confirm that you read the AGENTS.md and therefore know that you won't receive anything for this PR? We do not pay for code; only for bug reports.

PR #792, opened 2026-08-26
fix(keygen): reject empty --keyfile path before prompting for password (fixes #756)

### LINKED ISSUE #756: [Bug bounty] tarsnap-keygen accepts an empty --keyfile path and prompts for a password before failing

ranka23:
## Summary

`--keyfile=` yields a non-NULL empty string, which passed the `keyfilename == NULL` check. `tarsnap-keygen` (and `tarsnap-keyregen`, which shares the path) then prompted for the account **password** before failing with the ambiguous `Cannot create : No such file or directory` at `open("")`.

## Fix

Add a sanity check in `keygen_actual()` — alongside the existing user/machine-name checks and *before* the password prompt:

```c
if ((keyfilename == NULL) || (keyfilename[0] == '\0')) {
	fprintf(stderr, "Key file name must be non-empty\\n");
	goto err0;
}
```

## Verification

| Command | Before | After |
|---|---|---|
| `tarsnap-keygen --user u --machine m --keyfile=` | prompts for password, then `Cannot create :`, rc=1 | `Key file name must be non-empty`, rc=1, no prompt |
| `tarsnap-keygen --user u --machine m` (missing) | usage error | unchanged |
| normal keygen flow | ok | unchanged |

Built via `autoreconf -i && ./configure && make`; binary tested directly.

Closes #756

### DIFF
--- a/keygen/keygen_actual.c
+++ b/keygen/keygen_actual.c
@@ -84,6 +84,12 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
 		goto err0;
 	}
 
+	/* Sanity-check the key file name before asking for a password. */
+	if ((keyfilename == NULL) || (keyfilename[0] == '\0')) {
+		fprintf(stderr, "Key file name must be non-empty\n");
+		goto err0;
+	}
+
 	/* Get a password. */
 	if (readpass(&C->passwd, "Enter tarsnap account password", NULL, 0)) {
 		warnp("Error reading password");


keygen/keygen_actual.c
```diff
@@ -84,6 +84,12 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
 		goto err0;
 	}
 
+	/* Sanity-check the key file name before asking for a password. */
```
gperciva, line 87: This should happen above the memory size check.  Also, the comment should just be
```
/* Sanity-check the key file name. */
```

keygen/keygen_actual.c
```diff
@@ -84,6 +84,12 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
...
+	if ((keyfilename == NULL) || (keyfilename[0] == '\0')) {
```
gperciva, line 88: This function seems to use `strlen(foo) == 0` to check for an empty string, so please use that instead of `'\0`.

keygen/keygen_actual.c
```diff
@@ -84,6 +84,12 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
...
+		fprintf(stderr, "Key file name must be non-empty\n");
```
gperciva, line 89: This should be `warn0`.

Admittedly, you were misled by some other checks using `fprintf` instead of `warn0`, but I'm fixing those now.

keygen/keygen_actual.c
gperciva: Also, we already checked for `keyfilename == NULL` in `keygen.c`, so no need to do that here.

gperciva:
Closed due to lack of response.

Issue #785, opened 2026-08-25
[Bug bounty] --newer-mtime "Jun 15 2020" treats the year as 20:20 and silently omits files


mySebbe:
# `--newer-mtime "Jun 15 2020"` treats the year as 20:20 and silently omits files

Classification: non-security correctness bug causing an incomplete backup
selection with exit status 0.

## Product and currentness

- Tarsnap `1.0.41-head`, current `master` rechecked 2026-08-25
  `0bf0b299fed91139044c81cc6fcdc5521c34d235`
- `tar/getdate.c:289-312` (`datephrase`) and `:500-505` (`phrase`)
- Reached by `--newer`, `--newer-ctime`, and `--newer-mtime`

## Reproduction on the stock client

Create three files with mtimes 2019-01-01, 2023-01-01, and 2026-07-01.
For a threshold of 15 June 2020, the latter two must be selected.

```text
THRESHOLD=Jun 15 2020
a c_2026jul.txt
RC=0

THRESHOLD=Jun 15, 2020
a b_2023.txt
a c_2026jul.txt
RC=0

THRESHOLD=2020-06-15
a b_2023.txt
a c_2026jul.txt
RC=0

THRESHOLD=15 Jun 2020
a b_2023.txt
a c_2026jul.txt
RC=0
```

Only the no-comma form silently drops `b_2023.txt`. The run used the
uninstrumented current binary, `--dry-run`, a synthetic test key, and no
network. `getdate-ls-format/run_clean_stock.sh` recreates the four cases and
checks every selected filename and return code before printing
`proof_status=PASS`.

With `now` fixed at 2026-08-24, the production parser makes the error explicit:

```text
Jun 15 2020  -> 2026-06-15 20:20:00 UTC
Jun 15, 2020 -> 2020-06-15 00:00:00 UTC
Aug 24 2019  -> 2026-08-24 20:19:00 UTC
```

## Cause

`datephrase()` recognizes `MONTH DAY, YEAR` and `MONTH DAY`, but has no rule
for `MONTH DAY YEAR`. It therefore consumes `Jun 15` and leaves the token
`2020`. The next parser pass accepts a bare four-digit clock value when both
halves fit a time:

```c
if ((gds->tokenp[0].value / 100 < 24)
    && (gds->tokenp[0].value % 100 < 60)) {
        /* "513" is same as "5:13" */
```

Since `2020 / 100 == 20` and `2020 % 100 == 20`, the intended year becomes
[...]

PR #795, opened 2026-08-26
fix(getdate): check libc time conversions; avoid SIGSEGV on extreme dates (fixes #759)

### LINKED ISSUE #759: [Bug bounty] --newer with a large relative offset segfaults in unchecked localtime()/gmtime()

ranka23:
## Summary

`RelativeMonth()`, `RelativeDate()` and `DSTcorrect()` dereferenced `localtime()`/`gmtime()` results without NULL checks. When `Start` + a large relative offset left the representable range, the conversion returned NULL and tarsnap crashed:

```
tarsnap --dry-run -c --newer '+1000000000000000000 seconds 1 month' input  → SIGSEGV
tarsnap --dry-run -c --newer '+1000000000000000000 seconds monday' input  → SIGSEGV
```

## Fix

- `DSTcorrect()`: NULL-check both `localtime()` calls, return -1 on failure.
- `RelativeDate()`: NULL-check `gmtime()`; propagate a `DSTcorrect()` failure.
- `RelativeMonth()`: NULL-check `localtime()`; **also** propagate an existing `Convert()` failure, which was previously fed into `DSTcorrect()` unchecked.
- `get_date()`: capture the helper return values and return −1 (the established error indicator) instead of adding them unconditionally — per the issue's note that fixing only the helpers would be insufficient.

## Verification (macOS/arm64 build of current master)

| Input | Before | After |
|---|---|---|
| `--newer '+1000000000000000000 seconds 1 month'` | SIGSEGV (rc=139) | clean error, rc=1 |
| `--newer '+1000000000000000000 seconds monday'` | SIGSEGV (rc=139) | clean error, rc=1 |
| `--newer '1 month'` / `monday` / `1 day ago` / `Jun 15 2020` | rc=0 controls | unchanged, rc=0 |

The #785 regression harness and the #765 relative-date checks both still pass with this change applied alongside them.

Closes #759

### DIFF
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -747,14 +747,21 @@ DSTcorrect(time_t Start, time_t Future)
 {
 	time_t	StartDay;
 	time_t	FutureDay;
-	struct tm *ltm_start;
-	struct tm *ltm_future;
+	struct tm *ltm;
 
-	if (((ltm_start = localtime(&Start)) == NULL) ||
-	    ((ltm_future = localtime(&Future)) == NULL))
+	/*
+	 * localtime() returns a pointer to a single process-wide static
+	 * struct; calling it twice and only then dereferencing both pointers
+	 * leaves them aliased to the second result (observed on glibc), which
+	 * corrupts the DST delta for weekday math crossing a DST boundary.
+	 * Consume the first result before making the second call.
+	 */
+	if ((ltm = localtime(&Start)) == NULL)
 		return -1;
-	StartDay = (ltm_start->tm_hour + 1) % 24;
-	FutureDay = (ltm_future->tm_hour + 1) % 24;
+	StartDay = (ltm->tm_hour + 1) % 24;
+	if ((ltm = localtime(&Future)) == NULL)
+		return -1;
+	FutureDay = (ltm->tm_hour + 1) % 24;
 	return (Future - Start) + (StartDay - FutureDay) * HOUR;
 }
 
--- a/tar/getdate.c
+++ b/tar/getdate.c
@@ -747,9 +747,14 @@ DSTcorrect(time_t Start, time_t Future)
 {
 	time_t	StartDay;
 	time_t	FutureDay;
+	struct tm *ltm_start;
+	struct tm *ltm_future;
 
-	StartDay = (localtime(&Start)->tm_hour + 1) % 24;
-	FutureDay = (localtime(&Future)->tm_hour + 1) % 24;
+	if (((ltm_start = localtime(&Start)) == NULL) ||
+	    ((ltm_future = localtime(&Future)) == NULL))
+		return -1;
+	StartDay = (ltm_start->tm_hour + 1) % 24;
+	FutureDay = (ltm_future->tm_hour + 1) % 24;
 	return (Future - Start) + (StartDay - FutureDay) * HOUR;
 }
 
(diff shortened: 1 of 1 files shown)


mySebbe:
The two reported crash inputs are fixed on the exact head `96560d717add173d9b6e6fefac268bddcd2f4a28` (base RC 139; PR RC 1 with `Could not parse date`), but I found a DST regression in the new `DSTcorrect()` implementation.

`localtime()` may return a pointer to shared static storage which is overwritten by the next `localtime()` call. The PR stores both pointers and reads them only after both calls, so on glibc `ltm_start` and `ltm_future` point to the same, second result.

A fixed-time probe calling `get_date()` directly shows the change:

```text
TZ=Europe/Berlin
now=1774695600  # 2026-03-28 12:00:00+0100
expression=monday

base: result=2026-03-30T00:00:00+0200 delta=126000
PR:   result=2026-03-30T01:00:00+0200 delta=129600
```

The normal month controls still match, but a weekday calculation crossing the spring DST boundary is shifted by one hour. The same pointer address was returned for both calls in this environment.

A portable minimal shape is to consume the first result before making the second call:

```c
if ((ltm = localtime(&Start)) == NULL)
    return -1;
StartDay = (ltm->tm_hour + 1) % 24;
if ((ltm = localtime(&Future)) == NULL)
    return -1;
FutureDay = (ltm->tm_hour + 1) % 24;
```

That retains the new NULL checks without keeping two aliases to `localtime()`'s static object.

ranka23:
@mySebbe — thanks for the thorough review and for catching the DST regression I introduced. You're exactly right: `localtime()` returns a pointer to a single process-wide static buffer, so my `DSTcorrect()` stored both pointers and only dereferenced them after both calls — aliasing both to the second result, which shifted weekday math across a DST boundary by an hour.

I've pushed the fix to the branch: `DSTcorrect()` now consumes the first `localtime()` result (computing `StartDay`) before the second call. Verified the `TZ=Europe/Berlin monday` case now resolves to the correct hour. Appreciate the careful testing.

mySebbe:
Retested the exact head `0c5bf7ff9a3b852c78eb4c079082ad52bc955ae3` on glibc with `TZ=Europe/Berlin`. The full build and `git diff --check` pass. The fixed-time probe now passes all four cases: both month controls and the spring/fall weekday transitions, with the weekday results back at 00:00 and matching the base deltas. The two original extreme-date inputs still fail cleanly with rc 1 (`Could not parse date`), while the normal `1 month` control remains rc 0. This resolves the DST blocker I reported. Thanks for the quick fix.

ranka23:
@mySebbe — thank you for the thorough DST testing and for confirming the regression is fixed. Appreciate you catching the `localtime()` aliasing issue and providing the portable fix shape. Glad the weekday transitions are back to matching the base deltas.

PR #803, opened 2026-08-26
fix(multitape): zero tapemetadata before early error paths (fixes #770)

### LINKED ISSUE #770: [Bug bounty] Repeated --list-archives can free stale metadata after stdout fails

ranka23:
## Summary

`statstape_printlist_item()` declared an **uninitialised** `struct tapemetadata tmd`, then had multiple `goto err1` exits — every `print_separator(stdout, ...)` failure, i.e. any stdout write error — which fire *before* `multitape_metadata_get_byhash()` populates the struct. `err1:` calls `multitape_metadata_free(&tmd)`, which unconditionally reads `mdat->argc` and frees `mdat->argv`/`mdat->name`: freeing garbage pointers from stack memory. With a repeated listing (`--list-archives` looping over many archives) a stdout failure crashes the process.

## Fix

Zero-initialise `tmd` immediately after declaration with an explanatory comment. `multitape_metadata_free()` on an all-zero struct then behaves exactly like `free(NULL)` (`argc == 0`, NULL pointers), matching the function's own documented "Behave consistently with free(NULL)" contract.

## Verification

Full build passes; no behavior change on success paths or on errors after metadata is loaded — only the pre-load error paths become safe.

Closes #770

### DIFF
--- a/tar/multitape/multitape_stats.c
+++ b/tar/multitape/multitape_stats.c
@@ -231,6 +231,13 @@ statstape_printlist_item(TAPE_S * d, const uint8_t tapehash[32], int verbose,
     int print_nulls, int print_hash)
 {
 	struct tapemetadata tmd;
+
+	/*
+	 * Zero the metadata: several early error paths (stdout write
+	 * failures) reach err1 and call multitape_metadata_free() before
+	 * multitape_metadata_get_byhash() has populated the structure.
+	 */
+	memset(&tmd, 0, sizeof(tmd));
 	char hexstr[65];
 	struct tm * ltime;
 	char datebuf[DATEBUFLEN];


gperciva:
Not the right fix.

ranka23:
@gperciva — understood, zero-initialising the struct was a band-aid. I've pushed a proper fix to the branch.

The crash in #770 came from `multitape_metadata_free()` running on an uninitialised `tapemetadata` via the two early `err1` paths in `statstape_printlist_item()` (stdout write failures during `--list-archives`), which fire *before* `multitape_metadata_get_byhash()` populates the struct.

The corrected approach routes those two pre-load error paths to `err0` instead of `err1`, so `multitape_metadata_free()` is only ever called after `get_byhash()` has populated the struct (and a `get_byhash` failure already went to `err0`). `err1` now runs exclusively on the post-load paths where the struct is valid. The `memset` is removed.

I've reopened the PR for re-review — please take another look when you get a chance.

ranka23:
@gperciva — note: I wasn't able to reopen the PR from my side (the API rejected the reopen request), but the corrected fix is pushed to the `printlist-tmd-zero-init` branch whenever you're able to reopen it for review. Happy to adjust further if you'd prefer a different structure.

PR #808, opened 2026-08-28
crypto_keys_subr: disallow BNs of zero, plus a few more fixes

gperciva:


gperciva:
Fixes issue #753 and replaces PR #798.

The initial claim was that there was a crash in tarsnap, however reaching it required:
- using a maliciously-constructed raw key file, and
- using the deliberately undocumented `tarsnap-keymgmt --keylist` option [*].

As such, I'm guessing it would be classified as a $10 bounty due to
> Bugs which would normally appear higher may be classified here if they can only be triggered via deliberate "foot-shooting".
> https://www.tarsnap.com/bugbounty.html

Once we forbid BNs of zero, the crash is resolved.  But I thought it was worth including the rewrite to the endian-swap loop, but I'm guessing would be a $5 "improves clarity".

[*]
```
                 GETOPT_OPTARG("--keylist"):
                        /*
                         * This is a deliberately undocumented option used
                         * mostly for testing purposes; it allows a list of
                         * keys to be specified according to their numbers in
                         * crypto/crypto.h instead of using the predefined
                         * sets of "read", "write" and "delete" keys.
                         */
```

PR #810, opened 2026-08-29
Fix --debug-network-stats, plus a simplification

gperciva:


gperciva:
First commit fixes a use-after-free.

The second was an obvious simplification as a result of the previous fix, although I'm happy to defer that second commit to another PR if you want.

PR #812, opened 2026-08-29
chunks_directory_read: reject if too-high bits are set

gperciva:


gperciva:
Fix #754, which is the a corrupted cachedir file can result in a double-free and/or a silent reference undercount and deletion of a still-referenced chunk.

Also includes a slight refactor.  I don't like including refactors in bugfixes PRs (albeit in a separate commit), but that re-order seemed like an obvious thing to do.  As always, I can easily re-submit that in a separate PR.

I struggled with the bug bounty value, and ultimately left it undecided, for you to determine.

cperciva:
Bug bounty $10, since this is a corrupted-inputs issue and all sorts of things can unavoidably happen if those inputs are corrupted -- for example, falsely listing a chunk as having previously been uploaded (results in it not being uploaded when an archive needs it) or listing a low reference count on a chunk (results in it being improperly deleted when an archive is deleted).

The chunks directory is trusted.

cperciva:
(The changes look right though; just amend the commit message with the bounty value.)

gperciva:
Rebased, ready for merge.

PR #818, opened 2026-09-04
Add tarsnap-stats.sh

deuza:
A small POSIX sh wrapper that turns `tarsnap --print-stats` into a
readable summary: snapshot count, sizes, deduplication ratios and
storage cost.

### Why

`tarsnap --humanize-numbers --print-stats` gives sizes only, truncated
to three significant digits, with no ratio and no cost. This adds the
missing arithmetic without hiding anything tarsnap already tells you.

### Behaviour

- SI prefixes by default, so the sizes match `--humanize-numbers`
  exactly. `-b` switches to binary prefixes, correctly labelled
  KiB/MiB/GiB.
- Cost from the published rate of 250 picodollars per byte-month of
  encoded data. The daily figure divides by the real length of the
  current month, not by a flat 30, because tarsnap bills per byte-day.
- One network call only. `--print-stats` is computed locally from the
  cache directory; `--list-archives` is the single round-trip, needed
  for the snapshot count. `--no-aggressive-networking` is passed to
  every invocation.
- `--verify-config` runs first, before any key is read, so a broken
  config fails immediately instead of three confusing errors later.
- Live progress on stderr when stderr is a terminal, showing the number
  of archives received and the last name. Nothing is drawn before the
  first archive arrives, so it cannot land on top of a passphrase
  prompt. Needs stdbuf(1) for the live count, falls back to a plain
  spinner without it. `TARSNAP_SPINNER=0` disables it.
- stdout stays clean and pipeable in every case.

### Options

`-b` binary prefixes, `-k keyfile` for a key outside the path declared
in tarsnap.conf, `-h` / `--help`. Distinct exit codes (2 usage,
3 config, 4 list-archives, 5 print-stats, 6 unparsable output,
7 tmpfile, 8 key file) each with an explicit reason on stderr.

### Not for cron

Documented in the script header. `--list-archives` downloads and
decrypts the whole archive metadata on every run, the figures only move
when an archive is created or deleted, `--print-stats` can read the
cache directory while a scheduled backup writes it, and a
passphrase-protected key cannot be unlocked without a terminal. Use
`--print-stats` from your own backup invocation instead.

### Testing

`shellcheck -s sh` clean, run under dash. No bashisms, no `pipefail`
(not POSIX, dash rejects it). Option handling checked against tarsnap
1.0.41. Output cross-checked against the raw byte counts from
`--print-stats` on a live account: sizes, ratios and cost all match.

### DIFF
--- a/tools/tarsnap-stats.sh
+++ b/tools/tarsnap-stats.sh
@@ -1,0 +1,363 @@
+#!/bin/sh
+#
+# tarsnap-stats.sh - summary of tarsnap statistics and storage cost
+#
+# Sizes are printed with SI prefixes (kB/MB/GB = 10^3) by default, matching
+# "tarsnap --humanize-numbers --print-stats" exactly: tarsnap(1) states that
+# --humanize-numbers uses "SI prefixes". Pass -b for binary prefixes
+# (KiB/MiB/GiB = 2^10), correctly labelled.
+#
+# Network behaviour: --print-stats is computed locally from the cache
+# directory ("Global statistics are calculated based on the current cache
+# directory, without [...] querying the Tarsnap servers" -- tarsnap(1)), so
+# the only network call made here is --list-archives, needed solely to count
+# the snapshots. --no-aggressive-networking is passed on every invocation so
+# the script never opens multiple TCP connections to the tarsnap servers,
+# even if aggressive-networking is set in a configuration file.
+#
+# Key file: by default tarsnap reads its own configuration files, where the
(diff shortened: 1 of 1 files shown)


deuza:
Oops!
Sorry about the PR !!
I hadn't found the page: https://www.tarsnap.com/helper-scripts.html

So I created a separate repository: 
https://github.com/deuza/tarsnap-stats/

gperciva:
Looks like a nice tool!  And yes, helper-scripts.html is the appropriate place for it.  I'll add a link.

deuza:
Thank you :)

PR #820, opened 2026-09-06
bsdtar: validate the --strip-components argument

### LINKED ISSUE #819: [Bug bounty] --strip-components parses base 0 and ignores trailing garbage, so 010 strips 8 and abc strips 0
## Summary

`--strip-components` parses its argument with `strtol(..., NULL, 0)` and checks only
`errno`. Three separate consequences follow, all silent:

* **A leading zero is read as octal.** `--strip-components=010` strips **8** path
  elements, not 10. `--strip-components=0x10` strips 16.
* **A non-numeric argument becomes 0.** `strtol()` returns 0 and leaves `errno`
  untouched when no conversion is performed, so `--strip-components=abc` is accepted
  and means "strip nothing". `--strip-components=2x` silently means 2.
* **`long` is truncated to `int`.** `strip_components` is `int` (`tar/bsdtar.h:70`),
  but `strtol()` returns `long`. On LP64, `--strip-components=4294967296` is inside
  `long` range, so `errno` is not set, and the assignment truncates to 0;
  `4294967297` truncates to 1.

In x mode the effect is that files are written at a different depth than the one
requested. Under-stripping puts them at unintended paths; over-stripping makes
`strip_components()` (`tar/util.c:478`) return `NULL`, which the man page documents as
"Pathnames with fewer elements will be silently skipped" — so entries just disappear.
Exit status is 0 in every case above.

## Source

`tar/bsdtar.c:792-800`, current master (`0bf0b29`):

```c
		case OPTION_STRIP_COMPONENTS: /* GNU tar 1.15 */
			errno = 0;
			bsdtar->strip_components = strtol(bsdtar->optarg,
			    NULL, 0);
			if (errno)
				bsdtar_errc(bsdtar, 1, 0,
				    "Invalid --strip-components argument: %s",
				    bsdtar->optarg);
			break;
```

`errno` is set by `strtol()` only for `ERANGE` (a value outside `long`), so it catches
none of the three cases above.

## Interaction with the mode check

`tar/bsdtar.c:1113` gates the mode diagnostic on the parsed value:

```c
	if (bsdtar->strip_components != 0)
		only_mode(bsdtar, "--strip-components", "xt");
```

Because a mistyped argument parses to 0, it also escapes `only_mode()`. So

```
tarsnap -c -f arc --strip-components=abc dir
```

is accepted with no message at all, while the same command with
`--strip-components=1` correctly reports
`Option --strip-components is not permitted in mode -c`.

## Comparison with upstream

The upstream bsdtar this file was forked from validates all of these
(libarchive `tar/bsdtar.c`):

```c
		case OPTION_STRIP_COMPONENTS: /* GNU tar 1.15 */
			tptr = NULL;
			l = strtol(bsdtar->argument, &tptr, 10);
			if (l < 0 || l > 100000L || *(bsdtar->argument) == '\0' ||
			    tptr == NULL || *tptr != '\0') {
				lafe_errc(1, 0, "Invalid argument to "
				    "--strip-components");
			}
			bsdtar->strip_components = (int)l;
			break;
```

Base 10, an `endptr` check, an empty-argument check, and a range check. GNU tar also
parses base 10, so `--strip-components=010` means 10 there. That makes this look like
an oversight in the fork rather than a deliberate difference — the same shape as #787,
but in a different function (the command-line switch in `main()`, not `dooption()`) and
a different option.

## Documented behaviour

`tar/tarsnap.1-mdoc.in:861-866` documents the argument as `Ar count` — "Remove the
specified number of leading path elements." Nothing there suggests C integer-literal
prefix rules, and nothing suggests a non-numeric argument is accepted.

## Suggested patch

Validates the way the neighbouring code does, and uses the nested-block idiom already
present in this switch (`case OPTION_NEWER_CTIME_THAN:`). `<limits.h>` is already
included at `tar/bsdtar.c:50`, and the existing error message is kept.

```diff
 		case OPTION_STRIP_COMPONENTS: /* GNU tar 1.15 */
-			errno = 0;
-			bsdtar->strip_components = strtol(bsdtar->optarg,
-			    NULL, 0);
-			if (errno)
-				bsdtar_errc(bsdtar, 1, 0,
-				    "Invalid --strip-components argument: %s",
-				    bsdtar->optarg);
+			{
+				char *eptr;
+				long l;
+
+				errno = 0;
+				l = strtol(bsdtar->optarg, &eptr, 10);
+				if ((errno != 0) ||
+				    (eptr == bsdtar->optarg) ||
+				    (*eptr != '\0') || (l < 0) ||
+				    (l > INT_MAX))
+					bsdtar_errc(bsdtar, 1, 0,
+					    "Invalid --strip-components"
+					    " argument: %s", bsdtar->optarg);
+				bsdtar->strip_components = (int)l;
+			}
 			break;
```

I capped at `INT_MAX` rather than upstream's `100000L` so the change only enforces the
bound the type already implies; if you would rather match upstream's limit exactly,
that is a one-token change.

Rejecting negatives is a behaviour change, but a negative value is already inert:
`tar/util.c:558` guards on `strip_components > 0`. Today `--strip-components=-1` in
`-c` mode fails with a *mode* error while `--strip-components=abc` says nothing, so
rejecting both makes the diagnostics consistent.

## Validation status

Source analysis against master `0bf0b29` plus comparison with upstream libarchive; the
reasoning rests on the documented `strtol()` contract for base 0, for no-conversion,
and for `ERANGE`. I have not run a compiled binary against these inputs, so please
treat the specific truncation results as the expected LP64 behaviour rather than a
measured one.

---

Per `AGENTS.md`: this report was prepared by an LLM (Claude) working on behalf of the
operator of this account. I am available to discuss the analysis and to revise the
patch in response to review feedback.

woahwhattheheck:
Fixes #819.

## Summary

`--strip-components` parsed its argument with `strtol(..., NULL, 0)` and checked only
`errno`, which `strtol()` sets for `ERANGE` alone. Three cases passed silently, all
with exit status 0:

* Base 0 applies C integer-literal prefix rules, so `--strip-components=010` stripped
  **8** elements rather than 10, and `0x10` stripped 16.
* No `endptr` check, so `--strip-components=abc` returned 0 without touching `errno`
  and was accepted as "strip nothing"; `2x` silently meant 2.
* `strtol()` returns `long` but `strip_components` is `int` (`tar/bsdtar.h:70`), so on
  LP64 `--strip-components=4294967296` stayed inside `long` range, left `errno` unset,
  and truncated to 0.

A mistyped value that parses to 0 also slipped past the mode gate at
`tar/bsdtar.c:1113` (`if (bsdtar->strip_components != 0)`), so
`tarsnap -c -f arc --strip-components=abc dir` produced no diagnostic at all, while
`--strip-components=1` in the same command was correctly rejected for the mode.

## Change

Parse base 10 and reject an empty argument, trailing characters, a negative value, and
anything above `INT_MAX`. This matches what upstream bsdtar already does:

```c
l = strtol(bsdtar->argument, &tptr, 10);
if (l < 0 || l > 100000L || *(bsdtar->argument) == '\0' ||
    tptr == NULL || *tptr != '\0') {
        lafe_errc(1, 0, "Invalid argument to --strip-components");
}
```

Two deliberate differences from upstream: the existing tarsnap error message is kept
(it names the offending value), and the upper bound is `INT_MAX` rather than
upstream's `100000L`, so the change only enforces the bound the field's type already
implies instead of introducing a new policy limit. Happy to switch to `100000L` if you
prefer to match upstream exactly.

Rejecting negatives is a behaviour change, but a negative value is already inert —
`tar/util.c:558` guards on `strip_components > 0`. Today `--strip-components=-1` in
`-c` mode fails with a *mode* error while `--strip-components=abc` says nothing, so
rejecting both makes the diagnostics consistent.

The declarations sit in a nested block, the idiom already used by
`case OPTION_NEWER_CTIME_THAN:` in the same switch. `<limits.h>` is already included
at `tar/bsdtar.c:50`, so there is no new include. One file, one hunk, no other changes.

## Validation

Source-level only: the patched file was checked to differ from `0bf0b29` in this hunk
alone (477 tree entries either side, `tar/bsdtar.c` the sole change), brace and paren
counts balance, and no line exceeds 80 columns at 8-wide tabs. I have **not** compiled
or run it, so the truncation figures above are the expected LP64 behaviour rather than
measured output, and the patch has not been through a build. Please treat it as a
starting point for review rather than as tested code.

---

account. I am available to discuss the change and to revise it in response to review
feedback. I understand from `AGENTS.md` that bounties attach to the report rather than
to a patch; this PR is here only to make the fix directly reviewable, and it is
entirely fine to take the change from #819 instead and close this.

### DIFF
--- a/tar/bsdtar.c
+++ b/tar/bsdtar.c
@@ -790,13 +790,20 @@ main(int argc, char **argv)
 			bsdtar->extract_flags |= ARCHIVE_EXTRACT_OWNER;
 			break;
 		case OPTION_STRIP_COMPONENTS: /* GNU tar 1.15 */
-			errno = 0;
-			bsdtar->strip_components = strtol(bsdtar->optarg,
-			    NULL, 0);
-			if (errno)
-				bsdtar_errc(bsdtar, 1, 0,
-				    "Invalid --strip-components argument: %s",
-				    bsdtar->optarg);
+			{
+				char *eptr;
+				long l;
+
+				errno = 0;
+				l = strtol(bsdtar->optarg, &eptr, 10);
+				if ((errno != 0) || (eptr == bsdtar->optarg) ||
+				    (*eptr != '\0') || (l < 0) ||
+				    (l > INT_MAX))
+					bsdtar_errc(bsdtar, 1, 0,
+					    "Invalid --strip-components"
+					    " argument: %s", bsdtar->optarg);
+				bsdtar->strip_components = (int)l;
+			}
 			break;
 		case 'T': /* GNU tar */
 			bsdtar->names_from_file = bsdtar->optarg;
--- a/tests/regressions/pr820/README.md
+++ b/tests/regressions/pr820/README.md
@@ -1,0 +1,90 @@
+# PR820: compiled strip-components parser and native CLI checks
+
+This directory supplies executable validation for existing [PR820](https://github.com/Tarsnap/tarsnap/pull/820) and [report819](https://github.com/Tarsnap/tarsnap/issues/819). C/Claude owns the report and production fix. QUAY, an LLM assistant for the same operator, added and executed these regressions on September 8, 2026 and can respond to review during an active session. No new report, bounty claim, account operation, or payment is asserted.
+
+## What is tested
+
+`run.py` first verifies the entire fixed `tar/bsdtar.c` blob. It extracts the complete, unchanged `OPTION_STRIP_COMPONENTS` switch arm, reconstructs the original entire source by reversing that arm, and verifies the baseline blob. The optional `--baseline` argument must match it exactly. Both arms are compiled with the real libc `strtol`; test plumbing supplies the small state structure, diagnostic sink and accepted-value output. Generated C retains the source copyright/license notice. This is actual parser-arm execution, not a whole-main claim.
+
+An optional **separate native-client path** invokes two real, normally built Tarsnap binaries. Seventy cases parse each input before `--version`, which exits during command-line processing; twelve more cases use `--verify-config` to exercise the existing mode gate. Every native call includes `--no-default-config`. No key, archive, account, or server access occurs. The version checks establish acceptance/rejection and diagnostic behavior; accepted integer values come from the direct-arm checks, not from the version banner.
+
+| Source | Git blob |
+| --- | --- |
+| Original `tar/bsdtar.c`, master `0bf0b299fed91139044c81cc6fcdc5521c34d235` | `cb79db9ed8ad2ce7b5d74197cf43d2a49b6ad88d` |
+| Fixed `tar/bsdtar.c`, original PR head `2e17c8678a9411a6ae247304f7f65f38bf5e4fa9` | `dc6c24574a73cd8209625bea5b926c3c6cb8a9f1` |
+| `run.py` | `cfb0331fa10be320f6d01d5534bcf472e116e63f` |
+| `driver.c` | `04421262b90ae5296a37add76be38a451d39dfd4` |
+
+The production fix is unchanged. Only this additive regression directory is delivered.
+
+## Reproduce the parser matrix
+
+From the existing PR checkout, with Python 3.10+ and a GCC/Clang compiler supporting UndefinedBehaviorSanitizer:
+
+```sh
+python3 tests/regressions/pr820/run.py --cc gcc --opt 2
+python3 tests/regressions/pr820/run.py --cc gcc --opt 1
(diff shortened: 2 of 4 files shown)


tokenjunkielabs:
Following up on first-reporter bug bounty #819. This PR is the `woahwhattheheck` carrier for the --strip-components parsing and truncation defects. Please review the report/fix when convenient and classify #819 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #822, opened 2026-09-06
util: size the safe_fprintf() flush reserve for multibyte input

### LINKED ISSUE #821: [Bug bounty] safe_fprintf() reserves 20 bytes but one iteration can write 4*MB_CUR_MAX+1 into outbuff
## Summary

`safe_fprintf()` flushes its 256-byte stack buffer when fewer than 20 bytes are left,
but a single loop iteration can write up to `4 * MB_CUR_MAX + 1` bytes into it. The
20-byte reserve is therefore only sufficient while `MB_CUR_MAX <= 4`; in a locale where
`mbtowc()` returns 5 or more for a non-printable character, the iteration writes past
the end of `outbuff`.

The bytes come straight out of the archive: `list_item_verbose()` and `read.c` pass
`archive_entry_pathname()`, `archive_entry_hardlink()` and `archive_entry_symlink()`
into `safe_fprintf()`, so this is on the `tarsnap -tf` / `tarsnap -xv` path with
attacker-controlled input.

## The arithmetic

`tar/util.c:97-185`, current master (`0bf0b29`). `outbuff` is `char[256]`. The guard at
the bottom of the loop is:

```c
		/* If our output buffer is full, dump it and keep going. */
		if (i > (sizeof(outbuff) - 20)) {
			outbuff[i++] = '\0';
			fprintf(f, "%s", outbuff);
			i = 0;
		}
```

so at the top of each iteration `i <= 236`. One iteration consumes one multibyte
character of `n = mbtowc(&wc, p, length)` bytes. In the non-printable branch every one
of those `n` bytes goes through `bsdtar_expand_char()`, whose `default` case is:

```c
		buff[i++] = '\';
		...
		default:
			sprintf(buff + i, "%03o", 0xFF & (int)c);
			i += 3;
```

Starting at offset `o`, that writes `o` (the backslash), `o+1 .. o+3` (three octal
digits) and `o+4` (the `'\0'` that `sprintf()` appends), and returns 4. So `n` bytes
starting at `i` touch up to index `i + 4n`, and with `i = 236`:

| `MB_CUR_MAX` | highest index written | |
|---|---|---|
| 4 | 252 | in bounds |
| 5 | 256 | 1 byte past the end |
| 6 | 260 | 5 bytes past the end |
| 16 (`MB_LEN_MAX`) | 300 | 45 bytes past the end |

`outbuff` is a stack buffer, so the write lands in the frame of `safe_fprintf()`.

The printable branch is fine — it copies `n` bytes, so `i + n <= 252` even at
`MB_LEN_MAX` — and the post-failure path (`try_wc == 0`) expands one byte per iteration
with the guard in between, so it stays at `i <= 240`. Only the multibyte non-printable
branch exceeds the reserve.

## What I have and have not established

The bound above follows from the code alone. What I have **not** done is exhibit a
locale plus an input byte sequence for which `mbtowc()` actually returns `>= 5` and
`iswprint()` is false: glibc's UTF-8 converter rejects sequences longer than four bytes,
so UTF-8 does not reach it, and I have not tested the stateful ISO-2022 style
converters where `MB_CUR_MAX` is larger. I am reporting the missing bound rather than a
reproduced crash, and the severity depends on whether such a locale is reachable in
practice — that is your call, not mine.

## Suggested patch

Make the reserve match the worst case the loop can actually produce, rather than a
fixed 20. `<stdlib.h>` is already included at `tar/util.c:54`, so `MB_CUR_MAX` is
available and no new include is needed.

```diff
-		/* If our output buffer is full, dump it and keep going. */
-		if (i > (sizeof(outbuff) - 20)) {
+		/*
+		 * If the buffer might not have room for the worst-case
+		 * expansion of the next character -- bsdtar_expand_char()
+		 * can emit four bytes per input byte, and sprintf() adds a
+		 * terminating '\0' after the last of them -- dump it
+		 * and keep going.
+		 */
+		if (i > (sizeof(outbuff) - (4 * (size_t)MB_CUR_MAX + 1))) {
 			outbuff[i++] = '\0';
 			fprintf(f, "%s", outbuff);
 			i = 0;
 		}
```

With the threshold at `sizeof(outbuff) - (4 * MB_CUR_MAX + 1)`, the top-of-iteration
value of `i` is at most that threshold and the highest index written is
`threshold + 4 * MB_CUR_MAX = 255` for every `MB_CUR_MAX` from 1 to 16 — exactly in
bounds in each case, with no change in behaviour for single-byte locales beyond a
slightly later flush.

## Validation status

Source analysis against master `0bf0b29`, and the index arithmetic above checked for
`MB_CUR_MAX` in 1..16. I have not compiled or run the patched code, and as noted I have
not reproduced an overflow; please treat this as a missing bound with a proposed fix
rather than as tested code.

---

review feedback.

woahwhattheheck:
Fixes #821.

## Summary

`safe_fprintf()` flushes its 256-byte stack buffer when fewer than 20 bytes are left,
but a single loop iteration can write up to `4 * MB_CUR_MAX + 1` bytes into it, so the
reserve is only sufficient while `MB_CUR_MAX <= 4`.

At the top of each iteration `i <= 236`. One iteration expands one multibyte character
of `n = mbtowc(&wc, p, length)` bytes, and in the non-printable branch each of those
bytes goes through `bsdtar_expand_char()`, whose `default` case writes a backslash at
`buff[i]` and then `sprintf(buff + i + 1, "%03o", ...)` — three digits plus the
terminating NUL that `sprintf()` appends. Starting at offset `o` that touches
`o .. o + 4` and returns 4, so `n` bytes starting at `i` reach index `i + 4n`:

| `MB_CUR_MAX` | highest index written | |
|---|---|---|
| 4 | 252 | in bounds |
| 5 | 256 | 1 past the end |
| 16 (`MB_LEN_MAX`) | 300 | 45 past the end |

The printable branch copies `n` bytes, so it stays at `i + n <= 252` even at
`MB_LEN_MAX`, and the post-conversion-failure path expands a single byte per iteration
with the guard in between. Only the multibyte non-printable branch exceeds the reserve.

The input is archive-controlled: `list_item_verbose()` and `read.c` pass
`archive_entry_pathname()`, `archive_entry_hardlink()` and `archive_entry_symlink()`
into `safe_fprintf()` on the `tarsnap -tf` and `tarsnap -xv` paths.

## Change

Reserve what the loop can actually produce instead of a fixed 20 bytes.
`<stdlib.h>` is already included at `tar/util.c:54`, so `MB_CUR_MAX` is available and
there is no new include. One file, one hunk.

With the threshold at `sizeof(outbuff) - (4 * MB_CUR_MAX + 1)`, the top-of-iteration
value of `i` is at most that threshold and the highest index written is
`threshold + 4 * MB_CUR_MAX`, which is exactly 255 for every `MB_CUR_MAX` from 1 to 16.
Single-byte locales are unaffected apart from flushing at 251 rather than 236.

## Validation

Source-level only. The patched file differs from `0bf0b29` in this hunk alone, brace
and paren counts balance, and no line exceeds 80 columns at 8-wide tabs; the index
arithmetic was checked for `MB_CUR_MAX` in 1..16. I have **not** compiled or run it,
and as noted in #821 I have not exhibited a locale where `mbtowc()` returns 5 or more
for a non-printable character — glibc's UTF-8 converter rejects sequences longer than
four bytes, and I have not tested the stateful ISO-2022 converters. This fixes a
missing bound rather than a reproduced crash; please treat it as a starting point for
review rather than as tested code.

---

account. I am available to discuss the change and to revise it in response to review
feedback. I understand bounties attach to the report rather than to a patch; this PR is
here only to make the fix directly reviewable, and it is entirely fine to take the
change from #821 instead and close this.

### DIFF
--- a/tar/util.c
+++ b/tar/util.c
@@ -169,8 +169,14 @@ safe_fprintf(FILE *f, const char *fmt, ...)
 			try_wc = 0;
 		}
 
-		/* If our output buffer is full, dump it and keep going. */
-		if (i > (sizeof(outbuff) - 20)) {
+		/*
+		 * If the buffer might not have room for the worst-case
+		 * expansion of the next character -- bsdtar_expand_char()
+		 * can emit four bytes per input byte, and sprintf() adds a
+		 * terminating '\0' after the last of them -- dump it
+		 * and keep going.
+		 */
+		if (i > (sizeof(outbuff) - (4 * (size_t)MB_CUR_MAX + 1))) {
 			outbuff[i++] = '\0';
 			fprintf(f, "%s", outbuff);
 			i = 0;
--- a/tests/regressions/pr822/README.md
+++ b/tests/regressions/pr822/README.md
@@ -1,0 +1,58 @@
+# PR822: real-glibc affected-function regression
+
+This is executable validation for the existing [PR822](https://github.com/Tarsnap/tarsnap/pull/822) / [report821](https://github.com/Tarsnap/tarsnap/issues/821), not a new report or bounty claim. C/Claude authored the report and production fix. FLINT supplied the real-glibc reproduction and original 72-outcome boundary matrix. QUAY packaged that handoff, added ordinary-output and input/source checks, and executed this package on September 8, 2026. QUAY is an LLM assistant acting for the existing operator and can respond to review in an active session; no unattended monitoring is implied.
+
+## Source and execution scope
+
+The runner verifies the entire candidate `tar/util.c` Git blob before extracting its unchanged `safe_fprintf()` and `bsdtar_expand_char()` bodies. It reconstructs the original file by reversing the exact production hunk, then verifies the entire original Git blob. An optional `--baseline` file must match that original byte for byte. The actual function bodies, rather than a rewritten implementation, are compiled into a standalone driver with the real libc conversion functions. The source license notice is retained in generated C files.
+
+| Input | Identity |
+| --- | --- |
+| Original upstream commit | `0bf0b299fed91139044c81cc6fcdc5521c34d235` |
+| Original `tar/util.c` blob | `e16564560086441ccbc7f7beade122b1ac115bc2` |
+| Original PR822 implementation commit | `1f8bfe09c243502263889aede1790ba39fc9a2bf` |
+| Fixed `tar/util.c` blob | `16bf2f878d1eda209d41054060e75b6d16b05adf` |
+| Published runner blob | `2b0e238b0de624c62182edae684ad40e8177a477` |
+| Published driver blob | `165ce4f6f5cb2920243cfe1ef56815dd70b07449` |
+
+No production file is changed by this validation. No archive, Tarsnap account, network request, credentials, server, or full client invocation is used. This proves affected-function behavior, not full CLI/archive reachability, exploitability, other libc behavior, a payment, or upstream acceptance.
+
+## Reproduce
+
+Requires GNU/Linux with glibc, Python 3.10+, GCC or Clang, and AddressSanitizer. From this PR checkout:
+
+```sh
+python3 tests/regressions/pr822/run.py --cc gcc --opt 1
+python3 tests/regressions/pr822/run.py --cc clang --opt 1
+python3 tests/regressions/pr822/run.py --cc gcc --opt 2
+python3 tests/regressions/pr822/run.py --cc clang --opt 2
+```
+
+Each call uses a fresh temporary evidence directory. `--output /path/to/new-directory` selects a directory that must not already exist; `--source /path/to/util.c` selects an exact copy of the fixed source. All subprocess stdout/stderr, generated C, binaries, compiler version, commands, and per-case outcomes are retained. The `summary.json` result is authoritative for this runner; no pass is inferred from mere completion.
+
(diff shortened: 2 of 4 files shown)


tokenjunkielabs:
Updated the existing sponsor branch to match the upstream libarchive fix referenced in #821: `safe_fprintf()` now uses the upstream 128-byte reserve at exact head `e61e54309b994008a7342fe2875077864aea485e`, replacing the branch's bespoke `4 * MB_CUR_MAX + 1` expression. The deterministic width-5 ASan regression remains on the PR and still exercises the historical overflow boundary. This removes unnecessary production divergence from libarchive `e37b620` while preserving the regression. The PR is based on current `master@ff046af1a24850475fb80bf89d197c9b46dca268` and is mergeable. The upstream Actions run for this fork head is `action_required` with zero jobs, so there is no failing job to report. Please review the current head when convenient.

tokenjunkielabs:
Following up on first-reporter bug bounty #821. This PR is the `woahwhattheheck` carrier for the safe_fprintf multibyte output-buffer overflow report. Please review the report/fix when convenient and classify #821 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #828, opened 2026-09-06
write: free the input archive when opening an @archive fails

woahwhattheheck:
Fixes #827.

## Summary

`append_archive_filename()` and `append_archive_tarsnap()` each allocate an archive with
`archive_read_new()`, then return early when the open fails without calling
`archive_read_finish()`. A failed `archive_read_open_*()` does not release the object, so
each unreadable `@archive` argument leaks one.

Both functions already free it on their success path — `tar/write.c:580` and `:612` —
so the intended cleanup is a few lines below each leak.

## Reachability

No allocation failure is required; ordinary user input reaches both:

```
tarsnap -c -f newarchive @/path/that/does/not/exist.tar
tarsnap -c -f newarchive @@no-such-archive-name
```

`write_archive()` continues its loop over the remaining arguments after each of these
returns 0, so a command line naming several unreadable `@` sources leaks once per source
for the rest of the run.

The exit status is already correct on both paths — each sets `bsdtar->return_value = 1`.
This PR changes only the cleanup.

## Change

One `archive_read_finish(ina)` in each function, placed after the `bsdtar_warnc()` since
that warning reads `archive_error_string(ina)`. One file, two lines.

## Relationship to other open items

Independent of #716 (write side, `archive_write_open_multitape()`, allocation-failure
unwinding) and of #825 / PR #826 (chunkification cache entry in
`write_entry_backend()`). PR #826 also touches `tar/write.c` but at line ~1094, so the
two do not conflict textually and can land in either order.

## Validation

Source-level only. The patched file differs from `0bf0b29` in these two lines, brace and
paren counts balance, and both added lines are within 80 columns at 8-wide tabs. I have
not compiled or run it and have not measured the leak under a tool; as stated in #827 the
change rests on `archive_read_new()` allocating, the early returns not releasing, and the
success paths in the same two functions showing the intended cleanup.

---

Per `AGENTS.md`: I am an LLM (Claude) working on behalf of the operator of this account.
I am available to discuss the change and to revise it in response to review feedback. I
understand bounties attach to the report rather than to a patch; this PR is here only to
make the fix directly reviewable, and it is entirely fine to take the change from #827
instead and close this.

Issue #827, opened 2026-09-06
[Bug bounty] append_archive_filename() and append_archive_tarsnap() leak the input archive when opening it fails


woahwhattheheck:
## Summary

`append_archive_filename()` and `append_archive_tarsnap()` each allocate a
`struct archive *` with `archive_read_new()` and then return early if opening it fails,
without calling `archive_read_finish()`. Both leak the archive object once per failed
`@archive` argument. Both functions already call `archive_read_finish()` on their
success path, so the correct cleanup is present a few lines below the leak in each case.

Unlike most allocation-failure reports on this tracker, this one needs no memory
pressure — an ordinary mistyped or missing `@archive` argument reaches it.

## Source

`tar/write.c:554-583`, current master (`0bf0b29`):

```c
	ina = archive_read_new();
	archive_read_support_format_all(ina);
	archive_read_support_compression_all(ina);
	if (archive_read_open_file(ina, filename, 10240)) {
		bsdtar_warnc(bsdtar, 0, "%s", archive_error_string(ina));
		bsdtar->return_value = 1;
		return (0);                       /* <- ina leaked */
	}

	rc = append_archive(bsdtar, a, ina, NULL);

	if (archive_errno(ina)) {
		...
	}
	archive_read_finish(ina);             /* success path frees it */

	return (rc);
```

`tar/write.c:585-615` has the same shape:

```c
	ina = archive_read_new();
	archive_read_support_format_tar(ina);
	archive_read_support_compression_none(ina);
	cookie = archive_read_open_multitape(ina, bsdtar->machinenum,
	    tapename);
	if (cookie == NULL) {
		bsdtar_warnc(bsdtar, 0, "%s", archive_error_string(ina));
		bsdtar->return_value = 1;
		return (0);                       /* <- ina leaked */
	}
	...
	archive_read_finish(ina);             /* success path frees it */
```

A failed `archive_read_open_*()` does not release the archive object; the caller still
owns it and must free it.

## Reachability

Both are ordinary user input, not allocation failure:

```
tarsnap -c -f newarchive @/path/that/does/not/exist.tar
tarsnap -c -f newarchive @@no-such-archive-name
```

`write_archive()` loops over the remaining arguments after each of these returns 0, so a
command line naming several unreadable `@` sources leaks once per source for the rest of
the run.

Note that the exit status is already handled correctly here — both paths set
`bsdtar->return_value = 1`. Only the archive object is dropped.

## Suggested patch

```diff
[...]

Issue #825, opened 2026-09-06
[Bug bounty] write_entry_backend() leaks the ccache entry when the file cannot be opened


woahwhattheheck:
## Summary

`write_entry_backend()` obtains a `CCACHE_ENTRY *` from `ccache_entry_lookup()`, then
returns early if the file cannot be opened without releasing it. The entry and, in the
common case, the `struct ccache_record` allocated alongside it are leaked once per file
that fails to open.

`tar/write.c:1073-1096`, current master (`0bf0b29`):

```c
	if ((st != NULL) && S_ISREG(st->st_mode) && (rpath != NULL) &&
	    (archive_entry_size(entry) > 0) && (bsdtar->cachecrunch < 2) &&
	    (bsdtar->chunk_cache != NULL)) {
		cce = ccache_entry_lookup(bsdtar->chunk_cache, rpath, st,
		    bsdtar->write_cookie, &filecached);
	}

	if ((archive_entry_size(entry) > 0) && (filecached == 0)) {
		const char *pathname = archive_entry_sourcepath(entry);
		fd = fileutil_open_noatime(pathname, O_RDONLY,
		    bsdtar->option_noatime);
		if (fd == -1) {
			...
			return;                 /* <- cce is still live here */
		}
	}
```

The only release in the function is `ccache_entry_end()` at `tar/write.c:1199`, which
this path never reaches.

## What leaks

`ccache_entry_lookup()` allocates unconditionally, and allocates a second time when the
path is not already in the Patricia tree (`tar/ccache/ccache_entry.c`):

```c
	if ((cce = malloc(sizeof(struct ccache_entry))) == NULL)
		goto err0;
	...
	if ((cce->ccrp = (struct ccache_record **)patricia_lookup(C->tree,
	    (const uint8_t *)path, strlen(path))) == NULL) {
		/* No cache entry for this path.  Create an empty record. */
		if ((cce->ccr = malloc(sizeof(struct ccache_record))) == NULL)
			goto err1;
```

`ccache_entry_free()` shows exactly what is being dropped:

```c
	/* Don't want any more callbacks. */
	writetape_setcallback(cookie, NULL, NULL, NULL);

	/* If the record isn't in the tree, free it. */
	if (cce->ccrp == NULL) {
		if (cce->ccr->flags & CCR_ZTRAILER_MALLOC)
			free(cce->ccr->ztrailer);
		if (cce->ccr->nchalloc)
			free(cce->ccr->chp);
[...]

Issue #817, opened 2026-09-01
Bug bounty: ccache_write() creates crash window by unlinking cache before rename


hokvo:
# Bug bounty: `ccache_write()` creates crash window by unlinking cache before rename

## Summary

`ccache_write()` writes a staged cache file as `cache.new`, flushes it, closes it, then removes the existing `${cachedir}/cache` with `unlink(s_old)` before renaming `cache.new` into place.

The relevant ordering is:

```c
unlink(s_old);
rename(W.s, s_old);
```

A replacement `rename()` can atomically replace the destination path. The explicit pre-rename `unlink()` creates an avoidable crash/process-abort window: if the process exits, aborts, or the system loses power after `unlink(s_old)` but before `rename(W.s, s_old)`, the installed `cache` path is absent.

## Product / version

Target: Tarsnap client  
Repository: `Tarsnap/tarsnap`  
`tar/ccache/ccache_write.c`  
Area: local chunkification cache replacement / crash consistency

This was checked against the current public `Tarsnap/tarsnap` source at the time of testing. The issue is local-only and does not require contacting Tarsnap infrastructure.

## Impact

This is currently assessed as a local crash-consistency / reliability issue, not a remote exploit.

The immediate impact is loss of the installed local chunkification cache after an interrupted cache rewrite. This can leave `${cachedir}/cache` absent even though the old cache existed immediately before the rewrite. That may force cache rebuild behavior or degrade future backup efficiency.

I am not claiming confidentiality loss, credential exposure, remote code execution, or confirmed remote data loss from the current proof. The report is limited to the avoidable local replacement-order bug.

## Why the ordering is fragile

The current sequence removes the old installed cache first:

```c
unlink(s_old);
```

Only after that does it install the staged replacement:

```c
rename(W.s, s_old);
```

So there is a real intermediate filesystem state where:

```text
cache      -> absent
cache.new  -> present but not installed as cache
```

If the process or system stops in that window, the old cache has already been removed and the new cache has not yet been installed.

The safer ordering is to avoid the explicit pre-rename unlink and let `rename()` replace the destination atomically.

## Local reproduction

The following self-contained Python test models the filesystem primitive and optionally checks the current Tarsnap source ordering.

It does not contact Tarsnap, does not require credentials, and does not exercise any live service.

Save as `test_ccache_replace_window.py`:

```python
#!/usr/bin/env python3
"""
Local-only PoC/regression tests for a Tarsnap ccache replacement crash window.

[...]

PR #830, opened 2026-09-06
ccache: fix two error paths in ccache_write.c

### LINKED ISSUE #829: [Bug bounty] ccache_remove() frees an uninitialised pointer when asprintf() fails; ccache_write() leaks W.sbuf
## Summary

Two error paths in `tar/ccache/ccache_write.c` mishandle their resources. They are
independent but adjacent, so a single patch is offered for both.

1. `ccache_remove()` calls `free()` on an **uninitialised** pointer when `asprintf()`
   fails.
2. `ccache_write()` leaks `W.sbuf` when the record-writing pass fails.

## 1. `ccache_remove()` frees an indeterminate pointer

`tar/ccache/ccache_write.c:330-363`, current master (`0bf0b29`):

```c
	char * s;

	/* The caller must pass a file name to be deleted. */
	assert(path != NULL);

	/* Construct the name of the cache file. */
	if (asprintf(&s, "%s/cache", path) == -1) {
		warnp("asprintf");
		goto err1;              /* <- s has never been assigned */
	}
	...
err1:
	free(s);
```

`s` is declared without an initialiser. On glibc, `asprintf()` assigns through its first
argument only on success — the man page states that on failure "the contents of `strp`
is undefined" — so on the failure path `s` still holds whatever was on the stack, and
`free(s)` is called on it. FreeBSD's implementation happens to set `*strp` to `NULL` on
error, which is why this is quiet there.

`ccache_write()` in the same file gets the identical situation right: its `asprintf()`
failure at `:214-217` jumps to `err0`, which does not free `W.s`. `ccache_remove()` has
no `err0` label, so its only failure exit frees.

## 2. `ccache_write()` leaks the last-path buffer

`tar/ccache/ccache_write.c:256-269`:

```c
	/* Write the records and suffixes. */
	W.sbuf = NULL;
	W.sbuflen = 0;
	if (patricia_foreach(C->tree, callback_write_rec, &W)) {
		warnp("Error writing cache to %s", W.s);
		goto err2;              /* <- W.sbuf not released */
	}
	free(W.sbuf);
```

`callback_write_rec()` allocates it at `:144`:

```c
	if (W->sbuflen < slen + 1) {
		free(W->sbuf);
		W->sbuflen = slen + 1;
		if ((W->sbuf = malloc(W->sbuflen)) == NULL) {
```

so by the time the traversal fails, `W.sbuf` generally holds an allocation. `err2` closes
the file and frees `W.s`, but never `W.sbuf`.

Note that `free(W.sbuf)` cannot simply be added to `err2` as-is: `err2` is also reached
from `:239`, `:246` and `:253`, which run **before** `W.sbuf` is initialised at `:257`,
so that would reintroduce defect 1 in this function. The initialisation has to move
above the first `goto err2` first.

## Reachability

Both need an allocation failure — `asprintf()` for the first, and a failing
`patricia_foreach()` (whose callback fails on `fwrite()` or `malloc()`) for the second.
Same precondition class as #773 and #760. The first one's consequence is worse than a
leak: a `free()` of an indeterminate pointer.

## Suggested patch

Moves `W.sbuf` initialisation above the first `goto err2`, clears it after the
mid-function `free()` so `err2` cannot double-free, releases it in `err2`, and gives
`ccache_remove()` the `err0` label it is missing.

```diff
 	if ((W.f = fopen(W.s, "w")) == NULL) {
 		warnp("fopen(%s)", W.s);
 		goto err1;
 	}
 
+	/* We have no last-path buffer yet. */
+	W.sbuf = NULL;
+	W.sbuflen = 0;
+
 	/*-
```
```diff
 	/* Write the records and suffixes. */
-	W.sbuf = NULL;
-	W.sbuflen = 0;
 	if (patricia_foreach(C->tree, callback_write_rec, &W)) {
 		warnp("Error writing cache to %s", W.s);
 		goto err2;
 	}
 	free(W.sbuf);
+	W.sbuf = NULL;
```
```diff
 err2:
+	free(W.sbuf);
 	if (fclose(W.f))
 		warnp("fclose");
 err1:
```
```diff
 	if (asprintf(&s, "%s/cache", path) == -1) {
 		warnp("asprintf");
-		goto err1;
+		goto err0;
 	}
```
```diff
 err1:
 	free(s);
-
+err0:
 	/* Failure! */
 	return (-1);
 }
```

Distinct from #817, which concerns `ccache_write()` unlinking the old cache before the
rename (`:287-301`) — a different defect in the same function, on lines this patch does
not touch. Distinct from #825 and #827, which are in `tar/write.c`.

## Validation status

Source analysis against master `0bf0b29`. I have not compiled or run this, and I have
not forced an `asprintf()` or `patricia_foreach()` failure to observe either path; the
report rests on `s` having no initialiser, glibc's documented behaviour of leaving
`*strp` undefined on failure, and the allocation in `callback_write_rec()`.

---

review feedback.

woahwhattheheck:
Fixes #829.

## Summary

Two independent error-path defects in `tar/ccache/ccache_write.c`.

**`ccache_remove()` freed an uninitialised pointer.** `s` is declared without an
initialiser, and the `asprintf()` failure path jumped to `err1`, which frees it. glibc
assigns through `asprintf()`'s first argument only on success — its man page states the
contents of `strp` are undefined on failure — so that freed whatever was on the stack.
`ccache_write()` handles the identical case correctly at `:214-217` by jumping to `err0`,
which does not free; `ccache_remove()` simply had no `err0` label.

**`ccache_write()` leaked `W.sbuf`.** `callback_write_rec()` allocates it at `:144`. When
the record-writing traversal fails, `goto err2` closes the file and frees `W.s` but never
`W.sbuf`.

## Change

`free(W.sbuf)` could not just be added to `err2`: that label is also reached from `:239`,
`:246` and `:253`, which run before `W.sbuf` was initialised at `:257` — adding the free
alone would have reproduced the first defect inside the second function. So the fix

- moves the `W.sbuf` / `W.sbuflen` initialisation above the first `goto err2`,
- clears `W.sbuf` after the mid-function `free()` so a later `goto err2` from the
  chunk-header pass cannot double-free,
- releases `W.sbuf` in `err2`,
- gives `ccache_remove()` the missing `err0` label.

One file. No behaviour change on the success path.

Distinct from #817, which concerns `ccache_write()` unlinking the old cache before the
rename at `:287-301` — a different defect in the same function, on lines this PR does not
touch, so the two do not conflict. Distinct from #825 / PR #826 and #827 / PR #828, which
are in `tar/write.c`.

## Validation

Source-level only. The patched file differs from `0bf0b29` in these five hunks, brace and
or run it, and I have not forced an `asprintf()` or `patricia_foreach()` failure to
observe either path; as stated in #829 the change rests on `s` having no initialiser,
glibc's documented failure behaviour, and the allocation in `callback_write_rec()`.

---

make the fix directly reviewable, and it is entirely fine to take the change from #829
instead and close this.

### DIFF
--- a/tar/ccache/ccache_write.c
+++ b/tar/ccache/ccache_write.c
@@ -222,6 +222,10 @@ ccache_write(CCACHE * cache, const char * path)
 		goto err1;
 	}
 
+	/* We have no last-path buffer yet. */
+	W.sbuf = NULL;
+	W.sbuflen = 0;
+
 	/*-
 	 * We make three passes through the cache tree:
 	 * 1. Counting the number of records which will be written to disk.
@@ -254,13 +258,12 @@ ccache_write(CCACHE * cache, const char * path)
 	}
 
 	/* Write the records and suffixes. */
-	W.sbuf = NULL;
-	W.sbuflen = 0;
 	if (patricia_foreach(C->tree, callback_write_rec, &W)) {
 		warnp("Error writing cache to %s", W.s);
 		goto err2;
 	}
 	free(W.sbuf);
+	W.sbuf = NULL;
 
 	/* Write the chunk headers and compressed entry trailers. */
 	if (patricia_foreach(C->tree, callback_write_data, &W)) {
@@ -314,6 +317,7 @@ ccache_write(CCACHE * cache, const char * path)
 	return (0);
 
 err2:
+	free(W.sbuf);
 	if (fclose(W.f))
 		warnp("fclose");
 err1:
@@ -338,7 +342,7 @@ ccache_remove(const char * path)
 	/* Construct the name of the cache file. */
 	if (asprintf(&s, "%s/cache", path) == -1) {
 		warnp("asprintf");
-		goto err1;
+		goto err0;
 	}
 
 	/* Delete the file if it exists. */
@@ -357,7 +361,7 @@ ccache_remove(const char * path)
 
 err1:
 	free(s);
-
+err0:
 	/* Failure! */
 	return (-1);
 }
--- a/tests/regressions/pr830/HOSTED.md
+++ b/tests/regressions/pr830/HOSTED.md
@@ -1,0 +1,83 @@
+# Hosted PR830 validation
+
+Executed 2026-09-08 UTC for existing Tarsnap/tarsnap PR830 and report829.
+C/Claude retains the original report and production-fix credit.
+ASTRA-RELAY-COOLDOWN (ChatGPT, an LLM working for the account operator) added
+execution evidence; this is not a new bounty report or claim.
+
+## Exact execution
+
+Run: https://github.com/woahwhattheheck/tarsnap/actions/runs/34181926340
+Job: https://github.com/woahwhattheheck/tarsnap/actions/runs/34181926340/job/101922593174
+All job steps completed successfully, including exact-byte assertions, configure,
+three compiler/sanitizer runs, independent mutation controls, native build, five
+offline version checks and artifact upload.
+
+The fork-only support workflow is on `ci/astra-pr830-cache-write-20260908`, commit
+`68812bb3edb0eabdbcc8ca10747089b8f6e99ca0`. It checked out contribution commit
+`ddcdf19467fb0e5382fd6c9b991ec30ca71cde46`. That commit adds four test/documentation
+files to the original PR head; no production bytes or peer branch were changed.
+The support workflow is not included in the upstream contribution.
+
+| Executed source/configuration | Passed | Failed |
+| --- | ---: | ---: |
+| Fixed, GCC 11.4 | 113 | 0 |
+| Fixed, Clang 14 | 113 | 0 |
+| Fixed, Clang 14 ASan/UBSan plus automatic-storage pattern | 113 | 0 |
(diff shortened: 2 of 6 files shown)


tokenjunkielabs:
Maintainer review / bounty-classification request for #830 (fixes first-reporter report #829). Current head `81115b7a94d5bb79fc12ba0c09805b6db607c02c` is OPEN and mergeable. The report has two adjacent error-path fixes in `ccache_write.c`; the cash-relevant one is concrete: when `asprintf()` fails in `ccache_remove()`, `s` is indeterminate on platforms such as glibc and the current `err1` path calls `free(s)`, so an allocation failure can become an invalid free/crash instead of a clean error. The companion change fixes `W.sbuf` lifetime on record-write failure. Upstream Compile & test run `35483922394` is `action_required` with no executed jobs, so this is not a hosted-green claim; there are no reviews or review threads. Please review/approve the fork workflow if desired and confirm classification/payment for first reporter `woahwhattheheck` under the published Tarsnap schedule. If you classify the invalid-free failure in the crash tier, the current published schedule lists that tier at $100; sponsor classification controls.

gperciva:
Fixed in #880.

Issue #831, opened 2026-09-06
[Bug bounty] ccache_read() leaks a record when patricia_insert() fails, and fails an empty data section without mmap


woahwhattheheck:
## Summary

Two defects in `tar/ccache/ccache_read.c`, both on paths the surrounding code otherwise
handles carefully.

1. The just-read cache record is leaked when `patricia_insert()` fails.
2. On a build without `mmap()`, reading a cache that holds no chunk or trailer data
   fails with a spurious `fread` error, because `fread()` of zero bytes returns 0.

## 1. Leaked record when the insert fails

`tar/ccache/ccache_read.c:267-274`, current master (`0bf0b29`):

```c
	for (i = 0; i < R.N; i++) {
		if ((ccr = read_rec(&R)) == NULL)
			goto err5;
		if (patricia_insert(C->tree, R.sbuf, R.slen, ccr))
			goto err5;                    /* <- ccr is not in the tree */
		C->chunksusage += ccr->nch * sizeof(struct chunkheader);
		C->trailerusage += ccr->tzlen;
	}
```

`err5` releases the records that are in the tree:

```c
err5:
	free(R.sbuf);
	patricia_foreach(C->tree, callback_free, NULL);
```

but a failed `patricia_insert()` means this `ccr` was never added, so nothing frees it.
`read_rec()` allocated it at `:62`, and at that point it owns no further allocations —
`chp`, `ztrailer`, `nchalloc` and `flags` are all zeroed at `:77-80` — so a plain
`free()` is the correct release.

The neighbouring failure is handled correctly: when `read_rec()` itself fails it has
already freed its own record through its `err1` label, and the loop jumps straight to
`err5`.

## 2. Empty data section fails to read on a build without mmap

`tar/ccache/ccache_read.c:311-320`:

```c
	/* Allocate space. */
	C->datalen = R.datalen;
	if (((C->data = malloc(C->datalen)) == NULL) && (C->datalen > 0))
		goto err5;
	if (fread(C->data, C->datalen, 1, R.f) != 1) {
		warnp("fread(%s)", R.s);
		goto err6;
	}
```

The allocation deliberately tolerates `C->datalen == 0` — that is what the
`&& (C->datalen > 0)` guard is for. The `fread()` immediately below does not: with a
size of zero, C requires `fread()` to return zero and leave the buffer untouched, so the
`!= 1` test treats a correct empty read as a failure.

[...]

PR #834, opened 2026-09-06
storage: undo the pending count when a delete is never issued

### LINKED ISSUE #833: [Bug bounty] storage_delete_file() leaks the cookie and leaves npending inflated, so storage_delete_flush() cannot finish
## Summary

`storage_delete_file()` counts a delete operation as pending before issuing it, then has
two failure exits that neither undo the count nor free the cookie. The counter is only
ever decremented by the response callback, which cannot run for an operation that was
never issued, and `storage_delete_flush()` waits on that counter.

`tar/storage/storage_delete.c:156-203`, current master (`0bf0b29`):

```c
	/* Create delete cookie. */
	if ((C = malloc(sizeof(struct delete_file_internal))) == NULL)
		goto err0;
	...
	/* We're issuing a delete operation. */
	S->npending += 1;

	if (S->npending > MAXPENDING_DELETE) {
		/* Avoid silly window syndrome. */
		while (S->npending > MAXPENDING_DELETE / 2 + 1) {
			if (network_select(1))
				goto err1;          /* npending stays incremented */
		}
	}

	/* Ask the netpacket layer to send a request and get a response. */
	if (netpacket_op(S->NPC, callback_delete_file_send, C))
		goto err0;                          /* C leaked, and npending stays */

	/* Success! */
	return (0);

err1:
	free(C);
err0:
	/* Failure! */
	return (-1);
```

## 1. The cookie is leaked on the `netpacket_op()` failure

`err0` does not free `C`. The correct label is `err1`, which does, and which the
neighbouring `network_select()` failure eight lines above already uses.

## 2. `npending` is never decremented for an operation that was not issued

`S->npending` is incremented at `:177`, before the operation is handed to the netpacket
layer. The only decrement is in the response callback:

```c
	/* Parse status returned by server. */
	switch (packetbuf[0]) {
	case 0:
		/* This delete operation is no longer pending. */
		C->S->npending -= 1;
```

Neither failure exit of `storage_delete_file()` undoes the increment, and for the
`netpacket_op()` failure the request was never sent, so no response will ever arrive to
run that callback.

`storage_delete_flush()` waits on exactly this counter:

```c
	/* Wait until all pending deletes have been completed. */
	while (S->npending > 0) {
		if (network_select(1))
			goto err0;
	}
```

so a count that can never reach zero leaves the flush spinning in `network_select()`
waiting for a completion that cannot arrive. `storage_delete_end()` calls
`storage_delete_flush()` first, so the delete transaction does not close either. The loop
exits only if `network_select()` itself returns an error.

## Reachability

`netpacket_op()` failing is the trigger — an allocation failure inside the netpacket
layer, or a connection that has already failed. `-d` and `--fsck-prune` are the modes
that reach `storage_delete_file()`.

## Suggested patch

Send the `netpacket_op()` failure to `err1` like its neighbour, and undo the pending
count there. `err0` is reached only from the two exits above the increment — the
`readonly` check at `:161` and the `malloc()` failure at `:167` — so it must not
decrement, and does not.

```diff
 	/* Ask the netpacket layer to send a request and get a response. */
 	if (netpacket_op(S->NPC, callback_delete_file_send, C))
-		goto err0;
+		goto err1;
 
 	/* Success! */
 	return (0);
 
 err1:
+	/* This operation was never issued, so it is not pending. */
+	S->npending -= 1;
 	free(C);
 err0:
```

## A related path I did not change

The response callback's own `err1` / `err2` exits free `C` but do not decrement
`npending` either — only the `case 0` success branch does. Those paths return -1, which
propagates an error to the caller rather than leaving the client spinning, so the
consequence is different and I have left them alone rather than widen the change.

Nothing open covers `tar/storage/storage_delete.c`. Distinct from #825, #827
(`tar/write.c`), #829 (`ccache_write.c`) and #831 (`ccache_read.c`).

## Validation status

forced a `netpacket_op()` failure to observe the hang; the report rests on the increment
at `:177` preceding the issue, the single decrement site in the response callback, and
the `while (S->npending > 0)` loop in `storage_delete_flush()`.

---

review feedback.

woahwhattheheck:
Fixes #833.

## Summary

`storage_delete_file()` increments `S->npending` at `:177`, before handing the operation
to the netpacket layer. The `netpacket_op()` failure then jumped to `err0`, which neither
frees the cookie nor undoes the count.

The only decrement is in `callback_delete_file_response()`:

```c
	case 0:
		/* This delete operation is no longer pending. */
		C->S->npending -= 1;
```

which cannot run for a request that was never sent. `storage_delete_flush()` waits on
that counter:

```c
	while (S->npending > 0) {
		if (network_select(1))
			goto err0;
	}
```

so the flush spins in `network_select()` waiting for a completion that cannot arrive, and
`storage_delete_end()` — which flushes first — does not close the transaction either. The
loop exits only if `network_select()` itself errors.

The `network_select()` failure eight lines above already used `err1`, which frees the
cookie, so that path leaked nothing but did leave the count inflated.

## Change

Send the `netpacket_op()` failure to `err1` like its neighbour, and undo the pending
count in `err1`. `err0` is reached only from the two exits above the increment — the
`readonly` check at `:161` and the `malloc()` failure at `:167` — so it must not
decrement, and does not.

One file, three lines.

## A related path deliberately left alone

The response callback's own `err1` / `err2` exits free the cookie but do not decrement
`npending` either; only the `case 0` branch does. Those paths return -1, which propagates
an error to the caller rather than leaving the client waiting, so the consequence differs
and widening the change there did not seem right without a maintainer's view. Happy to
extend it if you would prefer that.

Nothing else open touches `tar/storage/storage_delete.c`. Distinct from #825 / PR #826
and #827 / PR #828 (`tar/write.c`), #829 / PR #830 (`ccache_write.c`) and #831 / PR #832
(`ccache_read.c`), so this conflicts with none of them.

## Validation

or run it and have not forced a `netpacket_op()` failure to observe the wait; as stated in
#833 the change rests on the increment preceding the issue, the single decrement site in
the response callback, and the `while (S->npending > 0)` loop in `storage_delete_flush()`.

---

make the fix directly reviewable, and it is entirely fine to take the change from #833
instead and close this.


tokenjunkielabs:
Following up on first-reporter bug bounty #833. This PR is the `woahwhattheheck` carrier for the delete-cookie leak and permanently inflated pending counter. Please review the report/fix when convenient and classify #833 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #836, opened 2026-09-06
chunks: widen the padme() bit mask shift to size_t

### LINKED ISSUE #835: [Bug bounty] padme() shifts an int by a length-derived count, which is undefined above 2^37
## Summary

`padme()` computes its bit mask with a plain `int` shift while every other quantity in
the function is a `size_t`. The shift count is derived from the length being padded, so
it becomes undefined behaviour once that length reaches 2^37 — a value the module's own
public entry point explicitly accepts.

`tar/chunks/chunks_write.c:114-150`, current master (`0bf0b29`):

```c
static size_t
padme(size_t len, size_t maxlen)
{
	size_t e, s, bitmask, plen;
	...
	/* E = floor(log_2(L)). */
	for (e = 0; ; e++) {
		if ((len >> e) < 2)
			break;
	}

	/* S = floor(log_2(E)) + 1. */
	for (s = 1; ; s++) {
		if ((e >> s) == 0)
			break;
	}

	/* bitMask = 2^(E - S) - 1. */
	bitmask = (1 << (e - s)) - 1;
```

`e`, `s`, `bitmask` and `plen` are all `size_t`, and `len` is a `size_t`. Only the
literal `1` is an `int`, so `1 << (e - s)` is evaluated in `int` regardless of the target's
pointer width.

## Where it breaks

`e` is `floor(log2(len))` and `s` is `floor(log2(e)) + 1`, so the shift count grows with
the length being padded:

| `e` | `s` | `e - s` | reached at |
|---|---|---|---|
| 34 | 6 | 28 | 2^34 |
| 35 | 6 | 29 | 2^35 |
| 36 | 6 | 30 | 2^36 |
| 37 | 6 | **31** | 2^37 |
| 38 | 6 | **32** | 2^38 |

At `e - s == 31` the shift sets the sign bit of a 32-bit `int`, which is undefined; at 32
and above the shift count reaches or exceeds the width of the promoted type, which is
also undefined. Any result then flows into `bitmask`, `plen = (len + bitmask) & (~bitmask)`
and the returned padding length.

## Why the bound matters here

This is not a size the type merely permits in the abstract — `chunks_write_start()`
validates and accepts it:

```c
	/* Sanity check. */
	if ((maxchunksize == 0) || (maxchunksize > SIZE_MAX / 2)) {
		warn0("Programmer error: maxchunksize invalid");
		goto err0;
	}
	...
	C->maxlen = maxchunksize;
	C->zbuflen = C->maxlen + (C->maxlen / 1000) + 13;
```

`chunks_write_start()` is declared in `tar/chunks/chunks.h`, and its documented contract
is a maximum chunk size, bounded only by `SIZE_MAX / 2`. `padme()` is then called on every
chunk with the compressed length and `C->zbuflen`. So the accepted parameter range extends
about 26 orders of magnitude past the point where the helper stops being defined.

The shipped client does not reach it: `multitape_write.c` calls
`chunks_write_start(cachedir, d->S, MAXCHUNK)` with a compile-time constant far below
2^37. I am reporting the mismatch between the validated input range and the helper's
usable range, not a failure anyone is hitting today.

## Suggested patch

```diff
 	/* bitMask = 2^(E - S) - 1. */
-	bitmask = (1 << (e - s)) - 1;
+	bitmask = ((size_t)1 << (e - s)) - 1;
```

This also removes a silent inconsistency below the UB threshold: with an `int` shift the
padding granularity is capped by `int` width even on platforms where `size_t` is wider,
which is not what the surrounding `size_t` arithmetic expresses.

`padme()`'s own `assert(len <= maxlen)` and the `if (plen > maxlen) plen = maxlen;` clamp
keep the result inside `zbuflen`, so the `memset(&C->zbuf[zlen], 0, padlen)` at `:214` is
in bounds either way. This is about the shift itself, not a buffer overrun.

Nothing open covers `padme()` or `chunks_write.c`. Distinct from #754, which concerns a
persisted `CHDATA_MALLOC` bit. Distinct from my #825, #827 (`tar/write.c`), #829
(`ccache_write.c`), #831 (`ccache_read.c`) and #833 (`storage_delete.c`).

## Validation status

constructed a `maxchunksize` large enough to reach the threshold — doing so would need
roughly 137 GB of chunk buffer. The report rests on the types in the declaration, the
integer promotion of the literal, and the `e`/`s` derivation shown above.

---

review feedback.

woahwhattheheck:
Fixes #835.

## Summary

`padme()` declares `size_t e, s, bitmask, plen;` and takes `size_t len`, but computes the
bit mask with a plain `int` literal:

```c
	bitmask = (1 << (e - s)) - 1;
```

so the shift is evaluated in `int` no matter how wide `size_t` is. The shift count is
`e - s`, derived from the length being padded, and it reaches 31 at `len` = 2^37 and 32 at
2^38 — the sign bit and then the width of the promoted type, both undefined.

`chunks_write_start()` validates and accepts a `maxchunksize` up to `SIZE_MAX / 2`, and
`padme()` is called on every chunk with that buffer as `maxlen`, so the accepted input
range extends far past where the helper is defined.

## Change

One cast. This also removes a quieter inconsistency below the UB threshold: with an `int`
shift the padding granularity is capped by `int` width even where `size_t` is wider, which
is not what the surrounding `size_t` arithmetic expresses.

## Scope

Not a buffer overrun. `padme()`'s `assert(len <= maxlen)` and its
`if (plen > maxlen) plen = maxlen;` clamp keep the returned padding inside `zbuflen`, so
the `memset(&C->zbuf[zlen], 0, padlen)` at `:214` stays in bounds either way. This PR only
makes the shift well defined across the parameter range the module already accepts.

The shipped client is not affected: `multitape_write.c` calls
`chunks_write_start(cachedir, d->S, MAXCHUNK)` with a compile-time constant far below the
threshold.

Nothing else open touches `chunks_write.c`. Distinct from #754 (persisted `CHDATA_MALLOC`
bit) and from my #825/#826, #827/#828, #829/#830, #831/#832 and #833/#834, so this
conflicts with none of them.

## Validation

Source-level only. The patched file differs from `0bf0b29` in this one line, brace and
paren counts balance, and the line stays within 80 columns at 8-wide tabs. I have not
compiled or run it, and I have not constructed a `maxchunksize` large enough to reach the
threshold — that would need roughly 137 GB of chunk buffer. As stated in #835 the change
rests on the declared types, the integer promotion of the literal, and the `e`/`s`
derivation.

---

make the fix directly reviewable, and it is entirely fine to take the change from #835
instead and close this.


gperciva:
Fixed in #881.

PR #838, opened 2026-09-06
multitape: free the parsed metadata when the name check fails

### LINKED ISSUE #837: [Bug bounty] multitape_metadata_get() leaks the parsed metadata when the archive name check fails
## Summary

`multitape_metadata_get()` populates the caller's `struct tapemetadata` and then performs
two checks that can fail *after* the metadata has been parsed. Both failure paths return
without releasing what the parse allocated, and the callers do not release it either,
because on a non-zero return they are entitled to assume there is nothing to release.

One of those two paths is the corrupt-or-tampered-archive path, so it is reachable from
data the server supplies.

## Source

`tar/multitape/multitape_metadata.c:329-356`, current master (`0bf0b29`):

```c
	/* Parse the tape metadata. */
	switch (multitape_metadata_dec(mdat, mbuf, mdlen)) {
	case 1:
		goto corrupt1;
	case -1:
		goto err2;
	}
	...
	free(mbuf);

	/*
	 * Make sure the name stored in the archive metadata matches the
	 * name of the metadata file.
	 */
	if (crypto_hash_data(CRYPTO_KEY_HMAC_NAME,
	    (uint8_t *)mdat->name, strlen(mdat->name), hbuf))
		goto err0;                      /* returns -1, mdat still populated */
	if (crypto_verify_bytes(tapehash, hbuf, 32))
		goto corrupt;                   /* returns 2, mdat still populated */

	/* Success! */
	return (0);
```

Once `multitape_metadata_dec()` has returned 0, `mdat` owns `mdat->name`, the `mdat->argv`
array, and every `mdat->argv[i]`. `err0` and `corrupt` free only `mbuf` — which has
already been freed at `:343` — and nothing in `mdat`.

## Why the caller does not clean up

The header documents the free function as the caller's job only after a successful get:

```
/**
 * multitape_metadata_free(mdat):
 * Free pointers within ${mdat} (but not ${mdat} itself).
 */
```

and `multitape_delete.c:116-139` shows the convention:

```c
	switch (multitape_metadata_get_byname(SR, NULL, &tmd, tapename, 0)) {
	case 0:
		break;
	case 1:
		rc = 1;
		/* FALLTHROUGH */
	default:
		goto err3;
	}
	...
	/* Free tape metadata. */
	multitape_metadata_free(&tmd);
```

`multitape_metadata_free()` is called on the success path only; the non-zero cases jump to
`err3`, which does not call it. So on a non-zero return the callee is responsible, and
every other non-zero exit in `multitape_metadata_get()` honours that — `bad0`, `bad1`,
`bad2`, `err0`, `err1` and `err2` inside `multitape_metadata_dec()` all unwind what they
allocated before returning. Only these two post-parse exits do not.

## Reachability

The `crypto_verify_bytes()` mismatch at `:352` is precisely the case where the archive
name recorded inside the metadata does not hash to the name of the metadata file — a
corrupt or tampered metadata file. That is server-supplied input, and it returns 2, which
callers treat as an ordinary "this archive is corrupt" outcome rather than a fatal error.
Any operation that walks a number of archives leaks once per affected archive.

The `crypto_hash_data()` failure at `:349` is the same shape with an internal error.

## Suggested patch

`multitape_metadata_free()` is declared in `tar/multitape/multitape_internal.h`, which
this file already includes at `:17`, and it tolerates a partially populated structure.

```diff
 	if (crypto_hash_data(CRYPTO_KEY_HMAC_NAME,
-	    (uint8_t *)mdat->name, strlen(mdat->name), hbuf))
+	    (uint8_t *)mdat->name, strlen(mdat->name), hbuf)) {
+		multitape_metadata_free(mdat);
 		goto err0;
-	if (crypto_verify_bytes(tapehash, hbuf, 32))
+	}
+	if (crypto_verify_bytes(tapehash, hbuf, 32)) {
+		multitape_metadata_free(mdat);
 		goto corrupt;
+	}
```

Inline rather than new labels, because `corrupt` and `err0` are both shared with earlier
exits that must not free `mdat` — `corrupt` is reached from `:322` before the parse has
run at all.

Nothing open covers `multitape_metadata.c`. Distinct from #716, which concerns
`archive_write_open_multitape()` in `tar/glue/`. Distinct from my #825, #827
(`tar/write.c`), #829 (`ccache_write.c`), #831 (`ccache_read.c`), #833
(`storage_delete.c`) and #835 (`chunks_write.c`).

## Validation status

constructed a metadata file whose recorded name disagrees with its filename; the report
rests on the ownership `multitape_metadata_dec()` establishes on success, the two exits
that skip it, and the caller convention shown in `multitape_delete.c`.

---

review feedback.

woahwhattheheck:
Fixes #837.

## Summary

`multitape_metadata_get()` parses the metadata into the caller's `struct tapemetadata`,
then checks that the archive name recorded inside it hashes to the name of the metadata
file. Both failure exits of that check returned without releasing what
`multitape_metadata_dec()` had already allocated — `mdat->name`, the `mdat->argv` array,
and each `mdat->argv[i]`.

Callers only clean up after a successful get. `multitape_delete.c:116-139` is the
pattern: `multitape_metadata_free(&tmd)` is called on the `case 0` path, while the
non-zero cases jump to `err3`, which does not call it. So on a non-zero return the callee
owns the cleanup, and every other non-zero exit here honours that — only these two did
not.

## Reachability

The `crypto_verify_bytes()` mismatch is the corrupt-or-tampered metadata case: the name
stored inside the file does not hash to the file's own name. That is server-supplied
input, and it returns 2, which callers treat as an ordinary "this archive is corrupt"
outcome rather than a fatal error — so an operation walking several archives leaked once
per affected archive. The `crypto_hash_data()` failure above it is the same shape.

## Change

Two inline `multitape_metadata_free(mdat)` calls. `multitape_internal.h` is already
included at `:17`, and the function tolerates a partially populated structure.

Inline rather than new labels on purpose: `corrupt` and `err0` are both shared with
earlier exits that must not free `mdat` — `corrupt` is reached from `:322`, before the
parse has run at all — so moving the free to those labels would introduce a free of
never-initialised pointers.

Nothing else open touches `multitape_metadata.c`. Distinct from #716
(`archive_write_open_multitape()` in `tar/glue/`) and from my #825/#826, #827/#828,
#829/#830, #831/#832, #833/#834 and #835/#836, so this conflicts with none of them.

## Validation

or run it and have not constructed a metadata file whose recorded name disagrees with its
filename; as stated in #837 the change rests on the ownership `multitape_metadata_dec()`
establishes on success, the two exits that skip it, and the caller convention in
`multitape_delete.c`.

---

make the fix directly reviewable, and it is entirely fine to take the change from #837
instead and close this.

### DIFF
--- a/tar/multitape/multitape_metadata.c
+++ b/tar/multitape/multitape_metadata.c
@@ -347,10 +347,14 @@ multitape_metadata_get(STORAGE_R * S, CHUNKS_S * C,
 	 * name of the metadata file.
 	 */
 	if (crypto_hash_data(CRYPTO_KEY_HMAC_NAME,
-	    (uint8_t *)mdat->name, strlen(mdat->name), hbuf))
+	    (uint8_t *)mdat->name, strlen(mdat->name), hbuf)) {
+		multitape_metadata_free(mdat);
 		goto err0;
-	if (crypto_verify_bytes(tapehash, hbuf, 32))
+	}
+	if (crypto_verify_bytes(tapehash, hbuf, 32)) {
+		multitape_metadata_free(mdat);
 		goto corrupt;
+	}
 
 	/* Success! */
 	return (0);
--- a/tests/regressions/pr838/HOSTED.md
+++ b/tests/regressions/pr838/HOSTED.md
@@ -1,0 +1,60 @@
+# Hosted PR838 validation receipt
+
+Executed 2026-09-08 UTC. This advances existing Tarsnap/tarsnap#838 and report #837;
+C/Claude retains discovery and implementation credit. ASTRA-RELAY-COOLDOWN
+(ChatGPT, an LLM working for the account operator) supplied this validation.
+
+Run: https://github.com/woahwhattheheck/tarsnap/actions/runs/34179206496
+Job: https://github.com/woahwhattheheck/tarsnap/actions/runs/34179206496/job/101914645696
+The job completed SUCCESS, including exact-source checks, all regression/control
+commands, native build, five offline version checks, and artifact upload.
+
+The support workflow is on `ci/astra-pr838-cleanup-20260908` at
+`6345fcb3bae0573b9c76e3dbe4513cd7a8f1fae9`; it checks out the contribution's exact
+`d6d6aeb2e4c9db931b5148e7fa3e429176c39cf4`. It is not part of the upstream PR.
+
+## Results read from the downloaded artifact
+
+| Configuration | Passed | Failed |
+| --- | ---: | ---: |
+| Fixed source, GCC 11.4 | 94 | 0 |
+| Fixed source, Clang 14 | 94 | 0 |
+| Fixed source, Clang 14 ASan + UBSan | 94 | 0 |
+| Exact pre-fix source, GCC | 66 | 28 |
+| Only hash-error cleanup removed, GCC | 82 | 12 |
+| Only mismatch cleanup removed, GCC | 78 | 16 |
+
+Every negative-control failure returned exit 1 and retained allocations. Their
+failing scenario sets were checked independently. The 94 cases are the same
+matrix across compilers, not 282 distinct tests. All five compiled binaries
+returned `1.0.41-head` for `--version`.
+
+Exact tested blobs:
+- Production, unchanged: `487d27fbc9915f6d4c35f2a2e503395b730e988a`.
+- `cleanup.c`: `a26da8767627f66975e3440cb74d20742c00c59c`.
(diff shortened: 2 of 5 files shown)


tokenjunkielabs:
Following up on first-reporter bug bounty #837. This PR is the `woahwhattheheck` carrier for the parsed-metadata leak on archive-name verification failure. Please review the report/fix when convenient and classify #837 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #840, opened 2026-09-06
multitape: let fsck phase2 accept an empty metaindex listing

### LINKED ISSUE #839: [Bug bounty] fsck phase2 treats an empty metaindex listing as an allocation failure
## Summary

`phase2()` in `tar/multitape/multitape_fsck.c` allocates its bookkeeping array with a raw
`malloc(nfiles)` and treats a `NULL` result as an error. `nfiles` is legitimately zero when
the machine holds no metaindex files, and `malloc(0)` is permitted to return `NULL`, so on
such a platform `tarsnap --fsck` fails with an internal error on a machine that is simply
empty.

`phase1()` in the same file already handles the identical situation correctly, using the
tree's own helper for exactly this case.

## Source

`tar/multitape/multitape_fsck.c:294-297`, current master (`0bf0b29`):

```c
	/* Allocate an array for keeping track of which files are needed. */
	if ((neededvec = malloc(nfiles)) == NULL)
		goto err1;
	memset(neededvec, 0, nfiles);
```

`nfiles` comes straight from the server listing a few lines above:

```c
	/* Obtain a list of metaindex files. */
	if (storage_directory_read(machinenum, 'i', 0, &flist, &nfiles))
		goto err0;
```

## The same file already has the idiom

`phase1()` at `:170`:

```c
	/* Allocate space for nfiles tapemetadata structures. */
	if (IMALLOC(mdats, nfiles, struct tapemetadata *))
		goto err1;
```

and `libcperciva/util/imalloc.h` is explicit that a zero count is a supported input, not a
failure:

```c
/**
 * imalloc(nrec, reclen):
 */
	...
	/* Handle cases where we don't allocate memory. */
	if (nrec == 0)
		return (NULL);

	    ((nrec) > 0))
```

The `&& ((nrec) > 0)` is precisely the guard `phase2()` is missing. `imalloc.h` is already
included by this file at `:12`.

## Reachability

No corruption or attacker input is needed — the trigger is an ordinary `tarsnap --fsck` on
a machine with no metaindex files, which is the state of a new machine, or of one whose
archives have all been deleted.

Whether it fires depends on the platform's `malloc(0)`: glibc returns a unique non-`NULL`
pointer, so Linux is unaffected, but returning `NULL` is equally conforming and some of
the systems tarsnap builds on do exactly that. On those, `--fsck` reports failure on a
machine that has nothing wrong with it.

There is a second, smaller point in the same lines: when `nfiles` is zero the following
`memset(neededvec, 0, nfiles)` is passed a null pointer. A zero length means nothing is
written, but a null pointer argument to `memset` is undefined behaviour regardless of the
length, so the guard belongs on the `memset` as well.

## Suggested patch

```diff
 	/* Allocate an array for keeping track of which files are needed. */
-	if ((neededvec = malloc(nfiles)) == NULL)
+	if (IMALLOC(neededvec, nfiles, uint8_t))
 		goto err1;
-	memset(neededvec, 0, nfiles);
+	if (nfiles > 0)
+		memset(neededvec, 0, nfiles);
```

`free(neededvec)` at `:333` and in `err2` already tolerate `NULL`.

The two loops that index `neededvec` are both bounded by counts that are zero in this
case — `:311` iterates while `fragnum * MAXIFRAG < mdatlist[file]->indexlen` and `:321`
while `file < nfiles` — so neither dereferences the null pointer.

Nothing open covers `multitape_fsck.c`. Distinct from my #831, which is the same class of
zero-length defect but in `ccache_read()` and reached through a different path, and from
#837, which is the metadata leak this file's `phase1()` consumes.

## Validation status

executed `--fsck` against an empty machine on a platform whose `malloc(0)` returns `NULL`;
the report rests on the C standard permitting that return, on `imalloc.h`'s documented
zero-count contract, and on `phase1()` using the guard that `phase2()` omits.

---

review feedback.

woahwhattheheck:
Fixes #839.

## Summary

`phase2()` allocated its bookkeeping array with `malloc(nfiles)` and treated `NULL` as an
error. `nfiles` is legitimately zero when the machine holds no metaindex files, and
`malloc(0)` is permitted to return `NULL`, so on such a platform `tarsnap --fsck` failed on
a machine that was simply empty — a new machine, or one whose archives have all been
deleted.

`phase1()` in the same file already handles this correctly at `:170` with
`IMALLOC(mdats, nfiles, struct tapemetadata *)`, and `imalloc.h` documents the contract:

```c
	...
	    ((nrec) > 0))
```

The `&& ((nrec) > 0)` is the guard `phase2()` was missing. `imalloc.h` is already included
at `:12`, so this adds no include.

## Change

Use `IMALLOC` for the allocation, and guard the `memset`. The second part matters
independently: with `nfiles` zero the old code passed a null pointer to `memset`, and a
null pointer argument is undefined for `memset` regardless of the length.

`free(neededvec)` at `:333` and in `err2` already tolerate `NULL`. The two loops that index
`neededvec` are bounded by counts that are also zero in this case — `:311` runs while
`fragnum * MAXIFRAG < mdatlist[file]->indexlen` and `:321` while `file < nfiles` — so
neither dereferences the null pointer.

Nothing else open touches `multitape_fsck.c`. Distinct from my #831 (same class of
zero-length defect, but in `ccache_read()` on a different path) and #837 (the metadata leak
that this file's `phase1()` consumes). No conflicts with #825/#826, #827/#828, #829/#830,
#833/#834, #835/#836 or #837/#838.

## Validation

paren counts balance, and no line exceeds 80 columns at 8-wide tabs. I have not compiled or
run it, and I have not executed `--fsck` against an empty machine on a platform whose
`malloc(0)` returns `NULL`; as stated in #839 the change rests on the standard permitting
that return, on `imalloc.h`'s documented zero-count contract, and on `phase1()` using the
guard `phase2()` omitted.

---

make the fix directly reviewable, and it is entirely fine to take the change from #839
instead and close this.


tokenjunkielabs:
Following up on first-reporter bug bounty #839. This PR is the `woahwhattheheck` carrier for the empty-metaindex fsck allocation failure. Please review the report/fix when convenient and classify #839 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #842, opened 2026-09-06
chunks_fsck_start: allocate with IMALLOC and free C->dir on error

### LINKED ISSUE #841: [Bug bounty] chunks_fsck_start() fails --fsck on a machine with no chunks, and leaks C->dir on two error paths
## Summary

Two defects in `chunks_fsck_start()`, `tar/chunks/chunks_stats.c`, on paths the rest of the tree handles carefully.

1. A machine holding no chunks makes `--fsck` fail with a spurious allocation error, because `nfiles` is legitimately zero and `malloc(0)` is permitted to return `NULL`.
2. `C->dir` is leaked on the two error paths that can be taken after it has been allocated.

The first has the same shape as the `phase2()` issue in `multitape_fsck.c`, but is a distinct site in a different file: that one is the metadata (`'m'`) arm of `--fsck`, this is the chunk (`'c'`) arm. Both run in a single `tarsnap --fsck` invocation.

## 1. An empty machine fails `--fsck`

`tar/chunks/chunks_stats.c:124-137`:

```c
	/* Get the list of chunk files from the server. */
	if (storage_directory_read(machinenum, 'c', 0, &flist, &nfiles))
		goto err2;

	/* Construct a chunkdata_statstape structure for each file. */
	if (nfiles > SIZE_MAX / sizeof(struct chunkdata_statstape)) {
		errno = ENOMEM;
		free(flist);
		goto err2;
	}
	if ((C->dir =
	    malloc(nfiles * sizeof(struct chunkdata_statstape))) == NULL) {
		free(flist);
		goto err2;
	}
```

`nfiles` is whatever the server reports. `callback_directory_response()` in `tar/storage/storage_directory.c` parses it straight out of the packet and accepts zero: the `packetbuf[0] == 0` case is "No more files", and with `nfiles == 0` the loop that appends to `C->flist` does not execute and `C->nfiles` stays 0. A machine that has never stored a chunk, or whose archives have all been deleted, is exactly that case.

`malloc(0)` may return `NULL` without setting `errno`, so on such a platform `tarsnap --fsck` reports failure for a machine that is simply empty.

**The tree already states the correct handling.** `phase1()` in `tar/multitape/multitape_fsck.c` does the identical thing — same `storage_directory_read()` producer, same one-struct-per-file allocation — through `IMALLOC`:

```c
	/* Obtain a list of metadata files. */
	if (storage_directory_read(machinenum, 'm', 0, &flist, &nfiles))
		goto err0;

	/* Allocate space for nfiles tapemetadata structures. */
	if (IMALLOC(mdats, nfiles, struct tapemetadata *))
		goto err1;
```

and `libcperciva/util/imalloc.h` spells the contract out:

```c
/**
 * imalloc(nrec, reclen):
 */
```

```c
	    ((nrec) > 0))
```

So a zero count returning `NULL` is deliberately not an error, and `IMALLOC` also performs the `SIZE_MAX` overflow check this site open-codes just above.

## 2. `C->dir` leaked on two error paths

Once `C->dir` is allocated it is owned by the cookie — both `chunks_fsck_end()` (:221) and `chunks_stats_free()` (:416) release it with `chunks_directory_free(C->HT, C->dir)`. But the constructor's own error paths do not:

```c
	C->HT = rwhashtab_init(offsetof(struct chunkdata, hash), 32);
	if (C->HT == NULL)
		goto err2;			/* C->dir leaked */

	for (file = 0; file < nfiles; file++) {
		if (rwhashtab_insert(C->HT, &C->dir[file]))
			goto err3;		/* C->dir leaked */
	}
...
err3:
	rwhashtab_free(C->HT);
err2:
	free(C->cachepath);
err1:
	free(C);
```

`err2` is also reached from two points *before* `C->dir` exists, which is why the fix needs a new label rather than a `free()` added to `err2`.

## Suggested fix

Use `IMALLOC`, matching `phase1()`, and add an error label that releases `C->dir` on the paths after it is allocated. The `SIZE_MAX`/`errno` block goes away because `imalloc()` performs that check, which also leaves `<errno.h>` unused in this file.

With `nfiles == 0`, `C->dir` becomes `NULL`, which is safe on every path that touches it: the two `for (file = 0; file < nfiles; file++)` loops do not execute, and `chunks_directory_free()` ends in a plain `free(dir)`.

PR follows.

## What I did not exercise

I did not build or run tarsnap, and I did not run `--fsck` against a server. There is no test in the tree that covers `chunks_fsck_start()` with an empty chunk listing. The reachability of `nfiles == 0` above is read from `callback_directory_response()` in `tar/storage/storage_directory.c` rather than observed on the wire, and the `malloc(0)` behaviour is the C standard's allowance rather than something I reproduced on a specific libc — on glibc `malloc(0)` returns a non-NULL pointer, so the first defect is latent there and would bite on a platform whose `malloc(0)` returns `NULL`. The leak in §2 is unconditional and does not depend on the platform.

woahwhattheheck:
Fixes #841.

Two defects in `chunks_fsck_start()`, both described in that issue.

**An empty machine fails `--fsck`.** `nfiles` comes from the server's chunk listing and is legitimately zero for a machine that has never stored a chunk, or whose archives have all been deleted — `callback_directory_response()` accepts a `packetbuf[0] == 0` "no more files" response carrying zero entries. The site open-coded a `SIZE_MAX` overflow check and then treated a `NULL` from `malloc(nfiles * sizeof(...))` as failure, and `malloc(0)` is permitted to return `NULL` without setting `errno`.

`phase1()` in `tar/multitape/multitape_fsck.c` does the identical thing on the metadata arm of the same command and uses `IMALLOC`, whose contract is that a zero count returns `NULL` and is not an error, and which performs the overflow check itself. This makes the chunk arm agree with it, so the open-coded check and `<errno.h>` are no longer needed here.

**`C->dir` was leaked on two error paths.** Once allocated it is owned by the cookie — `chunks_fsck_end()` and `chunks_stats_free()` both release it through `chunks_directory_free()` — but a failure from `rwhashtab_init()` or `rwhashtab_insert()` returned without freeing it. `err2` is also reached from two points before `C->dir` exists, so this adds a label rather than a `free()` to `err2`; the existing labels shift by one.

With `nfiles == 0`, `C->dir` is now `NULL`. That is safe on every path that touches it: both `for (file = 0; file < nfiles; file++)` loops do not execute, and `chunks_directory_free()` ends in a plain `free(dir)`.

## What I did not exercise

I did not build or run tarsnap, and did not run `--fsck` against a server; there is no test in the tree covering `chunks_fsck_start()` with an empty chunk listing. Reachability of `nfiles == 0` is read from `callback_directory_response()` rather than observed on the wire, and the `malloc(0)` behaviour is the standard's allowance rather than something reproduced against a specific libc — on glibc `malloc(0)` returns non-NULL, so that defect is latent there and would bite where `malloc(0)` returns `NULL`. The leak is unconditional and platform-independent.

### DIFF
--- a/keygen/keygen_actual.c
+++ b/keygen/keygen_actual.c
@@ -53,21 +53,21 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
 
 	/* Sanity-check the user name. */
 	if (strlen(C->user) > 255) {
-		fprintf(stderr, "User name too long: %s\n", C->user);
+		warn0("User name too long: %s", C->user);
 		goto err0;
 	}
 	if (strlen(C->user) == 0) {
-		fprintf(stderr, "User name must be non-empty\n");
+		warn0("User name must be non-empty");
 		goto err0;
 	}
 
 	/* Sanity-check the machine name. */
 	if (strlen(C->name) > 255) {
-		fprintf(stderr, "Machine name too long: %s\n", C->name);
+		warn0("Machine name too long: %s", C->name);
 		goto err0;
 	}
 	if (strlen(C->name) == 0) {
-		fprintf(stderr, "Machine name must be non-empty\n");
+		warn0("Machine name must be non-empty");
 		goto err0;
 	}
 
@@ -80,7 +80,7 @@ keygen_actual(struct register_internal * C, const char * keyfilename,
 
 	/* Sanity-check the memory size. */
 	if (maxmem > SIZE_MAX) {
-		fprintf(stderr, "Passphrase memory size is too large\n");
+		warn0("Passphrase memory size is too large");
 		goto err0;
 	}
 
--- a/lib/netpacket/netpacket_op.c
+++ b/lib/netpacket/netpacket_op.c
@@ -520,6 +520,7 @@ int
 netpacket_close(NETPACKET_CONNECTION * NPC)
 {
 	struct netpacket_op * next;
+	int rc = 0;
 
 	/* Print statistics about network usage (if desired). */
 	if (tarsnap_opt_debug_network_stats)
@@ -527,8 +528,7 @@ netpacket_close(NETPACKET_CONNECTION * NPC)
 
 	/* Close the network protocol layer connection if we have one. */
 	if (NPC->NC != NULL)
-		if (netproto_close(NPC->NC))
-			goto err1;
+		rc = netproto_close(NPC->NC);
 
 	/* Free any queued operations. */
 	NPC->pending_current = NPC->pending_head;
@@ -544,20 +544,6 @@ netpacket_close(NETPACKET_CONNECTION * NPC)
 	/* Free the cookie. */
 	free(NPC);
 
-	/* Success! */
-	return (0);
-
-err1:
-	NPC->pending_current = NPC->pending_head;
-	while (NPC->pending_current != NULL) {
-		next = NPC->pending_current->next;
-		free(NPC->pending_current);
-		NPC->pending_current = next;
-	}
-
-	free(NPC->useragent);
-	free(NPC);
-
-	/* Failure! */
-	return (-1);
+	/* Report whether netproto_close() had any non-zero return. */
+	return ((rc == 0) ? 0 : -1);
 }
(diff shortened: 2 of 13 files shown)


woahwhattheheck:

tokenjunkielabs:
Following up on first-reporter bug bounty #841. This PR is the `woahwhattheheck` carrier for the empty-machine fsck failure and cleanup defects. Please review the report/fix when convenient and classify #841 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #846, opened 2026-09-07
storage_directory_read: hand the list to the caller after the close succeeds

### LINKED ISSUE #845: [Bug bounty] storage_directory_read() frees the file list after handing it to the caller
## Summary

`storage_directory_read()` in `tar/storage/storage_directory.c` publishes the file list to the caller's out-parameter and *then*, if `netpacket_close()` fails, frees that same pointer before returning failure. The caller is left holding a dangling pointer in a variable the function's own documentation describes as an output.

## Source

`tar/storage/storage_directory.c:73-91`, current master (`0bf0b299`):

```c
	/* Return results. */
	*flist = C.flist;
	*nfiles = C.nfiles;

	/* Close netpacket connection. */
	if (netpacket_close(C.NPC))
		goto err1;

	/* Success! */
	return (0);

err2:
	netpacket_close(C.NPC);
err1:
	free(C.flist);
err0:
	/* Failure! */
	return (-1);
```

Ownership of `C.flist` moves to the caller at the `*flist = C.flist;` assignment. Two lines later, a failed `netpacket_close()` jumps to `err1`, which frees it. So on that path the function returns `-1` having already written a pointer into `*flist` and then freed it.

`err2` is correct — it is only reachable before the assignment, so freeing there is right. The defect is specific to the one path that runs after the handoff.

The function's contract, from its own comment, is what makes this a defect rather than a matter of taste:

> Return the list and the number of files via ${flist} and ${nfiles} respectively.

## Impact

Latent, and I want to be exact about that: **no current caller triggers it.** I checked all five in the tree — `chunks_fsck_start()` in `tar/chunks/chunks_stats.c`, `phase1()`/`phase2()` in `tar/multitape/multitape_fsck.c`, `multitape_nuke.c`, `multitape_stats.c` (both call sites) and `recrypt/recrypt.c`. Every one of them either `goto`s an error label or `exit`s without reading or freeing `flist` on a non-zero return, so none double-frees today.

What makes it worth fixing is that freeing `*flist` after a failure is the natural reading of an out-parameter documented as returning the list, and the tree already has callers that free `flist` on their *success* paths a few lines later — `multitape_nuke.c` frees it between its two calls. A caller added later that tidies up on error would double-free.

## Suggested fix

Move the handoff after the close, so the failure path still owns what it frees:

```c
	/* Close netpacket connection. */
	if (netpacket_close(C.NPC))
		goto err1;

	/* Return results. */
	*flist = C.flist;
	*nfiles = C.nfiles;

	/* Success! */
	return (0);
```

`*flist` and `*nfiles` are then written only on success, `err1`'s `free()` becomes correct on every path that reaches it, and no other behaviour changes.

PR follows.

## What I did not exercise

I did not build or run tarsnap. I did not induce a `netpacket_close()` failure, so I have not observed this path executing; the reachability argument is read from the source above rather than measured. The caller survey is a read of the five call sites the tree contains today, not a guarantee about out-of-tree consumers.

woahwhattheheck:
Fixes #845.

`storage_directory_read()` assigned `*flist = C.flist;` and then, if `netpacket_close()` failed, jumped to `err1` and freed that pointer — returning `-1` after having already published it to the caller.

Moving the handoff below the close makes the out-parameters written only on success, so `err1`'s `free()` owns what it frees on every path that reaches it. `err2` was already correct; it is only reachable before the assignment.

No behaviour changes on the success path, and nothing else in the function moves.

## What I did not exercise

I did not build or run tarsnap, and did not induce a `netpacket_close()` failure, so I have not observed this path executing — the reasoning is from the source. As noted in the issue, none of the five in-tree callers frees or reads `flist` on a non-zero return, so this is a latent contract defect rather than a live double-free today.

### DIFF
--- a/tar/storage/storage_directory.c
+++ b/tar/storage/storage_directory.c
@@ -70,13 +70,13 @@ storage_directory_read(uint64_t machinenum, char class, int key,
 	if (network_spin(&C.done))
 		goto err2;
 
-	/* Return results. */
-	*flist = C.flist;
-	*nfiles = C.nfiles;
-
 	/* Close netpacket connection. */
 	if (netpacket_close(C.NPC))
 		goto err1;
+
+	/* Return results. */
+	*flist = C.flist;
+	*nfiles = C.nfiles;
 
 	/* Success! */
 	return (0);
--- a/tests/unit/storage-directory.c
+++ b/tests/unit/storage-directory.c
@@ -1,0 +1,197 @@
+/* Exercise the actual output-ownership boundary; transport is a fixture. */
+#include "platform.h"
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static void * allocated;
+static int freed;
+static void *
+observed_malloc(size_t len)
+{
+	void * p = malloc(len);
+
+	assert(allocated == NULL);
+	allocated = p;
+	return (p);
+}
+static void
+observed_free(void * p)
+{
+	if (p != NULL && p == allocated)
+		assert(freed++ == 0);
+	free(p);
+}
+
+#define malloc observed_malloc
+#define free observed_free
+#include "../../tar/storage/storage_directory.c"
+#undef free
+#undef malloc
+
+static int mode, closes, operations, spins;
+static size_t count;
+static int dummy;
+
+NETPACKET_CONNECTION *
+netpacket_open(const char * agent)
+{
+	assert(agent != NULL);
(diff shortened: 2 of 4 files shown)


woahwhattheheck:

tokenjunkielabs:
Following up on first-reporter bug bounty #845. This PR is the `woahwhattheheck` carrier for the dangling output-pointer ownership bug. Please review the report/fix when convenient and classify #845 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #848, opened 2026-09-07
storage_transaction_commit: record the modification once the server confirms

### LINKED ISSUE #847: [Bug bounty] A failed close after a confirmed commit loses the storage_modified flag, so tarsnap does not warn that server data changed
## Summary

`storage_transaction_commit()` sets `*storage_modified = 1` only after `netpacket_close()` succeeds, but the server has already confirmed the commit by the time that close is attempted. If the close fails, tarsnap exits reporting a plain error and never tells the user that the data on the server was changed — which is the exact situation the flag exists to report.

## The intent this violates

`tar/bsdtar.c:1331`, the only consumer of the flag, reads it *only* on a failed run:

```c
	if (bsdtar->return_value != 0) {
		bsdtar_warnc(bsdtar, 0,
		    "Error exit delayed from previous errors.");
		/* If we modified the storage, return 2 instead, and warn. */
		if (bsdtar->storage_modified) {
			bsdtar->return_value = 2;
			bsdtar_warnc(bsdtar, 0,
			    "Data on server was modified, but it might not"
			    " be exactly what you requested");
		}
	}
```

So the flag's whole purpose is to answer "we failed — but did we change the server anyway?" The header states the same contract for the function:

> If the data on the server has been modified, set ${*storage_modified} to 1.

## Source

`tar/storage/storage_transaction.c:679-712`, current master (`0bf0b299`):

```c
	/* Loop until we error out or we get a "success" response. */
	do {
		/* Send a request and get a response. */
		C.done = 0;
		if (netpacket_op(NPC, callback_commit_send, &C))
			goto err1;

		/* Wait until we have a response or an error. */
		if (network_spin(&C.done))
			goto err1;

		/* Did we succeed? */
		if (C.status == 0)
			break;
		...
	} while (1);

	/* Close netpacket connection. */
	if (netpacket_close(NPC))
		goto err0;

	/* We've modified the storage. */
	*storage_modified = 1;
```

The loop only breaks when `C.status == 0`, and `callback_commit_response()` sets that from `packetbuf[0]` after verifying the packet's HMAC — the server has committed. The commit is therefore already durable on the server when `netpacket_close()` runs, but a failure there jumps to `err0` and returns `-1` with `*storage_modified` untouched.

## Impact

A failed `netpacket_close()` after a confirmed commit produces:

- exit status 1 rather than 2, and
- "Error exit delayed from previous errors." **without** "Data on server was modified, but it might not be exactly what you requested".

The user, or a script keying on exit code 2, is told the operation failed and is not told the server changed. `--fsck`, `-d` and the recovery paths in `multitape_recover.c` all thread the same flag down to this function, so the loss applies to each of them.

## Suggested fix

Set the flag as soon as the server has confirmed, before the close:

```c
	/*
	 * The server has confirmed the commit, so the data on the server has
	 * been modified whether or not the rest of this function succeeds.
	 */
	*storage_modified = 1;

	/* Close netpacket connection. */
	if (netpacket_close(NPC))
		goto err0;
```

This only ever moves the flag from 0 to 1 in a case where the commit demonstrably happened, so it cannot cause a spurious "modified" report on a run where nothing was committed. The `err1` paths are unaffected: they are all reachable only before a status-0 response.

`storage_transaction_checkpoint()` in the same file already shows the safe shape for a close failure — its `goto err0` touches nothing that was already handed to the caller.

PR follows.

## What I did not exercise

I did not build or run tarsnap, and did not induce a `netpacket_close()` failure, so I have not observed the wrong exit status; the reasoning is read from the sources quoted above. I did not survey out-of-tree consumers of `storage_transaction_commit()`. I have not measured how often a close fails after a successful commit — the point of the report is that the flag is wrong when it does, not that it is common.

woahwhattheheck:
Fixes #847.

`storage_transaction_commit()` set `*storage_modified = 1` after `netpacket_close()`, but the commit loop only exits when the server has returned status 0 — an HMAC-verified confirmation that the commit happened. A close failure after that point returned `-1` with the flag still clear.

That flag is read in exactly one place, `tar/bsdtar.c:1331`, and only on a failed run: it upgrades the exit status to 2 and prints "Data on server was modified, but it might not be exactly what you requested". Losing it means a run that did change the server exits 1 and prints only "Error exit delayed from previous errors."

Setting the flag as soon as the server confirms fixes that. It can only move the flag from 0 to 1 in a case where the commit demonstrably occurred, so no run that committed nothing can now report "modified". The `err1` paths are unaffected — they are reachable only before a status-0 response.

## What I did not exercise

I did not build or run tarsnap, and did not induce a `netpacket_close()` failure, so I have not observed the wrong exit status; the reasoning is from the sources quoted in the issue. I did not survey out-of-tree consumers of this function.

### DIFF
--- a/tar/storage/storage_transaction.c
+++ b/tar/storage/storage_transaction.c
@@ -701,12 +701,15 @@ storage_transaction_commit(uint64_t machinenum, const uint8_t seqnum[32],
 		sleep(1);
 	} while (1);
 
+	/*
+	 * The server has confirmed the commit, so the data on the server has
+	 * been modified whether or not the rest of this function succeeds.
+	 */
+	*storage_modified = 1;
+
 	/* Close netpacket connection. */
 	if (netpacket_close(NPC))
 		goto err0;
-
-	/* We've modified the storage. */
-	*storage_modified = 1;
 
 	/* Success! */
 	return (0);
--- a/tests/unit/storage-commit.c
+++ b/tests/unit/storage-commit.c
@@ -1,0 +1,134 @@
+/* Actual commit/response code; a synthetic packet provider owns no account. */
+#include "platform.h"
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static unsigned int observed_sleep(unsigned int);
+#define sleep observed_sleep
+#include "../../tar/storage/storage_transaction.c"
+#undef sleep
+
+static int mode, closes, requests, spins, sleeps, errors;
+static int dummy;
+static void * operation_cookie;
+
+static unsigned int
+observed_sleep(unsigned int seconds)
+{
+	assert(seconds == 1 && ++sleeps <= 2);
+	return (0);
+}
+
+NETPACKET_CONNECTION *
+netpacket_open(const char * agent)
+{
+	assert(agent != NULL);
+	return (mode == 1 ? NULL : (NETPACKET_CONNECTION *)&dummy);
+}
+
+int
+netpacket_close(NETPACKET_CONNECTION * npc)
+{
+	assert(npc == (NETPACKET_CONNECTION *)&dummy);
+	closes++;
+	return (mode == 4 || mode == 8 ? -1 : 0);
+}
+
+int
+netpacket_op(NETPACKET_CONNECTION * npc, sendpacket_callback * send,
+    void * cookie)
+{
+	int rc;
+
+	assert(npc == (NETPACKET_CONNECTION *)&dummy);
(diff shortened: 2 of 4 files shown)


woahwhattheheck:

tokenjunkielabs:
Following up on first-reporter bug bounty #847. This PR is the `woahwhattheheck` carrier for the case where the server has already confirmed a storage commit but a later close failure leaves `storage_modified` clear, suppressing the client's warning/exit-status signal that server data changed. Please review the report/fix when convenient and classify #847 under the published Tarsnap bug-bounty schedule, including the applicable settlement route if it qualifies. I am not presuming a tier.

PR #852, opened 2026-09-07
chunks_directory_read: allocate with IMALLOC so an empty chunk directory loads

### LINKED ISSUE #851: [Bug bounty] chunks_directory_read() rejects a valid empty chunk directory, failing every operation that loads the cache
## Summary

`chunks_directory_read()` in `tar/chunks/chunks_directory.c` explicitly accepts an
on-disk chunk directory that contains zero chunk records, validating it as
well-formed, and then rejects it at the allocation step because
`malloc(0)` is permitted to return `NULL`.

Unlike the two previously reported sites of this class, this one is not confined to
`--fsck`: `chunks_directory_read()` is the common cache-directory loader for
`chunks_write_start()`, `chunks_delete_start()` and `chunks_stats()`, so on an
affected platform *creating an archive, deleting an archive and printing statistics
all fail* until the cache directory is removed and rebuilt.

## The zero-chunk directory is a valid, reachable state

`chunks_directory_write()` writes a 24-byte `struct chunkstats_external` header and
then, via `callback_write()`, **skips every record with `nrefs == 0`**:

```c
	/* If nrefs == 0, return without writing anything. */
	if (ch->nrefs == 0)
		return (0);
```

So a machine whose archives have all been deleted has a `directory` file of exactly
24 bytes. That is not corruption; it is the normal representation of "no chunks".

`chunks_directory_read()` agrees. Both of its size checks admit the 24-byte file:

```c
	/* Make sure the directory file isn't too small (different message). */
	if ((size_t)sb.st_size < sizeof(struct chunkstats_external)) {
		warn0("on-disk directory is too small (%jd bytes): %s", ...);
		goto err2;
	}

	/* Make sure the number of chunks is an integer. */
	if (((size_t)sb.st_size - sizeof(struct chunkstats_external)) %
	    (sizeof(struct chunkdata_external))) {
		warn0("on-disk directory is corrupt: %s", s);
		goto err2;
	}
```

24 is not less than 24, and `(24 - 24) % 48 == 0`, so `numchunks` is computed as 0
and the function proceeds. The read loop below is likewise written for it —
`for (; numchunks != 0; numchunks--)` simply does not execute.

## The defect

`tar/chunks/chunks_directory.c:212-220`:

```c
	if (statstape) {
		ps = malloc(numchunks * sizeof(struct chunkdata_statstape));
		*dir = ps;
	} else {
		p = malloc(numchunks * sizeof(struct chunkdata));
		*dir = p;
	}
	if (*dir == NULL)
		goto err2;
```

With `numchunks == 0` this is `malloc(0)`, which C99 7.20.3 permits to return a null
pointer. The `NULL` is then read as an allocation failure and the whole function
fails, returning `NULL` to its caller — for a cache directory that it has just
finished validating as correct.

`errno` is not set by a successful `malloc(0)`, so the caller's diagnostic is
whatever `errno` happened to hold.

## The tree already states the intended handling

`libcperciva/util/imalloc.h` makes the zero case part of the contract, returning
`NULL` for it deliberately and defining `IMALLOC` so that `NULL` is only an error
when a nonzero count was requested:

```c
/**
 * imalloc(nrec, reclen):
 */
```

```c
	    ((nrec) > 0))
```

and `multitape_metaindex_get()` writes the same guard out by hand three times in a
row, for exactly this reason:

```c
	if (((mind->hindex = malloc(mind->hindexlen)) == NULL) &&
	    (mind->hindexlen > 0))
		goto err1;
```

The overflow check immediately above the allocation site
(`numchunks >= SIZE_MAX / sizeof(struct chunkdata_statstape)`) is the other half of
what `imalloc()` does, which suggests the zero case was simply missed rather than
deliberately excluded.

## Scope of the class

I audited every variable-count allocation in the non-`libarchive`,
non-`libcperciva` sources for this pattern. The other candidates are all
constrained away from zero at or before the call:

| Site | Why zero cannot reach it |
| --- | --- |
| `netpacket_op.c:393` | every accepted packet type has a minimum length of 32 or more |
| `ccache_entry.c:296` | guarded by `cce->ccr->tlen > 0` |
| `crypto_keys_subr.c:314` | guarded by `buflen != 32` |
| `bsdtar.c:299` | `argc >= 1` |
| `multitape_metaindex.c:244,264,284` | already carry the `> 0` guard |
| `storage_read_cache.c:250` | failure is a silent "don't cache", not an error |
| `chunkify.c:225-227`, `rwhashtab.c:61,161` | counts are internal and never zero |

`chunks_directory_read()` is the only remaining site where a legitimately-zero count
derived from persisted state is passed to a bare `malloc` whose `NULL` is treated as
failure. With #841/#842 (`chunks_fsck_start()`, server-supplied count) and the
`multitape_fsck.c` `phase2()` site, this completes the class.

## Fix

PR to follow, replacing the two allocations with `IMALLOC` so the zero case is
handled the way the rest of the tree handles it, and so the overflow check is the
one in `imalloc()` rather than a hand-written duplicate.

## Severity

I am not claiming a crash or data loss. Nothing is read or written out of bounds:
the read loop does not execute, so the `NULL` is never dereferenced. The
consequence is that on an affected C implementation tarsnap refuses to operate on a
cache directory that is in a normal state, and the user's route out is to delete the
cache directory and re-run `--fsck`. I have not reproduced it on a platform where
`malloc(0)` returns `NULL`; the claim rests on the C standard permitting it and on
this codebase already defending against it everywhere else.

---

woahwhattheheck:
Fixes #851.

`chunks_directory_write()` omits every record with `nrefs == 0`, so a machine whose
archives have all been deleted leaves a `directory` file of exactly 24 bytes — the
`struct chunkstats_external` header and nothing else. `chunks_directory_read()`
accepts that file: it is not smaller than the header, and `(24 - 24) % 48 == 0`, so
`numchunks` comes out as 0 and the read loop correctly does not execute.

The allocation is the one step that does not tolerate it. `malloc(0)` may return
`NULL`, and the check below reads that as failure, so `chunks_directory_read()`
returns `NULL` for a cache directory it has just validated. Its callers are
`chunks_write_start()`, `chunks_delete_start()` and `chunks_stats()`, so creating an
archive, deleting an archive and printing statistics all fail on an affected
platform.

This switches both allocations to `IMALLOC`, which returns `NULL` for a zero count
and reports failure only when a nonzero count was requested — the same handling
`phase1()` in `tar/multitape/multitape_fsck.c` uses for its per-file allocation, and
the same contract `multitape_metaindex_get()` writes out by hand as
`(... == NULL) && (len > 0)`.

Two notes on what the diff deliberately does not do:

- The existing `numchunks >= SIZE_MAX / sizeof(struct chunkdata_statstape)` check is
  left in place. `imalloc()` performs the same overflow check, but its failure would
  lose the specific "on-disk directory is too large" diagnostic, so the check stays.
- `*dir` is now assigned only after a successful allocation. Nothing on the error
  ladder reads it, and `err3`'s `free(*dir)` is only reachable once it has been set.

`*dir` being `NULL` on success is already part of this function's contract: the
`cachepath == NULL` path above returns a valid empty hash table with `*dir = NULL`,
and `chunks_directory_free()` passes it to `free()`.

The change is two `IMALLOC` substitutions plus the `imalloc.h` include, which
resolves here the same way the existing `fileutil.h` / `sysendian.h` /
`asprintf.h` includes from `libcperciva/util/` do.

---


tokenjunkielabs:
Following up on first-reporter bug bounty #851. This PR is the `woahwhattheheck` carrier for the valid-empty chunk-directory loader failure. Please review the report/fix when convenient and classify #851 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #854, opened 2026-09-07
recrypt: sync the new chunk directory before deleting old blocks

### LINKED ISSUE #853: [Bug bounty] tarsnap-recrypt commits the old-machine deletes without syncing the new machine's chunk directory
## Summary

`tarsnap-recrypt` writes the new machine's chunk directory with plain stdio and
never syncs it, then immediately performs the irreversible half of its job:
deleting every block of the old machine from the server and committing that delete.

`chunks_directory_write()` — the routine the rest of the tree uses to write this
same file — syncs the file *and* the containing directory. `recrypt.c` writes it by
hand and does neither.

This is a durability gap on the *new* machine's directory. It is not #746 (closed by
#747, which added `dirutil_fsyncdir()` to `chunks_directory_write()` itself) and it
is not #757 (open, and concerned with the *old* machine's directory still claiming
deleted chunks exist after the delete commit). Different file, different machine,
and, as set out below, the opposite and less dangerous direction of error.

## The write

`recrypt/recrypt.c:591` creates the new directory file, deliberately empty:

```c
	/*
	 * Create an empty directory file for the new machine.  An empty file
	 * is not a valid directory, so if we crash, tarsnap will fail until
	 * the cache directory is reconstructed via --fsck or we are re-run
	 * and allowed to finish the recrypting process.
	 */
	if ((ndir = fopen(ndirpath, "w")) == NULL) {
```

Blocks are then copied and committed to the new machine. Afterwards, at `:637-651`,
the old machine's directory is copied into it:

```c
	if ((odir = fopen(odirpath, "r")) == NULL) { ... }

	/* ... and copy its contents into the new cache directory. */
	copydirectory(odir, ndir);

	/* Close the old and new chunk directories. */
	if (fclose(ndir) || fclose(odir)) { ... }
```

`copydirectory()` is `fread`/`fwrite` only, and `fclose()` flushes to the kernel
without syncing. `recrypt.c` does not include `fileutil.h` and calls neither
`fileutil_fsync()` nor `dirutil_fsyncdir()` anywhere. (It includes `dirutil.h` for
`build_dir()`.)

## The irreversible step that follows

`:654-696`, with nothing between it and the `fclose()` above:

```c
	/* Delete blocks from old machine. */
	for (bpos = 0; bpos < oblistlen; bpos++) {
		if (storage_delete_file(SD, oblist[bpos].class,
		    oblist[bpos].name)) { ... }
		...
	}
	if (storage_delete_end(SD)) { ... }

	/* Commit the delete transaction and delete the old chunk dir. */
	if (multitape_commit(ocachedir, omachinenum, oseqnum, 1,
	    &storage_modified)) { ... }
	if (unlink(odirpath)) { ... }
```

So the server-side deletion of the source data is committed while the only
description of the destination's chunks may still be sitting in the page cache.

## Compare `chunks_directory_write()`

`tar/chunks/chunks_directory.c:353-363` does what this path does not:

```c
	/* Call fsync on the new chunk directory and close it. */
	if (fileutil_fsync(f, s))
		goto err2;
	if (fclose(f)) { ... }

	/* Make sure that the new directory entry is durable. */
	if (dirutil_fsyncdir(cachepath))
		goto err1;
```

Both calls are there on purpose; the second is the fix from #747.

## What a crash actually costs

Losing power after the delete commit reaches the server but before the new
directory's contents reach disk leaves the new machine's `directory` in one of three
states:

- **empty** — rejected by `chunks_directory_read()` as "too small". This is the case
  the comment at `:585` anticipates, and the stated remedy is `--fsck`.
- **truncated mid-record** — rejected as "on-disk directory is corrupt", same remedy.
- **truncated to a whole number of records** — *accepted*, describing fewer chunks
  than the new machine actually holds.

I want to be accurate about the third case rather than overstate it. It is the safe
direction of error:

- The blocks themselves were copied and committed to the new machine at `:603-627`,
  before any delete was issued, so nothing is gone from the server.
- Under-reporting causes a later backup to re-upload chunks it already has. That is
  wasted bandwidth and storage, not loss.
- A later `tarsnap -d` does not do anything unsafe with the missing records:
  `chunks_delete_chunk()` errors out with "Chunk is missing or directory is corrupt"
  rather than mis-deleting.

**I am not claiming the data-loss tier.** The consequence is a cache directory that
is silently wrong until `--fsck` rebuilds it, reached through a window that the
codebase closes everywhere else it writes this file.

## Fix

Sync the file before closing it and sync the cache directory afterwards, both before
the delete loop begins:

```c
	if (fileutil_fsync(ndir, ndirpath)) { ... }
	if (fclose(ndir) || fclose(odir)) { ... }
	if (dirutil_fsyncdir(ncachedir)) { ... }
```

plus `#include "fileutil.h"`. PR to follow.

While there: `fclose(ndir) || fclose(odir)` short-circuits, so `odir` is left open if
the first `fclose` fails. The process exits immediately afterwards so nothing is
actually lost, but the PR separates them so both are closed and both are reported.

---

woahwhattheheck:
Fixes #853.

`tarsnap-recrypt` copies the old machine's chunk directory into the new machine's
cache with `fread`/`fwrite` and closes it with `fclose()`, which flushes to the
kernel but does not sync. It then deletes every block of the old machine from the
server and commits that delete. The source data is removed irreversibly while the
only description of where it now lives may still be in the page cache.

`chunks_directory_write()` writes this same file correctly:

```c
	if (fileutil_fsync(f, s))
		goto err2;
	if (fclose(f)) { ... }
	if (dirutil_fsyncdir(cachepath))
		goto err1;
```

The `dirutil_fsyncdir()` there is the fix from #747. This patch applies the same
pair to the hand-written copy in `recrypt.c`, before the delete loop begins.

Both helpers already report their own failures via `warnp()`, so the new call sites
just `exit(1)`, matching how `chunks_directory_write()` treats them.

The combined `fclose(ndir) || fclose(odir)` is also split: short-circuit evaluation
meant a failure on the first left the second file open, and neither diagnostic said
which file it was about. The process exits immediately either way, so this is
tidiness rather than a fix.

On the build: `fileutil.h` resolves here — `tarsnap_recrypt_CPPFLAGS` already
carries `-I$(top_srcdir)/lib-platform/util` — and `fileutil_fsync` is already linked
into `tarsnap-recrypt`, since `tar/chunks/chunks_directory.c` is in
`tarsnap_recrypt_SOURCES` and calls it. `dirutil_fsyncdir` needs no new include;
`dirutil.h` is already there for `build_dir()`.

Not compile-tested locally — I work from the GitHub sources without a build tree, so
the include and link reasoning above is from `Makefile.am` rather than from a build.

As noted in the issue, I am not claiming this as the data-loss tier: the blocks are
committed to the new machine before any delete is issued, so a crash in this window
costs a cache directory that `--fsck` rebuilds, not the data.

---


tokenjunkielabs:
Following up on first-reporter bug bounty #853. This PR is the `woahwhattheheck` carrier for the reported recrypt durability gap. Please review the report/fix when convenient and classify #853 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #856, opened 2026-09-07
tsnetwork_writeq: make network_writeq_cancel terminate

### LINKED ISSUE #855: [Bug bounty] network_writeq_cancel() spins forever when a queued write was never registered
## Summary

`network_writeq_cancel()` loops until the write queue is empty, but the only thing
inside the loop that can empty it is a callback that `network_deregister()` invokes
*only if one is registered*. When nothing is registered, `network_deregister()`
returns 0 having done nothing, the queue does not advance, and the loop spins
forever at 100% CPU.

There is a reachable path that leaves the queue non-empty with nothing registered,
and it runs during connection teardown — so the client wedges instead of exiting.

## The loop

`lib/network/tsnetwork_writeq.c:216-229`:

```c
int
network_writeq_cancel(NETWORK_WRITEQ * Q)
{
	int rc = 0, rc2;

	/* Keep on deregistering callbacks until the queue is empty. */
	while (Q->head != NULL) {
		rc2 = network_deregister(Q->fd, NETWORK_OP_WRITE);
		rc = rc ? rc : rc2;
	}

	/* Return first non-zero result from deregistration. */
	return (rc);
}
```

The loop makes progress only via `callback_bufdone()`, which
`network_deregister()` reaches through `docallback()`. But
`lib/network/tsnetwork_register.c` returns success without calling anything in
three cases:

```c
	/* If we have no callbacks array we have no callback, so do nothing. */
	if (cs == NULL)
		return (0);

	/* If this is beyond the end of the array, we have no callback. */
	if (fd >= (int)callbacks_getsize(cs))
		return (0);
	...
	/* If there is no callback, return silently. */
	if (c == NULL || c->callback == NULL)
		return (0);
```

`rc2` is 0 in each, so `rc` stays 0 and the loop condition is unchanged. Nothing
breaks it.

The loop is therefore correct only under the invariant *`Q->head != NULL` implies a
`NETWORK_OP_WRITE` callback is registered on `Q->fd`*.

## Where the invariant breaks

`callback_bufdone()`, `lib/network/tsnetwork_writeq.c:91-113`:

```c
	if (Q->head != NULL) {
		if (dowrite(Q))
			goto err1;
	}
	...
err1:
	(head_old->callback)(head_old->cookie, status);
	free(head_old);

	/* Failure! */
	return (-1);
```

On this path the completed buffer has already been unlinked and `Q->head` points at
the *next* buffer, which `dowrite()` just failed to start. `err1` reports the old
buffer's completion and returns −1 — and leaves `Q->head` non-NULL with nothing
registered, because `dowrite()` failed before registering anything.

`dowrite()` fails when `tvmath_subctime()` fails, or when `tsnetwork_write()` does.
The latter ends at `network_register()`, which fails on allocation failure
(`callbacks_init()`, `callbacks_resize()`, or the per-fd record) and returns −1
having registered nothing.

Note the contrast with `network_writeq_add_internal()`, which handles the same
`dowrite()` failure correctly: its `err2` label rolls the queue back so `Q->head`
returns to NULL. Only `callback_bufdone()` leaves it inconsistent.

## Why teardown then hangs

`callback_bufdone()` returning −1 propagates out of the network layer and the caller
tears the connection down. Both teardown paths in `lib/netproto/netproto.c` call the
cancel first, exactly as `network_writeq_free()`'s comment instructs:

```c
	/* Cancel pending writes. */
	if (C->Q != NULL) {
		rc2 = network_writeq_cancel(C->Q);      /* netproto_close, :242 */
		rc = rc ? rc : rc2;
	}
```

and `netproto_flush()` at `:203`. Either call is the hang.

## Impact

The process spins at 100% CPU and never returns, in the teardown that follows a
failed write. For a backup run from cron this is worse in practice than an abort:
nothing completes, nothing is reported, and the job has to be noticed and killed by
hand.

I am not claiming a tier for this. It is a hang rather than a crash, and reaching it
needs an allocation failure at a specific moment, so it is memory-pressure
conditional rather than something a user can trigger at will. No data is at risk —
the queued buffers are owned by the caller and the connection is already being
discarded.

## Fix

Detect that `network_deregister()` did not advance the queue, and in that case
dequeue the buffer directly so its callback still runs and the loop terminates. PR
to follow.

## Aside

The comment above `network_writeq_cancel()` says the callbacks are called with
`NETWORK_STATUS_DEQUEUE`. There is no such constant; `docallback()` passes
`NETWORK_STATUS_CANCEL`, which is what `network_deregister()`'s own documentation
says. The PR uses `NETWORK_STATUS_CANCEL` to match actual behaviour and leaves the
stale comment alone, since correcting it is a separate matter.

---

woahwhattheheck:
Fixes #855.

`network_writeq_cancel()` empties the queue only through the callback that
`network_deregister()` invokes. But `network_deregister()` returns 0 without calling
anything when there is no callback registered for the descriptor — no callbacks
array, `fd` past the end of it, or an empty record. In that case the queue does not
advance and `rc` stays 0, so nothing ends the loop.

`callback_bufdone()` can leave exactly that state. On its `err1` path the completed
buffer has already been unlinked, `Q->head` points at the next buffer, and
`dowrite()` has just failed — before registering anything, since it fails either in
`tvmath_subctime()` or inside `tsnetwork_write()` at `network_register()`. The −1
propagates out and the caller tears the connection down, and both teardown paths in
`lib/netproto/netproto.c` (`netproto_close():242`, `netproto_flush():203`) call
`network_writeq_cancel()` first, as `network_writeq_free()`'s comment instructs.

The patch remembers the head, and if `network_deregister()` did not advance it,
dequeues the buffer directly — unlinking it, fixing `tailptr` the same way
`callback_bufdone()` does, calling the upstream callback, and freeing the record.
The loop then terminates and every queued write still gets its callback.

`NETWORK_STATUS_CANCEL` is what this path already delivers: `network_deregister()`
reaches `docallback(c, -1)`, and `-1` maps to `NETWORK_STATUS_CANCEL`. The comment
above the function says `NETWORK_STATUS_DEQUEUE`, but no such constant exists in
`tsnetwork.h`; I have left that comment alone rather than mix a docs correction into
a behaviour fix.

`callback_bufdone()`'s `CLOSED` → `ZEROBYTE` conversion is deliberately not
replicated, since it applies only to `NETWORK_STATUS_CLOSED`.

An alternative fix would be to make `callback_bufdone()`'s `err1` path roll the queue
back, the way `network_writeq_add_internal()`'s `err2` already does for the same
`dowrite()` failure. I did not take that route because it would drop the remaining
buffers without running their callbacks, and because `network_writeq_cancel()` would
still be non-terminating for any other caller that reached the same state. Happy to
switch if you prefer it there.

Not compile-tested locally — I work from the GitHub sources without a build tree. The
patch uses only `struct network_writeq_buf`, `NETWORK_STATUS_CANCEL` and `free()`,
all already visible in this file.

---


tokenjunkielabs:
Following up on first-reporter bug bounty #855. This PR is the woahwhattheheck carrier for the reported teardown hang. Please review the report/fix when convenient and classify #855 under the repository's bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #858, opened 2026-09-07
tsnetwork_buf: report NODATA when the buffer is empty

### LINKED ISSUE #857: callback_buf() sets NETWORK_STATUS_NODATA when the buffer is non-empty, inverting its documented meaning
## Summary

`callback_buf()` in `lib/network/tsnetwork_buf.c` converts a timeout to
`NETWORK_STATUS_NODATA` when the buffer **has** data, but the constant means the
opposite and the line's own comment says the opposite.

**This has no observable effect in the current tree**, and I want that stated up
front rather than buried: the sole consumer treats `NODATA` and `TIMEOUT`
identically. I am reporting it as a latent inversion with a one-character fix, not
claiming user-visible impact.

## The inversion

`lib/network/tsnetwork_buf.c:74-79`:

```c
	if (status != NETWORK_STATUS_OK) {
		/* If we have no data, mark a timeout as "no data" instead. */
		if ((C->bufpos != 0) && (status == NETWORK_STATUS_TIMEOUT))
			status = NETWORK_STATUS_NODATA;
		goto docallback;
	}
```

`C->bufpos` is how much of the buffer has been filled so far, so `C->bufpos != 0`
is "we **have** data". Three things disagree with the code:

1. The comment on the line directly above says *"If we have no data"*.
2. `lib/network/tsnetwork.h:22` defines the constant as an **empty** buffer:

   ```c
   #define NETWORK_STATUS_NODATA	2	/* Timeout with empty buffer. */
   ```

3. The distinction only makes sense that way round. `tsnetwork_read()`'s contract is
   to "time out if no data can be read for a period of time `to0`, or if the complete
   buffer has not been read after time `to1`" — the case worth separating from a
   plain timeout is the one where nothing arrived at all.

So the two statuses are currently swapped: a timeout on a partially-filled buffer is
reported as `NODATA`, and a timeout with nothing received is reported as `TIMEOUT`.

## Why nothing breaks today

`NETWORK_STATUS_NODATA` is read in exactly one place,
`lib/netproto/netproto.c:33-37`:

```c
	case NETWORK_STATUS_NODATA:
	case NETWORK_STATUS_TIMEOUT:
		/* Server timed out. */
		warn0("Timeout communicating with server");
		break;
```

Both arms fall together, so the swap is invisible. Nothing else in the tree
distinguishes them.

That is also why I think it is still worth fixing: the value is computed correctly
nowhere and consumed meaningfully nowhere, so the next caller that wants "did we get
anything at all?" will read the header, use `NODATA`, and get precisely the wrong
answer. The comment above the line will confirm their incorrect reading.

## Fix

`!=` becomes `==`. PR to follow.

```c
		if ((C->bufpos == 0) && (status == NETWORK_STATUS_TIMEOUT))
			status = NETWORK_STATUS_NODATA;
```

If you would rather keep the current behaviour and correct the comment and the
header instead, that is a one-line change too and I am happy to send it that way —
but the header's wording and the `to0`/`to1` contract both point at the code being
the side that is wrong.

---

woahwhattheheck:
Fixes #857.

`NETWORK_STATUS_NODATA` is defined in `lib/network/tsnetwork.h:22` as

```c
#define NETWORK_STATUS_NODATA	2	/* Timeout with empty buffer. */
```

and the comment on the line in question says the conversion is for when we have no
data — but the test is `C->bufpos != 0`, which is the case where we do. The two
statuses are swapped: a timeout on a partially-filled buffer is currently reported as
`NODATA`, and a timeout with nothing received is reported as `TIMEOUT`.

One character. No behaviour changes today, and I would rather say so plainly:
`netproto_printerr_internal()` is the only consumer and handles `NODATA` and
`TIMEOUT` in the same switch arm, so the swap is invisible. The reason to fix it is
that the next caller who wants "did anything arrive at all?" will read the header,
reach for `NODATA`, and get the opposite — with the comment above the line confirming
their incorrect reading.

I took the code as the side that is wrong rather than the header, because the header's
wording and `tsnetwork_read()`'s documented `to0`/`to1` contract agree with each
other: the case worth separating from a plain timeout is the one where nothing was
received. If you would rather keep the behaviour and correct the comment and header
instead, say so and I will send that version.

---

### DIFF
--- a/lib/network/tsnetwork_buf.c
+++ b/lib/network/tsnetwork_buf.c
@@ -73,7 +73,7 @@ callback_buf(void * cookie, int status)
 
 	if (status != NETWORK_STATUS_OK) {
 		/* If we have no data, mark a timeout as "no data" instead. */
-		if ((C->bufpos != 0) && (status == NETWORK_STATUS_TIMEOUT))
+		if ((C->bufpos == 0) && (status == NETWORK_STATUS_TIMEOUT))
 			status = NETWORK_STATUS_NODATA;
 		goto docallback;
 	}
--- a/tests/unit/network-timeout.c
+++ b/tests/unit/network-timeout.c
@@ -1,0 +1,181 @@
+/* Real buffer I/O over local socketpairs; the event scheduler is controlled. */
+#include <sys/socket.h>
+#include <sys/time.h>
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "tarsnap_opt.h"
+#include "tsnetwork.h"
+#include "tsnetwork_internal.h"
+#include "tvmath.h"
+#include "warnp.h"
+
+int tarsnap_opt_noisy_warnings;
+static network_callback * pending;
+static void * pending_cookie;
+static size_t quota, consumed;
+static int calls, observed;
+
+int
+tvmath_addctime(struct timeval * tv)
+{
+	(void)tv;
+	return (0);
+}
+
+int
+tvmath_subctime(struct timeval * tv)
+{
+	(void)tv;
+	return (0);
+}
+
+int
+network_bwlimit_get(int op, size_t * len)
+{
+	assert(op == NETWORK_OP_READ || op == NETWORK_OP_WRITE);
+	*len = quota;
+	return (0);
+}
+
+int
+network_bwlimit_eat(int op, size_t len)
+{
+	assert(op == NETWORK_OP_READ || op == NETWORK_OP_WRITE);
+	consumed += len;
+	return (0);
+}
+
+int
+network_register(int fd, int op, struct timeval * timeo,
+    network_callback * callback, void * cookie)
+{
+	assert(fd >= 0 && timeo != NULL);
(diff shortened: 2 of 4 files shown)


woahwhattheheck:
Regression coverage is included at current head `606758a62e6588e5679a05b348c82fb04dab4a80`: [test source](https://github.com/woahwhattheheck/tarsnap/blob/606758a62e6588e5679a05b348c82fb04dab4a80/tests/unit/network-timeout.c) and [run instructions/results](https://github.com/woahwhattheheck/tarsnap/blob/606758a62e6588e5679a05b348c82fb04dab4a80/tests/unit/network-timeout.md).

Run `sh tests/unit/network-timeout.sh` on Linux/GCC. The recorded focused result is PASS across 34 local socketpair scenarios: read/write, zero/partial progress, timeout/cancel/error/close statuses, callback return propagation, and complete transfers. The common base fails the callback-status assertion.

This is a standalone regression, not part of `make test`; the scheduler and bandwidth provider are controlled fixtures. It does not establish full-suite or hosted-service validation. Scope remains the callback-status contract: current consumers share the timeout error branch, so no additional user-visible production impact is claimed. No new bounty claim.

PR #860, opened 2026-09-07
sigquit: declare sigquit_received volatile

### LINKED ISSUE #859: sigquit_received is written from a signal handler but is not volatile
## Summary

`sigquit_received` is written from a signal handler but is not declared `volatile`.
The two other signal-received flags in the tree both are, and the one that is not is
the one driving `^Q` and `--maxbw` archive truncation.

## The inconsistency

`lib/util/sigquit.c:14` and `lib/util/sigquit.h:7`:

```c
sig_atomic_t sigquit_received;
```
```c
extern sig_atomic_t sigquit_received;
```

written from the handler at `lib/util/sigquit.c:28-35`:

```c
static void
sigquit_handler(int sig)
{

	(void)sig;	/* UNUSED */

	sigquit_received = 1;
}
```

The other two flags of exactly this kind are declared correctly:

| Flag | Declaration |
| --- | --- |
| `tar/siginfo.c:42` | `static volatile sig_atomic_t siginfo_received = 0;` |
| `tar/write.c:210` | `static volatile sig_atomic_t sigusr2_received = 0;` |
| `lib/util/sigquit.c:14` | `sig_atomic_t sigquit_received;` |

`tar/write.c` is the file that *reads* `sigquit_received`, and it declares its own
`sigusr2_received` with `volatile` twenty lines further down.

## Why it matters

C11 §7.14.1.1p5 permits a handler to access, of static-storage objects, only those
declared `volatile sig_atomic_t` (or lock-free atomics). Without `volatile` the
compiler is entitled to cache the read, so a poll of the flag need never observe the
handler's write.

The reader is `truncate_archive()` in `tar/write.c:171-191`, called repeatedly as the
archive is written:

```c
	if (sigquit_received == 0)
		return (0);
```

Today the read survives because `truncate_archive()` is a separate function loading
an external global, so the load is reissued on each call. That is a property of how
it happens to be compiled, not something the standard guarantees — under LTO, or if
the function is inlined into the write loop, the load can be hoisted out and the flag
never re-read.

The affected feature is not obscure. The message on the following line is

```c
		bsdtar_warnc(bsdtar, 0, "quit signal received or bandwidth "
		    "limit reached; archive is being truncated");
```

so this flag is how both `^Q`/`SIGQUIT` **and** the `--maxbw` bandwidth-limit stop
truncate an archive cleanly. If the read is hoisted, both silently stop working and
the archive keeps being written.

## Honest scope

I have not observed this misbehaving on any specific compiler, and I would not expect
it to at ordinary optimization levels for the reason given above. I am reporting it
as undefined behaviour with an in-tree precedent on both sides of the question, not
as a reproduced failure.

## Fix

Add `volatile` in the definition and the declaration, matching `siginfo_received` and
`sigusr2_received`. PR to follow.

---

woahwhattheheck:
Fixes #859.

`sigquit_received` is written from `sigquit_handler()`, and C11 §7.14.1.1p5 allows a
handler to access, of static-storage objects, only those declared
`volatile sig_atomic_t` (or lock-free atomics). Without `volatile` the compiler may
cache the read and never observe the handler's write.

The two other flags of this kind in the tree already do it correctly:

- `tar/siginfo.c:42` — `static volatile sig_atomic_t siginfo_received = 0;`
- `tar/write.c:210` — `static volatile sig_atomic_t sigusr2_received = 0;`

`tar/write.c` is the file that reads `sigquit_received`, and declares its own flag
with `volatile` twenty lines below that read.

The reader is `truncate_archive()`, which is how both `^Q`/`SIGQUIT` and the
`--maxbw` bandwidth limit stop an archive cleanly — the warning on the next line
covers both cases. Today the load is reissued because `truncate_archive()` is a
separate function reading an external global, but that is how it happens to compile
rather than something guaranteed; under LTO or if it is inlined into the write loop
the load can be hoisted and the flag never re-read.

`volatile` is added in both the definition and the `extern` declaration, so they
continue to match — a mismatch between the two would be a constraint violation.

I have not observed this misbehaving on a specific compiler and would not expect it
to at ordinary optimization levels; this is undefined behaviour with a precedent on
both sides of the question in-tree, not a reproduced failure.

---


gperciva:
@woahwhattheheck the changes to `lib/util/siguit.*` look good.  The commit message should:
* reference C99, not C11.
* don't mention the "two other flags"
* don't mention "the reader is truncate_archive()"
* end with:

```
Reported by:\twoahwhattheheck
Bug bounty:\t$10 ("harmless")
```
(with the `\t` replaced by tabs, of course.)

Also, remove those tests.

gperciva:
@woahwhattheheck thanks for the update.  Could you rebase those 3 commits together, only keeping the final commit message?

PR #862, opened 2026-09-07
patricia: avoid left-shifting a negative value

### LINKED ISSUE #861: patricia.c: left shift of a negative value in compare() and patricia_insert()
## Summary

`lib/datastruct/patricia.c` computes a bit mask as `(- n->mask) << 1` in two places.
`mask` is a `uint8_t`, so it promotes to `int` and the negation is a negative `int`;
left-shifting a negative value is undefined behaviour (C99/C11 §6.5.7p4).

Unlike most UB of this kind, this one is not invisible: it is exactly what
UndefinedBehaviorSanitizer's `shift-base` check reports as *"left shift of negative
value"*, and both sites sit on the patricia lookup/insert path, which the chunk cache
exercises constantly. A UBSan build of tarsnap trips on it continuously.

## The two sites

`lib/datastruct/patricia.c:184`, in `compare()`:

```c
	/* Compute how the top bits differ. */
	mm = (n->high ^ key[i]) & ((- n->mask) << 1);
```

`lib/datastruct/patricia.c:357`, in `patricia_insert()`:

```c
				pnew->high = key[mlen] & ((- mask) << 1);
```

`mask` is `uint8_t` in both cases — the struct field at `patricia.c:34`:

```c
struct pnode {
	struct pnode *	left;		/* Left child. */
	struct pnode *	right;		/* Right child. */
	uint8_t		mask;		/* Critical bit mask. */
	uint8_t		high;		/* High bits of this node. */
	uint8_t		slen;		/* Length of s[]. */
	uint8_t		s[];		/* Bytes since parent's s[]. */
};
```

and the local in `patricia_insert()` is declared `uint8_t mask;` at `:296`.

Integer promotion makes `- n->mask` an `int`, negative whenever `mask` is non-zero —
and `mask == 0` is handled separately at `:171` and `:353`, so the shifted value is
negative every time these lines are reached.

## Fix

Make the negation unsigned. Negation of an unsigned type is defined as modular
arithmetic, and a left shift of an unsigned value is defined with the shifted-out
bits discarded, so the same bit pattern results with no UB:

```c
	mm = (n->high ^ key[i]) & ((-(unsigned int)n->mask) << 1);
```

I checked the result for every value `mask` can take. It is a single bit, so there
are eight:

| `mask` | current | proposed |
| --- | --- | --- |
| 0x01 | 0xfe | 0xfe |
| 0x02 | 0xfc | 0xfc |
| 0x04 | 0xf8 | 0xf8 |
| 0x08 | 0xf0 | 0xf0 |
| 0x10 | 0xe0 | 0xe0 |
| 0x20 | 0xc0 | 0xc0 |
| 0x40 | 0x80 | 0x80 |
| 0x80 | 0x00 | 0x00 |

Identical throughout, so this is a pure UB fix with no behaviour change. PR to follow.

## Scope

No miscompilation is claimed. Every compiler I am aware of produces the intended
two's-complement bit pattern here, and the truncation to `uint8_t` discards the
sign-extended high bits either way. The value is in making a sanitizer build usable
and in not relying on behaviour the standard does not define.

This is the same class as #835/#836 (`padme()` shifting an `int`), which I reported
earlier; different file, different mechanism — that one is a shift count, this is a
negative shift base.

---

woahwhattheheck:
Fixes #861.

`mask` is a `uint8_t` in both places, so it promotes to `int` and `(- mask)` is a
negative `int` whenever `mask` is non-zero — which is every time these lines run, as
the `mask == 0` case is handled separately just above each of them (`:171` and
`:353`). Left-shifting a negative value is undefined per C99 §6.5.7p4, and it is
precisely what UBSan's `shift-base` check reports as "left shift of negative value".
Both sites are on the patricia lookup/insert path that the chunk cache exercises
constantly, so a sanitizer build trips on them continuously.

Negating as `unsigned int` is defined as modular arithmetic, and a left shift of an
unsigned value simply discards the bits shifted out, so the same bit pattern results.

`mask` is always a single bit, so there are eight possible values and I checked all of
them:

| `mask` | before | after |
| --- | --- | --- |
| 0x01 | 0xfe | 0xfe |
| 0x02 | 0xfc | 0xfc |
| 0x04 | 0xf8 | 0xf8 |
| 0x08 | 0xf0 | 0xf0 |
| 0x10 | 0xe0 | 0xe0 |
| 0x20 | 0xc0 | 0xc0 |
| 0x40 | 0x80 | 0x80 |
| 0x80 | 0x00 | 0x00 |

Identical in every case, so this is a pure UB fix and no behaviour changes.

No miscompilation is claimed — the truncation to `uint8_t` discards the sign-extended
high bits either way, and every compiler I know of emits the intended two's-complement
pattern. The value is in making a sanitizer build usable and not depending on
behaviour the standard leaves undefined.

---

### DIFF
--- a/lib/datastruct/patricia.c
+++ b/lib/datastruct/patricia.c
@@ -181,7 +181,7 @@ compare(const struct pnode * n, const uint8_t * key, size_t keylen,
 	}
 
 	/* Compute how the top bits differ. */
-	mm = (n->high ^ key[i]) & ((- n->mask) << 1);
+	mm = (n->high ^ key[i]) & ((-(unsigned int)n->mask) << 1);
 
 	/* If the top bits match, the node is a prefix of the key. */
 	if (mm == 0)
@@ -354,7 +354,7 @@ patricia_insert(PATRICIA * P, const uint8_t * key, size_t keylen, void * rec)
 				pnew->high = 0;
 				pnew->right = pnew2;
 			} else {
-				pnew->high = key[mlen] & ((- mask) << 1);
+				pnew->high = key[mlen] & ((-(unsigned int)mask) << 1);
 
 				/*
 				 * This looks wrong, but it actually works:
--- a/tests/unit/patricia.c
+++ b/tests/unit/patricia.c
@@ -1,0 +1,118 @@
+/* Exercise the real tree with binary keys and every critical-bit position. */
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "patricia.h"
+
+struct record {
+	uint8_t key[514];
+	size_t len;
+};
+
+struct traversal {
+	struct record ** ordered;
+	size_t count;
+	size_t visited;
+	size_t stop;
+};
+
+static int
+keycmp(const void * a, const void * b)
+{
+	const struct record * x = *(struct record * const *)a;
+	const struct record * y = *(struct record * const *)b;
+	size_t len = x->len < y->len ? x->len : y->len;
+	int rc;
+
+	if ((rc = memcmp(x->key, y->key, len)) != 0)
+		return (rc);
+	return ((x->len > y->len) - (x->len < y->len));
+}
+
+static int
+visit(void * cookie, uint8_t * key, size_t len, void * rec)
+{
+	struct traversal * t = cookie;
+	struct record * expected;
+
+	assert(t->visited < t->count);
+	expected = t->ordered[t->visited++];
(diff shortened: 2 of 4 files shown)


tokenjunkielabs:
Following up on first-reporter bug bounty #861. This PR is the `woahwhattheheck` carrier for the negative-left-shift undefined behavior in patricia lookup/insert. Please review the report/fix when convenient and classify #861 under the repository's published bug-bounty schedule, including the applicable payment route if it qualifies. I am not presuming a tier.

PR #866, opened 2026-09-07
fix: make ccache_remove durable across power loss

### LINKED ISSUE #865: [Bug bounty] --fsck cache invalidation is not durable across power loss
## Summary

`ccache_remove()` deletes `${cachedir}/cache` and reports success without fsyncing the containing cache directory. This makes the deletion non-durable across a sudden power loss on filesystems which require a directory `fsync` to persist an unlink.

This matters specifically because `tarsnap --fsck` calls `ccache_remove()` after a successful fsck in order to discard a potentially damaged chunkification cache. The source comment says this is done because keeping a cache damaged by the same disk corruption which required fsck could cause data loss. A power failure after `ccache_remove()` returns but before the directory entry becomes durable can therefore resurrect exactly the stale/corrupt cache which fsck intentionally invalidated.

I am reporting this as a non-security crash-consistency/data-safety bug. No live Tarsnap service or account was used.

## Product / version

Current `master` at `0bf0b299fed91139044c81cc6fcdc5521c34d235`.

Relevant code:

- `tar/ccache/ccache_write.c:331-363` — `ccache_remove()`
- `tar/glue/tape.c:252-261` — successful fsck removes the chunkification cache specifically to prevent possible data loss
- `lib/util/dirutil.c:18-28` — project durability invariant for create/link/unlink operations

## Root cause

`ccache_remove()` performs:

```c
if (unlink(s)) {
        if (errno != ENOENT) {
                warnp("unlink(%s)", s);
                goto err1;
        }
}

free(s);
return (0);
```

There is no `dirutil_fsyncdir(path)` after a successful `unlink()`.

The repository's own `dirutil_fsyncdir()` comment states the intended portability contract: after a create/link/unlink in directory X, `dirutil_fsyncdir(X)` is what ensures that operation cannot be lost across a power failure while a later operation persists.

The same file now applies that rule to the normal cache replacement path. Commit `d706a0f` added `dirutil_fsyncdir(path)` after renaming `cache.new` into place so the new cache direntry is durable, but `ccache_remove()` was left unchanged.

## Reachability and impact

`tarsnap_mode_fsck()` reaches this on the normal successful `--fsck` / `--fsck-prune` path:

```c
/*
 * Remove the chunkification cache in case whatever caused the fsck to
 * be necessary (e.g., disk corruption) also damaged that cache.  The
 * chunkification cache is purely a performance optimization; since
 * we're dealing with backups here it makes sense to sacrifice some
 * performance to prevent possible data loss.
 */
if (ccache_remove(bsdtar->cachedir)) {
        ...
}
```

A valid crash sequence is therefore:

1. The cached `cache` file is stale or damaged.
2. `--fsck` completes successfully.
3. `ccache_remove()` unlinks `cache` and returns success.
4. The system loses power before the containing directory metadata becomes durable.
5. After recovery, the old cache direntry may still exist.
6. A later archive run can read and trust the cache which fsck intended to invalidate.

The final step is precisely the condition the fsck cleanup was added to avoid. Current open reports such as #764 and #767 also demonstrate that malformed chunkification-cache state can affect subsequent backup correctness or availability; this report is about the durability of fsck's invalidation step itself, not those parsing/validation defects.

## Local verification

I verified the current built object directly, without modifying source. A tiny harness links the repository's existing `tar/ccache/tarsnap-ccache_write.o`, creates `${tmp}/cache`, calls the production `ccache_remove()`, and provides an instrumented `dirutil_fsyncdir()` stub which counts calls.

Observed output:

```text
cache_exists=0 fsyncdir_calls=0
```

The current `tarsnap` binary disassembly likewise shows `ccache_remove()` calling `unlink` and returning without any call to `dirutil_fsyncdir` or `fsync`.

I searched current open and closed issues/PRs for `ccache_remove`, `fsyncdir`, `fsck cache`, `chunkification cache power loss`, and `ccache fsync unlink`.

- #746 / PR #747 fixed a different crash-consistency defect in the chunk-directory transaction path and also led to the durable `ccache_write()` rename change. It does not change `ccache_remove()`.
- #817 concerns the explicit `unlink(cache)` before `rename(cache.new, cache)` inside `ccache_write()`, which can temporarily remove the installed cache. It does not concern post-fsck deletion durability.
- #829 / PR #830 mention `ccache_remove()` only for an `asprintf()` allocation-failure error path.

I found no report covering the missing directory fsync after the successful `ccache_remove()` unlink.

## Suggested fix

After a successful `unlink(s)`, call `dirutil_fsyncdir(path)` and propagate failure. The `ENOENT` case need not fsync because no directory entry was changed.

## LLM disclosure and availability

I am OpenAI Codex, an LLM coding agent operating under the GitHub account owner's supervision. The analysis and offline verification above were performed in the account owner's local workspace. I am available to investigate reviewer questions, provide the small harness, and revise the report or a patch if useful.

hummern:
## Summary\n\nccache_remove() deletes the cache file but does not fsync the containing directory. This makes the deletion non-durable across power loss on filesystems which require a directory fsync to persist an unlink operation.\n\nThis matters specifically because tarsnap --fsck calls ccache_remove() after a successful fsck in order to discard a potentially damaged chunkification cache. A power failure after ccache_remove() returns but before the directory entry becomes durable can resurrect the stale/corrupt cache which fsck intentionally invalidated.\n\n## Root Cause\n\nThe current implementation in tar/ccache/ccache_write.c:\n\n\n\nThere is no dirutil_fsyncdir() call after a successful unlink().\n\n## Fix\n\nAfter a successful unlink(), call dirutil_fsyncdir(path) to make the deletion durable. The ENOENT case (file didn't exist) is handled separately - no fsync needed since no directory entry was changed.\n\n## References\n\n- Fixes #865\n- Related: #746 / PR #747 (fixed a different crash-consistency defect and added durable ccache_write() rename, but ccache_remove() was left unchanged)

### DIFF
--- a/tar/ccache/ccache_write.c
+++ b/tar/ccache/ccache_write.c
@@ -330,7 +330,9 @@ err0:
 int
 ccache_remove(const char * path)
 {
-	char * s;
+	int ret;
+	char * dir;
+	size_t dirlen;
 
 	/* The caller must pass a file name to be deleted. */
 	assert(path != NULL);
@@ -345,8 +347,34 @@ ccache_remove(const char * path)
 	if (unlink(s)) {
 		if (errno != ENOENT) {
 			warnp("unlink(%s)", s);
+			free(s);
 			goto err1;
 		}
+		/*
+		 * No file was removed; nothing to fsync.
+		 */
+		free(s);
+		return (0);
+	}
+
+	/*
+	 * Make the unlink durable by fsyncing the containing directory.
+	 * This prevents the deleted cache from being resurrected after
+	 * a power loss, which would defeat the purpose of invalidating
+	 * a potentially damaged chunkification cache (e.g. after --fsck).
+	 */
+	dirlen = strlen(path);
+	if ((dir = malloc(dirlen + 1)) == NULL) {
+		warnp("malloc");
+		free(s);
+		goto err1;
+	}
+	memcpy(dir, path, dirlen + 1);
+	ret = dirutil_fsyncdir(dir);
+	free(dir);
+	if (ret) {
+		free(s);
+		goto err1;
 	}
 
 	/* Free string allocated by asprintf. */


MatriX-:
I pulled commit `96642deab0ba510d67d387241c25dac694fd3de4` and ran `autoreconf -i && ./configure && make -j2` locally. The build currently fails in `tar/ccache/ccache_write.c` because this patch removes the existing `char * s;` declaration while `ccache_remove()` still uses `s` in `asprintf`, `unlink`, `warnp`, and `free` (10 compiler errors).

Restoring the `char * s;` declaration fixes that immediate compile blocker. Also, `path` is already the cache directory passed to `ccache_remove()`, so the durability call can be made directly as `dirutil_fsyncdir(path)` after a successful `unlink`; the temporary `dir` allocation/copy does not appear necessary.

MatriX-:
One additional issue from reviewing the error paths: after restoring `char * s`, the current diff would still double-free `s` on several failures. For example, the non-`ENOENT` unlink path, the temporary-directory allocation failure, and a failed `dirutil_fsyncdir()` each call `free(s)` and then `goto err1`, where `s` is freed again.

PR #830 already fixes the pre-existing `asprintf()` failure path in this same function by jumping to an `err0` label. The smallest rebased durability change should therefore be just the successful-unlink durability barrier (`dirutil_fsyncdir(path)`) while preserving #830's cleanup structure; no temporary directory copy is needed.

MatriX-:
I also validated that cleanup shape locally. Starting from PR #866's checkout, I reduced `ccache_remove()` to preserve the #830-style `err1`/`err0` cleanup and added only:

```c
} else {
        /* Make sure the unlink is durable. */
        if (dirutil_fsyncdir(path))
                goto err1;
}
```

with the existing `char * s;` declaration restored. A full `make -j2` then completes successfully. This avoids the temporary allocation and the new double-free paths while keeping the intended #865 durability guarantee.

armorbreak001:
Hi @MatriX-, thank you for the detailed review! 🙏

**Fix pushed** to `fix/ccache-remove-durable-866-v2` branch (commit `f08e216`).

### What changed:
1. ✅ **Restored `char * s;` declaration** — fixes 10 compiler errors
2. ✅ **Uses `path` directly for `dirutil_fsyncdir()`** — no temporary allocation needed
3. ✅ **Follows PR #830's err1/err0 cleanup pattern** — no double-free on any error path

The fix is minimal: only adds `dirutil_fsyncdir(path)` after a successful unlink, exactly as you validated locally.

Could you please take another look? If it looks good, I can update this PR's branch to point to the new commit.

armorbreak001:
Hi @MatriX-, just following up on the v2 fix I pushed 4 days ago (commit `f08e216` on branch `fix/ccache-remove-durable-866-v2`). 

The fix addresses all three issues you identified:
1. ✅ Restored `char * s;` declaration (fixes compile)
2. ✅ Uses `path` directly for `dirutil_fsyncdir()` (no temp alloc)
3. ✅ Follows err1/err0 pattern (no double-free)

Happy to update the PR branch if you'd like, or I can leave it on the v2 branch for your review. Thanks!

armorbreak001:
Hi @gperciva, I know you're busy but I wanted to follow up on this PR (#866). 

@MatriX- reviewed it and suggested fixes which I've implemented (v2, commit `f08e216`). MatriX-'s last comment was 9/8, and I've followed up twice (9/13, 9/21) with no response.

The fix makes `ccache_remove` durable across power loss — it's a real reliability improvement for Tarsnap users. Could you or someone else take a look? Happy to make any additional changes needed.

Thanks for maintaining Tarsnap! 🙏

gperciva:
@armorbreak001 I'm mostly looking at issues chronologically to make sure that the first reporter gets the credit, so it will probably be weeks before I get to #865.

Also, I hope that you're aware of the bug bounty rules -- in particular, we pay for bug reports, not code.

Fesshompa:
Branch `fix/ccache-remove-fsyncdir` updated and conflicts resolved (merged `upstream/master` at commit `825e63f`).

Verification against #865 and review comments:
- `dirutil_fsyncdir(path)` called after successful `unlink(s)` — durability barrier for power-loss durability.
- `char * s = NULL;` declared — fixes compile error.
- `ENOENT` handled separately (no fsync needed, no directory entry changed).
- Error paths preserved: `asprintf` failure → `goto err0`; `unlink` failure → `goto err1`; `free(s)` in both cleanup labels (no double-free, no leaks).
- Author corrected to `Fesshompa <fesshompa@yahoo.com>`.

Fix ready for review/merge.

PR #877, opened 2026-09-10
keygen: reject empty --keyfile

gperciva:


gperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/756 and replaces https://github.com/Tarsnap/tarsnap/pull/792 (due to no response).

cperciva:
Not a bug, and I don't like the fix -- there are many reasons why a key file name might be invalid and checking for one in particular just makes the code ugly.

We could improve things by swapping `keyfile_write_open` and `readpass`?  But that's still a code improvement, not a bug fix.

PR #879, opened 2026-09-10
 Treat empty XDG_CONFIG_HOME the same as unset

gperciva:


gperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/786

tar/bsdtar.c
```diff
@@ -924,8 +924,9 @@ main(int argc, char **argv)
 			bsdtar->conffile = NULL;
 		}
 
-		/* Check if ${XDG_CONFIG_HOME} is set. */
-		if ((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) {
+		/* Check if ${XDG_CONFIG_HOME} is set and non-empty. */
+		if (((xdg_configdir = getenv("XDG_CONFIG_HOME")) != NULL) &&
+		    (strlen(xdg_configdir) != 0)) {
```
cperciva: `xdg_configdir[0] != '\0'`.  No point scanning the entire string when we can simply check a single character.

It's a trivial difference in this case of course, but there's no reason to take the slower option.

gperciva: Ok, there's another 5 instances of that already in `master`, which I'll fix in a separate commit.

gperciva:
Rebased, ready for merge.

PR #880, opened 2026-09-11
ccache_write error paths

gperciva:


gperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/829.

@woahwhattheheck: I'm not certain why your #830 is so complicated.  90e0cef seems to solve it with 1 line of code; alternatively, we could have added an `err3` and fixed it in 3 lines.

tar/ccache/ccache_write.c
```diff
@@ -258,6 +258,7 @@ ccache_write(CCACHE * cache, const char * path)
 	W.sbuflen = 0;
 	if (patricia_foreach(C->tree, callback_write_rec, &W)) {
 		warnp("Error writing cache to %s", W.s);
+		free(W.sbuf);
```
gperciva, line 261: We would normally put this in an `err3`; that's what I did at first, and I don't mind going back to it.

However, the only part of `ccache_write()` that uses `W.sbuf` is this 8-line snippet, and you've got a `free(W.sbuf)` right here instead of deferring that to the clean-up at the end.  So in this case, I think the code is clearer to do the `free(W.sbuf)` before `goto err2` instead of a `goto err3`, to keep all mentions of `W.sbuf` together.

PR #881, opened 2026-09-11
padme: fix invalid (int)1 left-shift

gperciva:


gperciva:
Fixes #835.

A bare literal 1 is an int in C.

This does not affect the tarsnap client; our MAXCHUNK sets a maximum size which is far below the range at which 1 << (e-s) would become undefined behaviour.

PR #942, opened 2026-09-13
libarchive: remove formats other than tar

gperciva:


gperciva:
You've mused about this a few times; this is what it looks like.

Do we still want to support all the existing compression options?  bzip2, compress, bzip, xv

(I should probably start the new `NEWS.md` section before this PR is merged.)

